PR #5, opened 2020-04-18
Add toggle fullscreen support

coderkalyan:
This adds fullscreen support to views. When the keybinding Mod+shift+f is pressed, the current view will become fullscreen (if not already fullscreen), otherwise it will return to regular size.

### DIFF
--- a/src/output.zig
+++ b/src/output.zig
@@ -24,6 +24,7 @@ pub const Output = struct {
 
     /// The top of the stack is the "most important" view.
     views: ViewStack(View),
+    fullscreen_view: ?*View,
 
     /// A bit field of focused tags
     current_focused_tags: u32,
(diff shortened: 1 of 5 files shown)


src/output.zig
```diff
@@ -133,7 +135,30 @@ pub const Output = struct {
     }
 
     pub fn arrange(self: *Self) void {
-        self.arrangeViews();
+        if (self.fullscreen_view) |view| {
+            if (self.fullscreen_view == null) {
+                Log.Debug.log("Yeah it is", .{});
+            }
```
ifreund, line 141: Looks like leftover printf debugging.

ifreund:
One last comment, I think this is actually closer to a monocle layout than true, exclusive full screen (which we probably want as well eventually). Could you adjust the nomenclature to respect that?

And thanks for working on this!

coderkalyan:
Sure thing, I'll implement the changes tomorrow!

ifreund:
I'm going to go ahead and close this as stalled. If you'd like to continue work on this feel free to resolve the conflicts and reopen or make a new PR.

PR #25, opened 2020-05-20
Add multiple layouts and a commands to switch the current layout

Leon-Plickat:
This PR adds multiple layouts to river.

State:

* [X] Layouts
  * [X] Full
  * [X] Top master
  * [X] Right master
  * [X] Bottom  master
  * [X] left master
* [x] Switching mechanism
  * [X] Automatically use "full" for one or no visible views
  * [X] Add `layout` variable to `Output` class to control which layout gets chosen in `arrangeViews()`
  * [X] Add `layout` variable to `Config` class to control the default layout of new outputs
  * [X] Set `LeftMaster` as the default layout
  * [x] Add command to change the layout of the focused output
  * [x] Add command to change the default layout of new outputs

### DIFF
--- a/src/Output.zig
+++ b/src/Output.zig
@@ -384,6 +384,16 @@ pub fn arrangeViews(self: *Self) void {
         break :blk count;
     };
 
+    // A single view should always use the maximum available space. This is
+    // implemented via the "full" layout to remove the need of every single
+    // layout to explicitly handle this edge case or the other edge case of
+    // no visible views.
+    if (visible_count <= 1) {
+        layoutFull(self, visible_count, output_tags);
+        return;
+    }
+
+    // TODO layout switching mechanism
     //layoutFull(self, visible_count, output_tags);
     //layoutTopMaster(self, visible_count, output_tags);
     //layoutRightMaster(self, visible_count, output_tags);
--- a/src/Output.zig
+++ b/src/Output.zig
@@ -52,10 +52,22 @@ master_count: u32,
 /// Percentage of the total screen that the master section takes up.
 master_factor: f64,
 
+/// Current layout of the output.
+layout: Layouts,
+
 // All listeners for this output, in alphabetical order
 listen_destroy: c.wl_listener,
 listen_frame: c.wl_listener,
 listen_mode: c.wl_listener,
+
+// All possible layouts.
+pub const Layouts = enum {
+    TopMaster,
+    RightMaster,
+    BottomMaster,
+    LeftMaster,
+    Full,
+};
 
 pub fn init(self: *Self, root: *Root, wlr_output: *c.wlr_output) !void {
     // Some backends don't have modes. DRM+KMS does, and we need to set a mode
@@ -91,6 +103,9 @@ pub fn init(self: *Self, root: *Root, wlr_output: *c.wlr_output) !void {
     self.master_count = 1;
 
     self.master_factor = 0.6;
+
+    // LeftMaster is the default layout for all outputs
+    self.layout = Layout.LeftMaster;
 
     // Set up listeners
     self.listen_destroy.notify = handleDestroy;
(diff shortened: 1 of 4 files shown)


src/Output.zig
```diff
@@ -69,6 +69,32 @@ pub const Layouts = enum {
...
+    .{ .name = "TopMaster",    .layout = Layouts.TopMaster, },
+    .{ .name = "RightMaster",  .layout = Layouts.RightMaster, },
+    .{ .name = "BottomMaster", .layout = Layouts.BottomMaster, },
+    .{ .name = "LeftMaster",   .layout = Layouts.LeftMaster, },
+    .{ .name = "Full",         .layout = Layouts.Full, },
+};
+// zig fmt: on
+
+pub fn getLayoutByName(self: Self, name: []const u8) Layouts {
+    const current = for (LayoutNames) |current| {
```
ifreund, line 88: This variable is unused 
```suggestion
    for (LayoutNames) |current| {
```

src/Output.zig
```diff
@@ -52,11 +52,49 @@ master_count: u32,
...
 listen_mode: c.wl_listener,
 
+// All possible layouts.
+pub const Layouts = enum {
+    TopMaster,
+    RightMaster,
+    BottomMaster,
+    LeftMaster,
+    Full,
+};
```
ifreund, line 70: Style nitpick, this enum should be called `Layout` not `Layouts`.

Leon-Plickat:
I fixed the style issues.

Is the git history fine or should I cleanup the commits?

ifreund:
A bit of clean up would be appreciated, I'm testing this locally now to make sure the code I read works as I think it does :)

Leon-Plickat:
Commit history should now be cleaner.

src/command/layout.zig
```diff
@@ -0,0 +1,35 @@
...
+const c = @import("../c.zig");
+
+const Arg = @import("../Command.zig").Arg;
+const Seat = @import("../Seat.zig");
+
+pub fn defaultLayout(seat: *Seat, arg: Arg) void {
+    const layout_name = arg.str;
+    var config = seat.input_manager.server.config;
+    config.layout = seat.focused_output.getLayoutByName(layout_name);
+}
```
ifreund, line 27: Hmm, I just realized that this is never used. Maybe we should just get rid of it. (also this modifies a copy of `server.config` not the member itself, but that's irrelevant if we remove it)

Leon-Plickat: The idea behind that command is to allow changing the default layout. If the user can not change it, all new outputs will start with the LeftMaster layout.

Leon-Plickat: Although thinking about it, that command is probably flawed anyway: If it gets executed in a `river-startup.sh` script or something similar, all outputs are already connected...

ifreund: That makes sense, though in it's current state it doesn't seem to do that.

Leon-Plickat: I'll remove it. Changing the default layout of outputs can probably be done with a script that listens for river events and automatically changes the layout.

Leon-Plickat: Done. I also removed the `layout` variable from the `Config` class. Since the default layout is no longer configurable, it does not need to be in `Config`.

lazy-dolphin:
@Leon-Plickat 
>   * [x]  Automatically use "full" for one or no visible views

How is this feature useful though? Doesn't the window span the entire screen when there is one visible view in any other layout?

Leon-Plickat:
> Doesn't the window span the entire screen when there is one visible view in any other layout?

Not necessarily. You have to understand that that no views or one view are *edge cases*, which need to be specifically handled.

Initially I experimented with a bunch more layouts (like a grid one and a fancy overlapping one, etc...). The majority of them would have massively increased complexity if I have handled these two edge cases in all of them individually. And one of the most important axioms of programming is "Do not repeat yourself".

The final PR does not have these layouts, as we discussed on IRC that river should remain fairly minimal, but I decided to leave this in anyway. Call it future proofing: The work on layouts is not done yet, and having this in the code has the capability of removing a lot of complexity later on. 

If we eventually switch to fully external layouts, this might get removed again to give users a bit more power over window placement.

Issue #23, opened 2020-05-12
Suggestion for improving zoom command


lazy-dolphin:
Performing `zoom` command two times on the same window should bring it back to the initial position, similar to `spectrwm`.

PR #46, opened 2020-06-08
Implement `swap-master` command

lazy-dolphin:
Implements #23. `zoom` command was quite different in behavior so a new command had to be written.

### DIFF
--- a/doc/riverctl.1
+++ b/doc/riverctl.1
@@ -4,7 +4,7 @@
 riverctl \- command-line interface for controlling river
 
 .SH SYNOPSIS
-.BR riverctl " " close | declare-mode | enter-mode | exit | focus-output | focus-view | layout | map | mod-master-count | mod-master-factor | send-to-output | set-focused-tags | set-view-tags | spawn | toggle-float | toggle-focused-tags | toggle-view-tags | zoom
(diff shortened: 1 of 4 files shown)


ifreund:
This is currently inconsistent with `zoom` in that when the master view is focused and `swap-master` is executed focus is not switched to the new master.

Also, when swapping e.g. the third view in the stack with the master and immediately swapping back, then switching to the master with `focus-view next`/`focus-view previous` executing `swap-master` again will swap with the second view not the third view which also seems inconsistent.

Honestly I'm not 100% convinced that this is a good feature to have. It kinda breaks the stack-based workflow in which the views stay roughly ordered by importance because usage of the zoom command effectively sorts them in the order of most recent focus. It Is also quite a complex command and imo not entirely obvious/intuitive in its behavior due to being dependent on the previous execution of the command, which goes against river's goal of simplicity.

However, I do think river should let you do this if you want to. I'd prefer adding several simpler commands which can be used together with a bit of a shell scripting to achieve this behavior.

lazy-dolphin:
> Honestly I'm not 100% convinced that this is a good feature to have. It kinda breaks the stack-based workflow in which the views stay roughly ordered by importance because usage of the zoom command effectively sorts them in the order of most recent focus. It Is also quite a complex command and imo not entirely obvious/intuitive in its behavior due to being dependent on the previous execution of the command, which goes against river's goal of simplicity.

I didn't think of this justification about why most stack based window managers shipped with a `zoom` command. In `spectrwm`, my workflow consisted of ordering windows in the stack and I realize now it's more time consuming to maintain a certain order of windows, as it takes less keystrokes to switch to recent windows one was working with. I agree with the sentiment of keeping `river` as simple as possible.

Issue #42, opened 2020-06-04
Add command to set border/padding options


ifreund:
I think it should look something like this:
```
riverctl set-option border_width 4
```
```
riverctl set-option outer_padding 0
```

PR #47, opened 2020-06-09
Implement set-option command and border/padding options

lazy-dolphin:
Implements feature described in issue #42.
I still need to implement:
- [ ] `man` page description for `set-option` command.
- [ ] `man` page section for options.
- [ ] `set-option` doc comment.

Need further guidance for implementing the following:
- [ ] Display the updated `outer_padding` as soon as `set-option` command is called.


lazy-dolphin:
Setting a large value like `1000` to `border_width` or `outer_padding` crashes `river`. I will try to look for a fix for this tomorrow.

Leon-Plickat:
> Setting a large value like 1000 to border_width or outer_padding crashes river. I will try to look for a fix for this tomorrow.

I have two ideas what could cause this. Which such a large border or padding,
* you might cause a division by zero or other unhandled edge cases in the layout functions (maybe even something related to the excessive type casting there).
* you might cause surfaces with a negative size, which is a protocol error and will likely cause a crash in some (or even all) libwayland functions.

A Possible solution consists of two steps:
* Limit the maximum border size of views in the layout functions when the view size is reduced to make space for border and padding.
* Limit the minimum width and size of widths to some arbitrary number (like 50 by 50). If you do this after the positioning, then layouts might be a bit crooked for these extreme cases, but the logic would be considerably easier.

Another possible solution is to not solve this problem, as I am in the middle of reworking layouts anyway.

river/command/set_option.zig
```diff
@@ -0,0 +1,45 @@
...
+
+const std = @import("std");
+
+const c = @import("../c.zig");
+
+const Error = @import("../command.zig").Error;
+const Option = @import("../command.zig").Option;
+const Seat = @import("../Seat.zig");
+
+/// Close the focused view, if any.
```
Leon-Plickat, line 26: This comment is wrong.

river/command/set_option.zig
```diff
@@ -0,0 +1,45 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 Isaac Freund
```
Leon-Plickat, line 3: Nit-picking, but you should add your own name, since you created this file.

ifreund:
> Another possible solution is to not solve this problem yet, as I am in the middle of reworking layouts anyway.

I'd be fine with this for the time being. Since the value that will trigger a crash will depend on the size of outputs/number of surfaces determining what that value is in a robust way wont be too straight forward. Really, I think what we need to do is add sanity checks in the layout functions to prevent the crash and print an error message instead, but that can wait for after @Leon-Plickat finishes his work.

river/command/set_option.zig
```diff
@@ -0,0 +1,45 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const c = @import("../c.zig");
```
Leon-Plickat, line 20: I do not think this is needed here.

Leon-Plickat:
> I'd be fine with this for the time being. Since the value that will trigger a crash will depend on the size of outputs/number of surfaces determining what that value is in a robust way wont be too straight forward. Really, I think what we need to do is add sanity checks in the layout functions to prevent the crash and print an error message instead, [...]

Honestly, I think the only thing we actually need is an arbitrary minimum size of the surfaces of views. In my efforts I have moved the "shrinking" of the pending box into its own function, where I could implement this trivially with a simply add a check (something like `size = maximum(size, 50)`). That way, crashes are avoided. The user could still use not sensible values for padding and border, but since it a) does not crash and b) layouts will be handled by user-written, external executables, we should just go with them.

Leon-Plickat:
Speaking generally, does this PR not break the current settings convention? I was under the impression that each setting kinda gets its own command (example: `layout full`) instead of one global command to set everything (`set-option variable value`). I am perfectly fine with both ways, but we should stay consistent.

ifreund:
> Speaking generally, does this PR not break the current settings convention? I was under the impression that each setting kinda gets its own command (example: layout full) instead of one global command to set everything (set-option variable value). I am perfectly fine with both ways, but we should stay consistent.

This is the suggestion I made in #42, I think it makes sense to put all the options behind a `set-option` command to separate them from "action" commands like zoom, spawn, or focus-view. In the end it's still pretty arbitrary though.

Leon-Plickat:
> I think it makes sense to put all the options behind a set-option command to separate them from "action" commands like zoom, spawn, or focus-view.

I see. So the idea is to eventually port all options to `set-option`? In that case, logic might be required to check whether all views on an output (or even globally) need to be rearranged for the changed option.

ifreund:
> I see. So the idea is to eventually port all options to set-option? In that case, logic might be required to check whether all views on an output (or even globally) need to be rearranged for the changed option.

If it becomes an issue yeah, for now we can just keep things simple. Wouldn't be all that hard to implement if needed.

river/command/set_option.zig
lazy-dolphin: Thanks for reviewing!

PR #48, opened 2020-06-10
[WIP] External layouts

Leon-Plickat:
This is not even remotely finished yet, but I'll park it here anyway so I have a place to brain-dump.

Progress:
* [X] Remove `layoutMasterStack()`
* [X] Refactor `layoutFull()`
  * [X] Minimal window size
* [x] layout switching
  * [X] simplify `arrangeViews()`
  * [X] fallback to `layoutFull()` when `layoutExternal()` fails.
  * [x] Find a sane way to set and store the path to the layout executable as well as possible commands flags to it (which I think are required; example: `riverctl layout ~/my-fancy-layout --master-top`)
  * [x] Change layout command
* [x] `layoutExternal()`
  * [X] create command
  * [X] execute layout executable
  * [x] catch output in buffer
  * [x] Parse buffer
  * [X] apply to views
  * [X] Offset, shrinking, minimal window size
* [x] Write documentation
* [x] Cleanup commit history

The biggest hurdle by far is definitely the zig standard library not being documented.

A big problem for users will be, that river will halt while the layout executable is running. This is fine for now, but before version 1.0 it should probably be made async somehow, so that a faulty layout executable does not block the entire system, requiring a restart or Sysrq magic. Or instead of async, just abort after an arbitrary amount of time (like 1 second) and fall back to `layoutFull()`.

### DIFF
--- a/contrib/random-correct-layout.sh
+++ b/contrib/random-correct-layout.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Randomized Layout for debug purposes.
+
+CLIENTS="$1"
+OUTPUT_WIDTH="$4"
+OUTPUT_HEIGHT="$5"
+
+for _ in $(seq 1 "$CLIENTS")
+do
+	WIDTH="$(( ( OUTPUT_WIDTH  / 5 ) ))"
+	HEIGHT="$(( ( OUTPUT_HEIGHT  / 5 ) ))"
+	X="$(( ( RANDOM % ( OUTPUT_WIDTH  - WIDTH  ) )  + 1 ))"
+	Y="$(( ( RANDOM % ( OUTPUT_HEIGHT - HEIGHT ) )  + 1 ))"
+	echo "$X $Y $WIDTH $HEIGHT"
+done
+
--- a/contrib/random-incorrect-layout.sh
+++ b/contrib/random-incorrect-layout.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+# Randomized Layout for debug purposes. This version randomly makes some errors
+# see how river handles incorrect output of layout executables.
+
+CLIENTS="$1"
+OUTPUT_WIDTH="$4"
+OUTPUT_HEIGHT="$5"
+
+for _ in $(seq 1 "$CLIENTS")
+do
+	WIDTH="$(( ( OUTPUT_WIDTH  / 5 ) ))"
+	HEIGHT="$(( ( OUTPUT_HEIGHT  / 5 ) ))"
+	X="$(( ( RANDOM % ( OUTPUT_WIDTH  - WIDTH  ) )  + 1 ))"
+	Y="$(( ( RANDOM % ( OUTPUT_HEIGHT - HEIGHT ) )  + 1 ))"
+
+	# Mix in some errors
+	case "$(( ( RANDOM % 10 ) ))" in
+		0) # Too few layout rows
+			;;
+
+		1) # Too many layout rows
+			echo "$X $Y $WIDTH $HEIGHT"
+			echo "$X $Y $WIDTH $HEIGHT"
+			;;
+
+		2) # Too few layout columns
+			echo "$X $Y $WIDTH"
+			;;
+
+		3) # Too many layout columns
+			echo "$X $Y $WIDTH $HEIGHT $X"
+			;;
+
+
+		4) # Negative view size
+			echo "$X $Y -$WIDTH $HEIGHT $X"
+			;;
+
+		*) # Expected behaviour
+			echo "$X $Y $WIDTH $HEIGHT"
+			;;
+	esac
+done
+
--- a/doc/river-layouts.1.scd
+++ b/doc/river-layouts.1.scd
@@ -1,0 +1,74 @@
+RIVER-LAYOUTS(7) "github.com/ifreund/river"
+
+# NAME
+river-layouts - Details on layout generators for river
+
+# DESCRIPTION
+River can use external window management layouts. To get such a layout, river
+will run an executable and parse its output. This document outlines how such
+a layout generator interacts with river.
+
+# INPUT
+When running the executable, river will provide it with five parameters which
+are appended to the end of the command in the following order:
+
+. The amount of visible clients (integer)
+. The amount of views dedicated as master (integer)
+. The screen size multiplier for the master area (float between 0.0 and 1.0)
+. The useable width of the output (integer)
+. The useable height of the output (integer)
+
+A layout generator may choose to ignore any of these values except
+for the first one.
+
+# OUTPUT
+River expects four integer values for each window: The x position, the y
+position, the width and the height. These must be separated by spaces. A
+window configuration having fewer or more than four values is an error and will
+cause river to fall back the full layout.
+
+A layout generator needs to output position and size for every visible window.
+The window configurations are separated by a newline. Too few or too many
+outputted window configurations is an error and will cause river to fall back
+to the full layout.
+
+River will apply the position and dimensions in the order they are outputted to
+the visible windows in the stack from top to bottom.
+
+The output of a layout generator is not required to remain the same when called
+with identical parameters. Layouts are allowed to also depend on external
+factors or be completely random.
+
+# WINDOW DIMENSIONS and POSITION
+Layout generators are not supposed to include padding or leave space for window
+borders. The window dimensions will be shrunk by river to make space for these.
+River enforces a minimal window width and height of 50.
+
+Layout generators operate on a special coordinate grid from 0 to the maximum
+useable width or height of an output with the coordinate 0-0 being positioned at
+the top-left corner of the useable area of an output. While layout generators
+are free to place windows everywhere (including coordinates below zero or above
+the maximum width or height of an output), beware that the relative positioning
+of this grid on the screen can not be expected to remain constant. River applies
+an offset to window positions, depending on outer padding and the presence of
+desktop widgets like bars. Layout generators can therefore not position windows
+at exact screen coordinates.
+
+Layout generators are not required to make use of the entire available space.
+Windows may overlap.
+
+# EXAMPLE
+Below is an example output of a layout generator for four visible windows. In
+this example layout all four windows have a size of 500 by 500 and are arranged
+in a grid.
+
+```
+0 0 500 500
+500 0 500 500
+0 500 500 500
+500 500 500 500
+```
+
+# SEE ALSO
+*river*(1), *riverctl*(1)
+
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -31,8 +31,13 @@ used to control and configure river.
 *focus-view* *next*|*previous*
 	Focus next or previous view in the stack.
 
-*layout* *top-master*|*right-master*|*bottom-master*|*left-master*|*full*
-	Change the view layout.
+*layout* *full*|_command_
+	Provide a command which river will use for generating the layout of
+	non-floating windows on the currently focused output. See
+	*river-layouts*(1) for details on the expected formatting of the output
+	of layout commands. Alternatively, "full"can be given instead of a
+	command to cause river to use its single internal layout, in which
+	windows span the entire width and height of the output.
 
 *mod-master-count* _integer_
 	Increase or decrease the number of master views. _integer_ can be
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -61,8 +61,10 @@ master_count: u32,
 /// Percentage of the total screen that the master section takes up.
 master_factor: f64,
 
-/// Current layout of the output.
-layout: Layout,
+/// Current layout of the output. If it is "full", river will use the full
+/// layout. Otherwise river assumes it contains a string which, when executed
+/// with sh, will result in a layout.
+layout: []const u8,
 
 /// List of status tracking objects relaying changes to this output to clients.
 status_trackers: std.SinglyLinkedList(OutputStatus),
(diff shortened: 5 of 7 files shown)


Leon-Plickat:
> If you take that route it would be as simple as saving the slice mem.join instead of freeing it right away.

`std.mem.join`, does allocates memory which "must be freed by the caller" (quoting `lib/std/mem.zig`). Therefore `Output.layout` is a pointer to a heap object when the layout command has been called. To simplify code, it makes sense to also make the initial default value, the string "full", a heap object. So it will *always* point to an allocated heap object and therefore needs to be freed before we create a new object for it to point at.

(At least that is how I understand it, might be a bit to C-ish for zig.)

ifreund:
> To simplify code, it makes sense to also make the initial default value, the string "full", a heap object. So it will _always_ point to an allocated heap object and therefore needs to be freed before we create a new object for it to point at.

Yes, this sounds like a very reasonable thing to do.

Leon-Plickat:
The PR is not quite done yet, but the documentation can probably already be reviewed.

doc/river-layouts.1
```diff
@@ -0,0 +1,69 @@
+.TH RIVER-LAYOUTS 1 2020-06-03 github.com/ifreund/river "General Commands Manual"
+
+.SH NAME
+river-layouts \- Details on layout generators for river
+
+.SH DESCRIPTION
+River can use external window management layouts. To get such a layout, river
+will execute an executable and parse its output. This document outlines how such
```
ifreund, line 8:
```suggestion
will run an executable and parse its output. This document outlines how such
```

doc/river-layouts.1
```diff
@@ -0,0 +1,69 @@
...
+a layout generator interacts with river.
+
+.SH INPUT
+When executing the executable, river will provide it with five parameters which
```
ifreund, line 12:
```suggestion
When running the executable, river will provide it with five parameters which
```

doc/river-layouts.1
```diff
@@ -0,0 +1,69 @@
...
+generator may choose to ignore any of these except for the first one.
+
+.SH OUTPUT
+River expects four integer values for each window: The x position, the y
+position, the width and the height. These must be separated by spaces. A
+window configuration having fewer or more then four values is an error and will
+cause river to fall back the full layout.
+
+A layout generator needs to output position and size for every visible window.
+The window configurations are separated by a newline. To few or to many
```
ifreund, line 26:
```suggestion
The window configurations are separated by a newline. Too few or too many
```

doc/river-layouts.1
```diff
@@ -0,0 +1,69 @@
...
+outputted window configurations is an error and will cause river to fall back
+to the full layout.
+
+River will apply the position and dimensions in the order they are outputted to
+the visible windows in the stack from top to bottom.
+
+The output of a layout generator is not expected to remain the same when called
```
ifreund, line 33:
```suggestion
The output of a layout generator is not required to remain the same when called
```

Leon-Plickat:
Slowly getting there. If I made no grave mistakes in other areas, the only two remaining things to do are 1) writing the commands output to a buffer / opening a pipe and 2) parsing the buffer / pipes contents. That should hopefully be done before the end of the week.

Major hurdle currently: `std.mem.strip` as well as `std.mem.tokenize` are a bit meh IMO, so I might approach parsing more manually, meaning iterating over chars instead of slices. Additionally, that way parsing could be done in a single loop instead of a two nested loops, which may be a bit easier to understand (and therefore work on) for future contributors.

Leon-Plickat:
Seems to work, now comes the testing and reviewing.

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
-pub fn layoutBottomMaster(self: *Self, visible_count: u32, output_tags: u32) void {
-    layoutMasterStack(self, visible_count, output_tags, MasterPosition.Bottom);
+    Log.Debug.log("Parsed window layout: x={} y={} width={} height={}", .{ box.x, box.y, box.width, box.height });
+    return true;
 }
 
-/// Wrapper for default layout with master area on the left
-pub fn layoutLeftMaster(self: *Self, visible_count: u32, output_tags: u32) void {
-    layoutMasterStack(self, visible_count, output_tags, MasterPosition.Left);
+test "parse window configuration" {
```
ifreund, line 251: Love to see the test cases!

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
-        (outer_padding * 2) - (border_width * 2) - (view_padding * 2);
-    const layout_height = @intCast(u32, self.usable_box.height) -
-        (outer_padding * 2) - (border_width * 2) - (view_padding * 2);
-    const x_offset = self.usable_box.x + @intCast(i32, outer_padding + border_width + view_padding);
-    const y_offset = self.usable_box.y + @intCast(i32, outer_padding + border_width + view_padding);
+    const visible_count_str = std.fmt.allocPrint(self.root.server.allocator, "{}", .{visible_count}) catch |err| switch (err) {
+        error.OutOfMemory => {
+            @panic("out of memory");
+        },
+    };
```
ifreund, line 276: These could all be "stringified" at once with one call to `allocPrint`. Something like
```zig
const params = std.fmt.allocPrint(allocator, "{} {} {d} {} {}", .{self.master_count, self.master_factor, ...
```
Also, there's only one possible error here so we can just `catch @panic("out of memory");`

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
+    testing.expect(box.y == 10);
+    testing.expect(box.width == 100);
+    testing.expect(box.height == 200);
 }
 
-/// A layout in which every window uses the maximum available space.
-pub fn layoutFull(self: *Self, visible_count: u32, output_tags: u32) void {
+/// Execute an external layout function, parse its output and apply the layout
+/// to the output.
+fn layoutExternal(self: *Self, visible_count: u32, output_tags: u32) bool {
```
ifreund, line 263: Again, returning a `bool` to indicate failure isn't very idiomatic. Prefer a return type of `!void` to infer the error set, using `try functionThatMayFail()` in the body of `layoutExternal`, and a `catch`/`switch` handling the errors in the calling function.

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
+        self.layout,
+        visible_count_str,
+        master_count_str,
+        master_factor_str,
+        layout_width_str,
+        layout_height_str,
+    }) catch |err| {
+        Log.Error.log("Error assemblind layout command.", .{});
+        return false;
+    };
```
ifreund, line 310: add a 
```zig
defer allocator.free(layout_command);
```
here and remove all the calls farther down.

Also I think an 
```zig
const allocator = self.root.server.allocator;
```
at the top of the function would be justified.

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
 
-        if (view.floating) {
-            continue;
-        }
+    // Execute layout executable
+    const child = std.ChildProcess.init(&cmd, self.root.server.allocator) catch |err| {
+        Log.Error.log("Error initializing layout command.", .{});
+        return false;
+    };
```
ifreund, line 321: add defer and remove later `deinit()` calls
```zig
defer child.deinit();
```

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
-            .x = x_offset,
-            .y = y_offset,
-            .width = layout_width,
-            .height = layout_height,
+    const max_output_size = 400 * 1024;
+    const buffer = child.stdout.?.inStream().readAllAlloc(self.root.server.allocator, max_output_size) catch |err| {
+        child.deinit();
+        self.root.server.allocator.free(layout_command);
+        return false;
+    };
```
ifreund, line 337: same as above, add a `defer` to free the memory and remove later individual calls to free.

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
+            }
+        },
+    }
+
+    // TODO abort after 1 second
+
+    child.deinit();
+
+    // Parse layout command output
+    var view_boxen = std.ArrayList(Box).init(self.root.server.allocator);
```
ifreund, line 361: add  a `defer` to deinit the ArrayList

river/Output.zig
```diff
@@ -416,29 +426,19 @@ pub fn arrangeViews(self: *Self) void {
...
-    switch (self.layout) {
-        .Full => layoutFull(self, visible_count, output_tags),
-        .TopMaster => layoutTopMaster(self, visible_count, output_tags),
-        .RightMaster => layoutRightMaster(self, visible_count, output_tags),
-        .BottomMaster => layoutBottomMaster(self, visible_count, output_tags),
-        .LeftMaster => layoutLeftMaster(self, visible_count, output_tags),
+    if (std.mem.eql(u8, self.layout, "full")) {
+        layoutFull(self, visible_count, output_tags);
+    } else if (!layoutExternal(self, visible_count, output_tags)) {
+        Log.Error.log("Falling back to build in 'Full' layout", .{});
```
ifreund, line 440:
```suggestion
        Log.Error.log("Falling back to built in 'Full' layout", .{});
```

contrib/random-correct-layout.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
Leon-Plickat: Nope, `RANDOM` is definitely a bash-ism. At least shell-check tells me so.

river/Output.zig
m-hugo: built-in

contrib/random-incorrect-layout.sh
```diff
@@ -0,0 +1,44 @@
...
+for _ in $(seq 1 "$CLIENTS")
+do
+	WIDTH="$(( ( OUTPUT_WIDTH  / 5 ) ))"
+	HEIGHT="$(( ( OUTPUT_HEIGHT  / 5 ) ))"
+	X="$(( ( RANDOM % ( OUTPUT_WIDTH  - WIDTH  ) )  + 1 ))"
+	Y="$(( ( RANDOM % ( OUTPUT_HEIGHT - HEIGHT ) )  + 1 ))"
+
+	# Mix in some errors
+	case "$(( ( RANDOM % 10 ) ))" in
+		0) # To few layout rows
```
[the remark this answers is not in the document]
m-hugo:
```suggestion
		0) # Not enough layout rows
```

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
+    var it = std.mem.split(buffer, " ");
+    while (it.next()) |token| {
+        Log.Debug.log("Current token: '{}'", .{token});
+        switch (i) {
+            // X
+            0 => {
+                box.x = std.fmt.parseInt(i32, token, 10) catch |err| {
+                    Log.Error.log("Invalid layout: Invalid window X position.", .{});
+                    return false;
                 };
```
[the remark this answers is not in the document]
Leon-Plickat: That function signature won't work. Before calling `parseWindowConfig()`, a new member is added to the array list. We only have the pointer to this member, so we need to work on that. Returning `Box` will make things more complicated.

I'll see what I can do about returning a `LayoutError` enum thingy.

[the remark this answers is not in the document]
ifreund: This would work fine:
```zig
try view_boxen.append(try parseWindowConfig(buffer));
```

Also realized that I typoed the declaration of `LayoutError` I gave, should be 
```zig
const LayoutError = error{
    ...
};
```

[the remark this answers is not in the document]
ifreund: If you want to stick with `addOne()` to allow handling the OOM first, we could do
```zig
const neu_box = try view_boxen.addOne();
neu_box.* = try parseWindowConfig(buffer);
```
Don't worry about what looks like an unnecessary copy, Zig's copy elision and/or result location semantics should eliminate it.

river/Output.zig
```diff
@@ -185,227 +163,259 @@ pub fn sendViewTags(self: Self) void {
...
-                    .height = slave_height + if (i == master_count) slave_height_rem else 0,
+/// Parse a window configuration string and write values to the box
+fn parseWindowConfig(buffer: []const u8, box: *Box) bool {
+    Log.Debug.log("Trying to parse window configuration string \"{}\"", .{buffer});
+    var i: u32 = 0;
+    var it = std.mem.split(buffer, " ");
+    while (it.next()) |token| {
+        Log.Debug.log("Current token: '{}'", .{token});
+        switch (i) {
+            // X
```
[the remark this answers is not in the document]
Leon-Plickat: I cleaned that loop up a bit and I think it is not too bad now.

Leon-Plickat:
I fixed all the issues and (hopefully) made the code a bit more idiomatic.

I decided against too verbose debug and error messages. They do not really help someone writing a layout, when layouts can be effectively audited by ~~hand~~ eye.

doc/river-layouts.1.scd
```diff
@@ -0,0 +1,74 @@
...
+. The useable width of the output (integer)
+. The useable height of the output (integer)
+
+A layout generator may choose to ignore any of these values except
+for the first one.
+
+# OUTPUT
+River expects four integer values for each window: The x position, the y
+position, the width and the height. These must be separated by spaces. A
+window configuration having fewer or more then four values is an error and will
```
Ordoviz, line 27: then → than

doc/riverctl.1.scd
```diff
@@ -31,8 +31,13 @@ used to control and configure river.
 *focus-view* *next*|*previous*
 	Focus next or previous view in the stack.
 
-*layout* *top-master*|*right-master*|*bottom-master*|*left-master*|*full*
-	Change the view layout.
+*layout <command>*
```
Ordoviz, line 34: Should be `*layout* *full*|_command_` for consistency.

ifreund:
I've written an implementation of the "left tiled" layout in python as additional docs and have a fast zig implementation for all 4 standard tiled layouts that I intend to ship with river. Will push those after this is merged.

Leon-Plickat:
Fixed those as well.

river/Output.zig
```diff
@@ -185,214 +159,148 @@ pub fn sendViewTags(self: Self) void {
...
-                    .width = slave_stack_size,
-                    .height = slave_height + if (i == master_count) slave_height_rem else 0,
-                };
-            }
+/// Parse a window configuration string and write values to the box
+fn parseWindowConfig(buffer: []const u8) LayoutError!Box {
+    var i: u32 = 0;
+    var box: Box = undefined;
+    var it = std.mem.split(buffer, " ");
+    while (it.next()) |token| {
```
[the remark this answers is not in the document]
Leon-Plickat: Ah, I was wondering why this was so "manual" in zig, turns out it isn't if done right.

Issue #43, opened 2020-06-04
Make border color configureable


ifreund:
Border color is currently hard-coded to my personal preference. However, this should be made configurable through a command. I'm thinking something like
```
riverctl set-option border_unfocused #0000ff
```
```
riverctl set-option border_focused #ff0000
```

ifreund:
closed by #50

PR #50, opened 2020-06-12
command: Added border_focused_color and border_unfocused_color options

lazy-dolphin:
Implements #43. Hex code is required to be enclosed in double quotes.
```
riverctl set-option border_focused_color "#a1a1a1"
```

### DIFF
--- a/river/Rgb.zig
+++ b/river/Rgb.zig
@@ -1,0 +1,45 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 Rishabh Das
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


river/Rgb.zig
```diff
@@ -0,0 +1,43 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+const Self = @This();
+
+const std = @import("std");
+
+const Error = @import("command.zig").Error;
```
ifreund, line 22: This doesn't get used
```suggestion
```

lazy-dolphin: Fixed.

lazy-dolphin:
Also, improved `Rgb.parseString` function and renamed commit message to be more consistent.

PR #53, opened 2020-06-15
Use scdoc for man pages

### LINKED ISSUE #51: docs: migrate man pages to scdoc
[scdoc](https://git.sr.ht/~sircmpwn/scdoc) is a lightweight man page generator that allows for a much more pleasant writing experience than pure roff. This will make the docs more readable in their plaintext form and allow for easier maintenance.

Ordoviz:
I ran `pandoc -t markdown --atx-headers` on the roff man pages and edited the output manually.

scdoc comes with some downsides, though:
- A new dependency and an extra step between writing and checking the output
- Generated roff files are less readable than hand-written ones
- loss of semantics: In roff we can use `.TP` to create a definition list. In scdoc we need to use indentation.
- scdoc is less established than roff: We cannot use pandoc to convert scdoc files to HTML for example. Text editors and GitHub do not know scdoc: syntax highlighting is non-existing or wrong.

Closes #51


Leon-Plickat:
> Generated roff files are less readable than hand-written ones

Let's be honest, roff files are basically never readable. The format sucks and is overly complex. Also you can read them with `man`.

Writing is more important than reading.

> loss of semantics: In roff we can use .TP to create a definition list. In scdoc we need to use indentation.

This is a definite improvement, IMO.

>scdoc is less established than roff: We cannot use pandoc to convert scdoc files to HTML for example. Text editors and GitHub do not know scdoc: syntax highlighting is non-existing or wrong.

scdoc is widely available.

Why would we need to convert manpages to html?

Why would you really need syntax highlighting for a simple, markdown-like format? Also: `set ft=markdown` works for me.

doc/river.1.scd
```diff
@@ -0,0 +1,35 @@
...
+*river* [*-c* _shell_command_]
+
+# DESCRIPTION
+
+*river* is a dynamic tiling Wayland compositor inspired by dwm and
+bspwm based on wlroots and written in Zig.
+
+# OPTIONS
+
+*-c* _shell_command_
```
Leon-Plickat, line 17: I think `*-c <shell comand>*` is more idiomatic for scdoc.

doc/river.1.scd
```diff
@@ -0,0 +1,35 @@
+RIVER(1) "github.com/ifreund/river" "General Commands Manual"
```
Leon-Plickat, line 1: That is not correct scdoc syntax. It should be something like `river(1) ["github.com/ifreund/river" ["General Commands Manual"]]`

doc/river.1.scd
```diff
@@ -0,0 +1,35 @@
...
+*-c* _shell_command_
+	Run a shell command or give the path to a script that will be run
+	before river gets started. This is where you define keymaps using
+	*riverctl*(1) and autostart programs. Make sure the script is
+	executable.
+
+# CONFIGURATION
+
+View border width, gap sizes, and programs which should float are
+configured in _Config.zig_. Make your changes and recompile.
```
Leon-Plickat, line 26: This is no longer true for most configuration options.

doc/river.1.scd
```diff
@@ -0,0 +1,35 @@
...
+
+Experimental XWayland support can be enabled on compile-time with the
+_-Dxwayland=true_ flag.
```
Leon-Plickat, line 29: Man pages are not compilation instructions. Should be something like "XWayland support may be available if set during compilation."

doc/river.1.scd
```diff
@@ -0,0 +1,35 @@
...
+
+Keyboard layout and other XKB options can be set in _Keyboard.zig_.
```
Leon-Plickat, line 31: See previous comments.

doc/river.1.scd
```diff
@@ -0,0 +1,35 @@
...
+
+# SEE ALSO
+
+*riverctl*(1), *bspwm*(1), *dwm*(1)
```
Leon-Plickat, line 35: Do we really want to link to bspwm and dwm? I do not think those man pages help understanding river any better.

doc/riverctl.1.scd
```diff
@@ -0,0 +1,130 @@
+RIVERCTL(1) "github.com/ifreund/river" "General Commands Manual"
```
Leon-Plickat, line 1: See previous comments.

doc/riverctl.1.scd
```diff
@@ -0,0 +1,130 @@
...
+
+*riverctl* *close*|*declare-mode*|*enter-mode*|*exit*|*focus-output*|
+*focus-view*|*layout*|*map*|*mod-master-count*|*mod-master-factor*|
+*send-to-output*|*set-focused-tags*|*set-view-tags*|*spawn*|*toggle-float*|
+*toggle-focused-tags*|*toggle-view-tags*|*zoom* [_command specific arguments_]
+
+# DESCRIPTION
+
+*riverctl* is a command-line interface inspired by bspc from bspwm
+used to control and configure river.
```
Leon-Plickat, line 16: Readers of the man page do not really care what the inspiration was. The man page should describe what a program is without falling back to "it's just like $other_program".

ifreund:
Welp, looks like we hit a race condition @Leon-Plickat If you'd like to open a follow up PR with those changes they look good to me.

Leon-Plickat:
Sure, I'll go though them in detail to see what can be improved.

But later, I prefer working on one PR at a time.

doc/river.1.scd
Ordoviz: The convention of man-pages(7) is to use a capital title, e.g. RIVER not river. The square brackets indicate optional arguments. They are not part of the syntax.

doc/river.1.scd
Ordoviz: I also see angle brackets used for arguments a lot. Man pages use italics by convention, though.

doc/river.1.scd
Ordoviz: I will fix the outdated info in a separate PR.

doc/river.1.scd
Ordoviz: IMO the compile flag is too important to remove it. XWayland support is already stated in the README but mentioning it twice does not hurt. I think we should either keep it or remove the entire sentence.

doc/river.1.scd
Ordoviz: Agreed, should be replaced with environment variables.

doc/riverctl.1.scd
Ordoviz: Fair point.

PR #57, opened 2020-06-18
Implement moving views via cursor

Leon-Plickat:
As discussed in IRC.

This is supposed to work similar to dwm: Trying to move a window will float it. I ended up implementing cursor warp, because it actually simplifies things a bit; I can remove it later though if desired..

TODO:
* [X] Moving logic
  * [X] Enter move mode
  * [X] Leave move mode
  * [X] Handle cursor events when in move mode
  * [X] Make it easy to implement pointer resizing and other mouse operations in the future
* [X] Add `pointer_modifier: bool` to seat, to track whether the pointer modifier(s) are currently pressed.
* [x] Seat `pointer_modifer` when correct modifier key has been pressed.
* [ ] ~~Implement way to configure modifier key.~~

As it is now, this PR contains some C-like stuff I have to zig-ify first before it can be tested.  Also it might be a good idea to investigate whether function pointers instead of switches in cursor event handling code are more efficient.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -1,6 +1,7 @@
 // This file is part of river, a dynamic tiling wayland compositor.
 //
 // Copyright 2020 Isaac Freund
+// Copyright 2020 Leon Henrik Plickat
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
@@ -24,6 +25,8 @@ const c = @import("c.zig");
 const log = @import("log.zig");
 const util = @import("util.zig");
 
+const Box = @import("Box.zig");
+const Config = @import("Config.zig");
 const LayerSurface = @import("LayerSurface.zig");
 const Output = @import("Output.zig");
 const Seat = @import("Seat.zig");
@@ -41,12 +44,20 @@ wlr_cursor: *c.wlr_cursor,
 wlr_xcursor_manager: *c.wlr_xcursor_manager,
 
 mode: CursorMode,
-grabbed_view: ?*View,
-grab_x: f64,
-grab_y: f64,
-grab_width: c_int,
-grab_height: c_int,
-resize_edges: u32,
+grabbed_view: *View,
+
+/// Distance between cursor and top-left corner of grabbed view
+grab_delta_x: f64,
+grab_delta_y: f64,
+
+/// Dimensions of the output the grabbed view is on
+grab_output_width: u64,
+grab_output_height: u64,
+
+const CursorPosition = struct {
+    x: f64,
+    y: f64,
+};
 
 listen_axis: c.wl_listener,
 listen_button: c.wl_listener,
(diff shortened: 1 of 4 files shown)


Leon-Plickat:
> Didn't notice anything glaringly C-ish.

See enum elements such as `WLR_BUTTON_PRESSED` or `BTN_RIGHT` which I probably have to explicitly import.

> I'd be surprised if function pointers were faster, the switches are likely a lot easier to optimize.

I believe that depends on whether these switches will be turned into jump tables or`if else` chains. Might be worth investigating later, as this is code that is run fairly often, but I'll keep it as is for this PR.

> Maybe we want to allow modifier combinations too as it would barely increase the code complexity.

Agreed.

Leon-Plickat:
A rebase against upstream/master seems to have fixed my issues with `riverctl`.

Moving generally works now. Also working is the automatic float-ing of windows when the user tries to move them. Constraining the moving view to its output also works, as does warping the cursor. The cursor is also constrained, meaning it as well will not move any further when the view is "stuck". On the ToDo-List are still the toggling of pointer operations and implementing a configurable modifier key(-combination).


 Absolute movements are still a bit weird, but this is by design (kinda): Upon initiating a move, river tries to warp the cursor to the top left corner of a view. From there on river will always move the view so that its top left corner is under the cursor, expecting that the cursor warp succeeded. However, since the wlroots Wayland backend does **not** warp the cursor (when river is nested in another compositor), when the user clicks on a view in the river window, river just tries to move the views top left corner to the un-warped cursor, which results in a sudden "jump" of the window (which might hit the output border, making the behaviour non-obvious).

Other absolute movements are probably also affected, but I lack the necessary input devices to test that.

I am debating removing cursor warp because of this. That will require more math per move handling, but make the behaviour for absolute movements more predictable. I will keep the cursor movement tied to the view movement, so that the cursor can not be moved independently of the view during movement (which would look pretty bad), but it will keep its relative position on the view. I will however add cursor warp to resizing, when I eventually implement that, because here it makes sense and is predictable.

ifreund:
This feel nice and smooth on the DRM backend! As you noted, it behaves a bit strangely in a nested session.

> I am debating removing cursor warp because of this. That will require more math per move handling, but make the behaviour for absolute movements more predictable. I will keep the cursor movement tied to the view movement, so that the cursor can not be moved independently of the view during movement (which would look pretty bad), but it will keep its relative position on the view. I will however add cursor warp to resizing, when I eventually implement that, because there it makes sense and is predictable.

This sounds like a good path to take. I'm not 100% sold on cursor warp for moving to be honest, but for resizing I do think it makes a lot of sense. Even for resizing it would still behave strangely in a nested session though no?

I know this is still a draft, but I found a couple of bugs:

1. Releasing a dragged view while the cursor is over another view switches focus to that view.
2. Dragging a tiled view to make it floating doesn't cause the layout to be re-arranged, leaving a gap.
image

If these prove annoying/tedious to fix I don't think they need to block merging this PR as slightly buggy view moving is better than no view moving.

Leon-Plickat:
> Even for resizing it would still behave strangely in a nested session though no?

There would still be an initial "jump", but I think overall the behaviour would be predictable.

Leon-Plickat:
Found some time to work on this again. I removed cursor warp and the output will now be re-arranged when a view is set to floatin by a move operation.

ifreund:
Nice! I pulled this down to test if the first bug I listed was fixed in master and it looks like it is. This involved a small patch to make it work with the changes I made to pending state on master recently. Pushed the result to this branch if you'd like to grab those changes: <https://github.com/ifreund/river/tree/cursor-move>

ifreund:
Also note that configureability  of the pointer modifier doesn't really need to happen in this PR, that can totally wait for a follow up.

Leon-Plickat:
> Also note that configureability of the pointer modifier doesn't really need to happen in this PR, that can totally wait for a follow up.

Probably a good idea. I will finish this up without a configurable modifier.

> Pushed the result to this branch if you'd like to grab those changes

Sure will do.

Leon-Plickat:
Ready for review. Works as expected and should have no nasty edge cases (If I did not forget something).

PR #63, opened 2020-07-08
cursor: close view on middleclick when pointer_modifier is active

Leon-Plickat:
When a user is interacting with (possible floating) views via the pointer, they may want to close a view that way as well.

The middle mouse button is not something you'd click accidentally, especially when a modifier key is also pressed, so I think this has no drawbacks, not even code complexity, as it is literally a single line.


Leon-Plickat:
Also we should probably document pointer behaviour...

ifreund:
Opened up #64 to explore making this more configurable.

PR #67, opened 2020-07-15
Implement csd_filter list

Leon-Plickat:
As discussed in IRC an hour ago.

This adds a list of app_ids which are  allowed to use client side decorations. These views will not be set to tiled (beware: the views will still be tiled by river and her layouts, this is just a status influencing how applications draw their windows), they are instructed to use client side decorations and they will not get a border drawn by river.

Swappshot Thu Jul 16 01:19:40 2020

This settles the CSD vs SSD debate by simply doing both, allowing applications of both kinds to look as intended. Beware that the default is still SSD though.

Currently there is no way of adding any app_ids to the list. I will create a follow-up PR implementing a commands for pushing app_ids to the csd_filter list and the float_filter list.

For testing purposes, the app_id "csd" is hardcoded into the list for now (will be removed with the above PR). You can test this for example with `alacritty --class csd` (alacritty has client side decorations).

### DIFF
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -52,6 +52,9 @@ modes: std.ArrayList(std.ArrayList(Mapping)),
 /// List of app_ids which will be started floating
 float_filter: std.ArrayList([*:0]const u8),
 
+/// List of app_ids which are allowed to use client side decorations
+csd_filter: std.ArrayList([]const u8),
+
 pub fn init(self: *Self) !void {
     self.background_color = [_]f32{ 0.0, 0.16862745, 0.21176471, 1.0 }; // Solarized base03
     self.border_width = 2;
@@ -73,8 +76,14 @@ pub fn init(self: *Self) !void {
     self.float_filter = std.ArrayList([*:0]const u8).init(util.gpa);
     errdefer self.float_filter.deinit();
 
+    self.csd_filter = std.ArrayList([]const u8).init(util.gpa);
+    errdefer self.csd_filter.deinit();
+
     // Float views with app_id "float"
     try self.float_filter.append("float");
+
+    // Client side decorations for views with app_id "csd"
+    try self.csd_filter.append("csd");
 }
 
 pub fn deinit(self: Self) void {
@@ -89,4 +98,5 @@ pub fn deinit(self: Self) void {
     self.modes.deinit();
 
     self.float_filter.deinit();
+    self.csd_filter.deinit();
 }
--- a/river/Decoration.zig
+++ b/river/Decoration.zig
@@ -22,6 +22,10 @@ const std = @import("std");
 const c = @import("c.zig");
 const util = @import("util.zig");
 
+const Server = @import("Server.zig");
+
+server: *Server,
+
 wlr_xdg_toplevel_decoration: *c.wlr_xdg_toplevel_decoration_v1,
 
 listen_destroy: c.wl_listener,
@@ -29,8 +33,11 @@ listen_request_mode: c.wl_listener,
 
 pub fn init(
     self: *Self,
+    server: *Server,
     wlr_xdg_toplevel_decoration: *c.wlr_xdg_toplevel_decoration_v1,
 ) void {
+    self.server = server;
+
     self.wlr_xdg_toplevel_decoration = wlr_xdg_toplevel_decoration;
 
     self.listen_destroy.notify = handleDestroy;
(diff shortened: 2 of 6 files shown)


river/Config.zig
```diff
@@ -73,8 +76,14 @@ pub fn init(self: *Self) !void {
     self.float_filter = std.ArrayList([*:0]const u8).init(util.gpa);
     errdefer self.float_filter.deinit();
 
+    self.csd_filter = std.ArrayList([*:0]const u8).init(util.gpa);
```
ifreund, line 79: This should just store a slice. (the float filter should as well but that'd be out of scope for this)
```suggestion
    self.csd_filter = std.ArrayList([]const u8).init(util.gpa);
```

river/Decoration.zig
```diff
@@ -49,8 +56,27 @@ fn handleDestroy(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
+    const wlr_xdg_surface: *c.wlr_xdg_surface = self.wlr_xdg_toplevel_decoration.surface;
+    const wlr_xdg_toplevel: *c.wlr_xdg_toplevel = @field(wlr_xdg_surface, c.wlr_xdg_surface_union).toplevel;
+    const app_id: [*:0]const u8 = if (wlr_xdg_toplevel.app_id) |id| id else "NULL";
+
+    var use_csd = false;
+    for (self.server.config.csd_filter.items) |filter_app_id| {
+        if (std.mem.eql(u8, std.mem.span(app_id), std.mem.span(filter_app_id))) {
+            use_csd = true;
+        }
+    }
```
ifreund, line 69: for/else would be cleaner here
```suggestion
    const use_csd = for (self.server.config.csd_filter.items) |filter_app_id| {
        if (std.mem.eql(u8, std.mem.span(app_id), filter_app_id)) break true;
    } else false;
```

river/XdgToplevel.zig
```diff
@@ -189,13 +189,26 @@ fn handleMap(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
+    // to use client side decorations, inform it that it is tiled. This
     // prevents firefox, for example, from drawing shadows around itself.
-    if (wlr_xdg_toplevel.parent == null)
-        _ = c.wlr_xdg_toplevel_set_tiled(
-            self.wlr_xdg_surface,
-            c.WLR_EDGE_LEFT | c.WLR_EDGE_RIGHT | c.WLR_EDGE_TOP | c.WLR_EDGE_BOTTOM,
-        );
+    if (wlr_xdg_toplevel.parent == null) {
+        var use_csd = false;
+        for (root.server.config.csd_filter.items) |filter_app_id| {
```
ifreund, line 197: Could be a for/else

river/XdgToplevel.zig
```diff
@@ -189,13 +189,26 @@ fn handleMap(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
+            if (std.mem.eql(u8, std.mem.span(app_id), std.mem.span(filter_app_id))) {
+                use_csd = true;
+            }
+        }
+
+        if (!use_csd) {
```
ifreund, line 203: I prefer to put the true branch first if there's an else instead of doing `if (!foo)`.

river/render.zig
```diff
@@ -88,7 +88,9 @@ pub fn renderOutput(output: *Output) void {
             if (view.focused) continue;
 
             renderView(output.*, view, &now);
-            renderBorders(output.*, view, &now);
+            if (view.draw_borders) {
+                renderBorders(output.*, view, &now);
+            }
```
ifreund, line 93: I'd personally put this on one line, feel free to ignore this one though, it's pretty arbitrary.
```suggestion
            if (view.draw_borders) renderBorders(output.*, view, &now);
```

river/Decoration.zig
```diff
@@ -49,8 +56,24 @@ fn handleDestroy(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
-        self.wlr_xdg_toplevel_decoration,
-        .WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE,
-    );
+
+    const wlr_xdg_surface: *c.wlr_xdg_surface = self.wlr_xdg_toplevel_decoration.surface;
+    const wlr_xdg_toplevel: *c.wlr_xdg_toplevel = @field(wlr_xdg_surface, c.wlr_xdg_surface_union).toplevel;
+    const app_id: [*:0]const u8 = if (wlr_xdg_toplevel.app_id) |id| id else "NULL";
+
+    const use_csd = for (self.server.config.csd_filter.items) |filter_app_id| {
+        if (std.mem.eql(u8, std.mem.span(app_id), std.mem.span(filter_app_id))) break true;
```
ifreund, line 65: Don't need the span here as it's already a slice
```suggestion
        if (std.mem.eql(u8, std.mem.span(app_id), filter_app_id)) break true;
```

river/XdgToplevel.zig
```diff
@@ -189,13 +189,23 @@ fn handleMap(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
+    // to use client side decorations, inform it that it is tiled. This
     // prevents firefox, for example, from drawing shadows around itself.
-    if (wlr_xdg_toplevel.parent == null)
-        _ = c.wlr_xdg_toplevel_set_tiled(
-            self.wlr_xdg_surface,
-            c.WLR_EDGE_LEFT | c.WLR_EDGE_RIGHT | c.WLR_EDGE_TOP | c.WLR_EDGE_BOTTOM,
-        );
+    if (wlr_xdg_toplevel.parent == null) {
+        const use_csd = for (root.server.config.csd_filter.items) |filter_app_id| {
+            if (std.mem.eql(u8, std.mem.span(app_id), std.mem.span(filter_app_id))) break true;
```
ifreund, line 197: same here.

PR #68, opened 2020-07-16
Implement filter commands.

Leon-Plickat:
This adds the two commands "add-csd-filter" and "add-float-filter" to populate the two filter lists.

This also contains a small bugfix for the CSD filter, which would not be respected if an XDG toplevel had a parent. I removed parent checking entirely, as river now has two atomic decoration states (SSD and CSD) and having windows be "in between" the states (by setting toplevels with parents to not tiled but still enforcing SSD, like it did before my previous PR) is not so great, IMO.


river/XdgToplevel.zig
```diff
@@ -189,22 +189,17 @@ fn handleMap(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
+    // decorations, inform it that it is tiled.
+    if (for (root.server.config.csd_filter.items) |filter_app_id| {
+        if (std.mem.eql(u8, std.mem.span(app_id), filter_app_id)) break true;
+    } else false) {
+        view.draw_borders = false;
+    } else {
+        _ = c.wlr_xdg_toplevel_set_tiled(
+            self.wlr_xdg_surface,
+            c.WLR_EDGE_LEFT | c.WLR_EDGE_RIGHT | c.WLR_EDGE_TOP | c.WLR_EDGE_BOTTOM,
+        );
```
ifreund, line 202:
```suggestion
    for (root.server.config.csd_filter.items) |filter_app_id| {
        if (std.mem.eql(u8, std.mem.span(app_id), filter_app_id)) {
            view.draw_borders = false;
            break;
        }
    } else {
        _ = c.wlr_xdg_toplevel_set_tiled(
            self.wlr_xdg_surface,
            c.WLR_EDGE_LEFT | c.WLR_EDGE_RIGHT | c.WLR_EDGE_TOP | c.WLR_EDGE_BOTTOM,
        );
```

river/command/filter.zig
```diff
@@ -0,0 +1,63 @@
...
+}
+
+pub fn addFloatFilter(
+    allocator: *std.mem.Allocator,
+    seat: *Seat,
+    args: []const []const u8,
+    out: *?[]const u8,
+) Error!void {
+    if (args.len < 2) return Error.NotEnoughArguments;
+    if (args.len > 2) return Error.TooManyArguments;
```
ifreund, line 40: If you're going to have the helper, move these checks to it as well for greater deduplication.

doc/riverctl.1.scd
```diff
@@ -17,6 +17,14 @@ used to control and configure river.
 
 ## ACTIONS
 
+*add-csd-filter* _app-id_
+	Add an app-id to the CSD filter list. Windows with this app-id are allowed
+	to use client side decoration instead of the default server side decoration.
+
+*add-float-filter* _app-id_
+	Add an app-id to the float filter list. Windows with this app-id will start
+	floating.
```
ifreund, line 26: Bikeshedding, but I'd prefer `csd-filter-add`/`float-filter-add` or maybe even just `csd`/`float`.

PR #74, opened 2020-07-29
cursor: implement resize

ifreund:
This currently works great when run from a tty, but is wacky albeit
somewhat functional when run in a nested session.


ifreund:
Fixed the funkiness with nested sessions and made things smoother when resizing slow to configure clients like firefox.

This is now blocked by some strange behavior triggered when toggling views between floating and fullscreen causing transaction timeouts.

Issue #79, opened 2020-08-03
Issues with frame prefection of transactions


ifreund:
Despite my best efforts, there are still a couple of edge cases where imperfect frames may be shown when modifying the state of views in the layout.

- Focus is not double buffered, which may cause border color to change out of sync with the layout. This is most noticeable when running the zoom command with the top visible view in the stack focused.
- When views are destroyed, they are destroyed immediately and not kept around until the resulting transaction is finished. This results in imperfect frames as there is a view missing from the layout during the transaction.

PR #80, opened 2020-08-03
view: double buffer focus, use counter not bool

ifreund:
- Double buffering focus state ensures that border color is kept in sync
with the transaction state of views in the layout.
- Using a counter instead of a bool will allow for proper handling of
multiple seats. This is done in the same commit to avoid more churn in
the future.


ifreund:
This currently breaks keyboard focus of xwayland views for whatever reason. However, it does solve the first issue mentioned in #79 which makes thing feel significantly smoother.

ifreund:
Issues with xwayland fixed.

PR #83, opened 2020-08-10
Fix bug causing river to crash when layouts define a too small view size

Leon-Plickat:
~~Currently, river allows resizing views to 0x0, with mouse and via layout. To small sizes may cause severe issues in some applications and a size of 0x0 is a protocol error. Some views, such as GTK ones, set constraints, however this is only an optional feature of the XDG-Shell. To prevent too small views, this PR enforces a minimal view size of 50x50 for both layouts and mouse resizing. This is a totally arbitrary number, low enough to not get in the way of the user but high enough to not cause issues in applications.~~

~~Additionally,~~ this PR fixes a crashing bug which is encountered when a layout defines a width or height smaller than `delta_size`.

### DIFF
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -269,8 +269,10 @@ fn layoutExternal(self: *Self, visible_count: u32) !void {
         var box = try parseBox(token);
         box.x += self.usable_box.x + xy_offset;
         box.y += self.usable_box.y + xy_offset;
-        box.width -= delta_size;
-        box.height -= delta_size;
+
+        if (box.width > delta_size) box.width -= delta_size;
+        if (box.height > delta_size) box.height -= delta_size;
+
         try view_boxen.append(box);
     }
 


Leon-Plickat:
> the minimum size should already be enforced through View.getConstraints()/View.applyConstraints(), which seems to be working for me at least.

Just read the code and yes, in theory it should do that, but not on my machine. At least with alacritty it does not work correctly. That means that either 1) alacritty does set constraint but they are different from the minimum window size it actually supports or 2) the bug is somewhere in river.

I'll investigate with a few more applications.

Leon-Plickat:
I tested it with the [wayland-demo-client](https://git.sr.ht/~sircmpwn/wayland-demo-client) which uses the XDG-Shell and does not set constraints. With that, the minimum size enforcement in `getConstraints()` does work. So apparently this is a bug in alacritty, not river.

It is possible for a client to set constraints smaller than the minimal window size, but I guess that is fine, as it is not the compositors job to protect clients from themselves (We should however try make sure river does not crash on stupid constraints). Therefore I will edit this PR to only contain the bugfix.

PR #84, opened 2020-08-10
Enforce minimum window size also for views with constraints

Leon-Plickat:


river/XdgToplevel.zig
```diff
@@ -132,9 +132,9 @@ pub fn getTitle(self: Self) [*:0]const u8 {
 pub fn getConstraints(self: Self) View.Constraints {
     const state = @field(self.wlr_xdg_surface, c.wlr_xdg_surface_union).toplevel.*.current;
     return .{
-        .min_width = if (state.min_width > 0) state.min_width else View.min_size,
+        .min_width = if (state.min_width > View.min_size) state.min_width else View.min_size,
```
ifreund, line 135: How about just using `max()`?
```suggestion
        .min_width = std.math.max(state.min_width, View.min_size),
```

Leon-Plickat: That is a better idea. I am just not too familiar with the standard library yet.

ifreund: That's totally understandable, not all that many people are :P

Issue #85, opened 2020-08-10
Imperfect frames due to race on child toplevel creation


ifreund:
For some reason, the `surface_box` of toplevels created with a parent is sometimes `undefined` during rendering. This is likely related to the fact that such views start floating by default. This happens quite rarely for me at least, I can usually reproduce it by repeatedly opening and closing a dialog in firefox.

ifreund:
I can no longer reproduce the crash, but I do see imperfect frames occasionally on opening a popup in firefox.

ifreund:
I can't reproduce this any more

PR #86, opened 2020-08-11
root: refactor transaction initiation

ifreund:
- require the caller to use Root.startTransaction() directly
- introduce View.applyPending() to unify logic
- introduce View.shouldTrackConfigure() to unify more logic
- update all callsites to intelligently rearrange only what is necessary


ifreund:
This seems to fix #85 for me. More testing is needed.

Issue #90, opened 2020-08-17
[Feature request] Toggle add to top/add to bottom


BenEvolent333:
I think it would be useful to be able to toggle whether new windows are added to the top or the bottom of the stack. I made a layout which shows the top windows from the stack at the bottom, and puts the bottom windows in the master area, but this does have different functionality (i.e. swapping windows to the master now puts them at the bottom).

ifreund:
This sounds like a reasonable feature to me, there are in fact a few more potentially useful behaviors for where to create new windows as well.

As #12 already includes this, I'm going do go ahead and close this as a duplicate.

PR #91, opened 2020-08-17
Implement "attach-mode"

Leon-Plickat:
Solves #90 

More modes (like "below focus" or "above focus") can be easily added in a future commit.

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -95,6 +95,9 @@ that tag 1 through 9 are visible.
 
 ## CONFIGURATION COMMANDS
 
+*attach-mode* *top*|*bottom*
+	Configure where new views should attach in the view stack for the currently focused output.
+
 *background-color* _#RRGGBB_|_#RRGGBBAA_
 	Set the background color.
 
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -30,12 +30,15 @@ const LayerSurface = @import("LayerSurface.zig");
 const Root = @import("Root.zig");
 const View = @import("View.zig");
 const ViewStack = @import("view_stack.zig").ViewStack;
+const AttachMode = @import("view_stack.zig").AttachMode;
 const OutputStatus = @import("OutputStatus.zig");
 
 const State = struct {
     /// A bit field of focused tags
     tags: u32,
 };
+
+attach_mode: AttachMode,
 
 root: *Root,
 wlr_output: *c.wlr_output,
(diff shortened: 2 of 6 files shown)


river/Output.zig
```diff
@@ -107,6 +114,8 @@ pub fn init(self: *Self, root: *Root, wlr_output: *c.wlr_output) !void {
 
     self.layout = try std.mem.dupe(util.gpa, u8, "full");
 
+    self.attach_mode = AttachMode.AttachTop;
```
ifreund, line 117: can use an enum literal here
```suggestion
    self.attach_mode = .top;
```

river/command/attach_mode.zig
```diff
@@ -0,0 +1,42 @@
...
+    allocator: *std.mem.Allocator,
+    seat: *Seat,
+    args: []const []const u8,
+    out: *?[]const u8,
+) Error!void {
+    if (args.len < 2) return Error.NotEnoughArguments;
+    if (args.len > 2) return Error.TooManyArguments;
+
+    if (std.mem.eql(u8, "top", args[1])) {
+        seat.focused_output.attach_mode = Output.AttachMode.AttachTop;
```
ifreund, line 36: Here as well, can also drop the `Output` import.
```suggestion
        seat.focused_output.attach_mode = .top;
```

doc/riverctl.1.scd
```diff
@@ -95,6 +95,9 @@ that tag 1 through 9 are visible.
 
 ## CONFIGURATION COMMANDS
 
+*attach-mode* *top*|*bottom*
+	Configure where new views should attach into the view stack.
```
ifreund, line 99:
```suggestion
	Configure where new views should attach to the view stack for the currently focused output.
```

Leon-Plickat:
Maybe it would make more sense to have the mode switching be part of the `viewstack.push()` function instead of different kinds of having multiple push functions and switching every time  a view is attached?

ifreund:
> Maybe it would make more sense to have the mode switching be part of the viewstack.push() function instead of having multiple push functions and switching every time a view is attached?

There are other places that `ViewStack.push()` is used (e.g. the zoom command or the focus stack of seats). Putting this logic in a new `ViewStack.attach()` function would make sense though imo.

river/command/attach_mode.zig
```diff
@@ -0,0 +1,42 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const util = @import("../util.zig");
+
+const Error = @import("../command.zig").Error;
+const Seat = @import("../Seat.zig");
+const Output = @import("../Output.zig");
```
ifreund, line 24:
```suggestion
```

Leon-Plickat:
Renamed `viewstack.push_bottom()` to `viewstack.append()` which is a better name IMO.

Also added `viewstack.attach()`, which now handles the switch. The idea is to call `.attach()` when the way a node is attached can change during runtime and to use `.push()` and the other specific attach methods when the mode is hardcoded anyway for whatever reason (as with the zoom command).

Issue #98, opened 2020-09-14
More advanced map 


maringuu:
I suggest implementing some mapping features from sway.

https://github.com/swaywm/sway/blob/afa890e8e9f10667a56e896a114bf81fbc3ff54a/sway/sway.5.scd#L384

`--no-repeat --locked --release` can be implemented right now I think. `--to-code --input-device` might require some more work.

I'll see how far I can get with this one tomorrow.

ifreund:
I'd hesitate to implement features just because sway has them. We should make sure that new features have a real use-case if we want river to stay simple. Is there a specific use case you have that these new features would address.

maringuu:
I have a use cases for `--no-repeat` and `--release`. It is rather gimmiky but still.

```
riverctl map normal $mod F toggle-fullscreen
riverctl map --release normal $mod F toggle-fullscreen
```
Often I see my self fullscreening a window for just several seconds. Also I sometimes want to have a quick look at the windows in the layout that are not currently show because there is a fullscreened window.
This could of course be archived with pressing `$mod F` twice but it would feel cooler if I could just hold it for a few seconds.
Also pausing music for some seconds would be cool.

One use case for `--locked` when I leave my laptop (locked) in a room with my friends. When they feel very funny they might exit river and thus close all my windows because the mapping got executed.

For the others I don't have a concrete usecase for my self.

maringuu:
I just realised that commands are only executed once and not repeated.
A use case for repeating them would be turning up/down brightness or volume.

ifreund:
> I just realised that commands are only executed once and not repeated.
A use case for repeating them would be turning up/down brightness or volume.
Maybe we should add (if desired) --repeat instead.

Agreed, I think a `--repeat` option would be preferable to a `--no-repeat`

> One use case for --locked when I leave my laptop (locked) in a room with my friends. When they feel very funny they might exit river and thus close all my windows because the mapping got executed.

I think we should disable all keybindings by default when input is inhibited (e.g. by a lockscreen). We could then consider adding a `--locked` option to bypass this restriction which might be desirable for e.g. music player controls (which is how sway's works as well iirc).

`--release` should be fairly trivial to implement and I can see it being useful so I wouldn't mind having that.

ifreund:
Oh, and since these are the first options we are adding to a command, let's make sure we establish good conventions. I think I'd like to take a page out of kakoune's book and use only a single `-` to prefix flags as I don't think we should support "short" flags. This would give us for example:
```
[...]

PR #99, opened 2020-09-15
Implement map -release

maringuu:
This implements `-release` from #98.
It also paves the way to implement more optional arguments for map which I plan doing.
~One controversial thing might be moving up the allocation of `Mapping` in `map.zig`. I think it is acceptable because it will only occur when a duplicate mapping would be assigned and this should not happen often. The benefit is the `isEqual` function in `Mapping` which IMO improves readablity.~

To test you can use:
```
riverctl map normal Mod4 F toggle-fullscreen
riverctl map -release normal Mod4 F toggle-fullscreen
```

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -123,9 +123,11 @@ that tag 1 through 9 are visible.
 	but let the currently focused window in focus.
 	When set to _strict_ this is not the case. The focus will be updated on every cursor movement.
 
-*map* _mode_ _modifiers_ _key_ _command_
+*map* [-release] _mode_ _modifiers_ _key_ _command_
 	_mode_ is either “normal” (the default mode) or a mode created with
-	*declare-mode*. _modifiers_ is a list of one or more of the following
+	*declare-mode*.
+	If _-release_ is specified the mapping is executed on key release rather than key press.
+	_modifiers_ is a list of one or more of the following
 	modifiers separated with a plus sign:
 
 	- Shift
--- a/river/Mapping.zig
+++ b/river/Mapping.zig
@@ -25,10 +25,14 @@ keysym: c.xkb_keysym_t,
 modifiers: u32,
 command_args: []const []const u8,
 
+/// When set to true the mapping will be executed on key release rather than on press
+release: bool,
+
 pub fn init(
     allocator: *std.mem.Allocator,
     keysym: c.xkb_keysym_t,
     modifiers: u32,
+    release: bool,
     command_args: []const []const u8,
 ) !Self {
     const owned_args = try allocator.alloc([]u8, command_args.len);
@@ -40,6 +44,7 @@ pub fn init(
     return Self{
         .keysym = keysym,
         .modifiers = modifiers,
+        .release = release,
         .command_args = owned_args,
     };
 }
--- a/river/command/map.zig
+++ b/river/command/map.zig
@@ -50,20 +50,23 @@ pub fn map(
     args: []const []const u8,
     out: *?[]const u8,
 ) Error!void {
-    if (args.len < 5) return Error.NotEnoughArguments;
+    const optionals = parseOptionalArgs(args[1..]);
+    // offset caused by optional arguments
+    const offset = optionals.i;
+    if (args.len - offset < 5) return Error.NotEnoughArguments;
 
-    const mode_id = try modeNameToId(allocator, seat, args[1], out);
-    const modifiers = try parseModifiers(allocator, args[2], out);
+    const mode_id = try modeNameToId(allocator, seat, args[1 + offset], out);
+    const modifiers = try parseModifiers(allocator, args[2 + offset], out);
 
     // Parse the keysym
-    const keysym_name = try std.cstr.addNullByte(allocator, args[3]);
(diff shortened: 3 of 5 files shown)


doc/riverctl.1.scd
```diff
@@ -116,9 +116,11 @@ that tag 1 through 9 are visible.
 *enter-mode* _name_
 	Switch to given mode if it exits.
 
-*map* _mode_ _modifiers_ _key_ _command_
+*map* _mode_ [-release] _modifiers_ _key_ _command_
```
ifreund, line 119: Should this not be 
```suggestion
*map* [-release] _mode_ _modifiers_ _key_ _command_
```

river/Mapping.zig
```diff
@@ -48,3 +53,13 @@ pub fn deinit(self: Self, allocator: *std.mem.Allocator) void {
...
+
+/// Returns whether the mapping matches the keysym, modifiers and release status
+pub fn match(self: *const Self, keysym: c.xkb_keysym_t, modifiers: u32, released: bool) bool {
+    return modifiers == self.modifiers and keysym == self.keysym and released == self.release;
+}
+
+/// Returns whether the mappings are euqal
+pub fn isEqual(self: *const Self, other: *const Self) bool {
+    return other.modifiers == self.modifiers and other.keysym == self.keysym;
+}
```
ifreund, line 65: I'm not convinced that these functions are a net win for readability. Both are only called in one place and both require the reader to check the function definition to see exactly what is being compared.

(Aside on zig style, these should take `self: Self` as you don't need the pointer, the compiler will optimize this to a pass-by-pointer if it likes)

river/command/map.zig
```diff
@@ -53,35 +53,41 @@ pub fn map(
     if (args.len < 5) return Error.NotEnoughArguments;
```
ifreund, line 53: This check is no longer sufficient.

river/command/map.zig
```diff
@@ -53,35 +53,41 @@ pub fn map(
...
 
+    const mapping = try Mapping.init(util.gpa, keysym, modifiers, optionals.release, args[4..]);
+    errdefer mapping.deinit(util.gpa);
+
     // Check if the mapping already exists
     const mode_mappings = &seat.input_manager.server.config.modes.items[mode_id].mappings;
-    for (mode_mappings.items) |existant_mapping| {
-        if (existant_mapping.modifiers == modifiers and existant_mapping.keysym == keysym) {
+    for (mode_mappings.items) |*existant_mapping| {
+        if (mapping.isEqual(existant_mapping)) {
```
ifreund, line 80: Why not allow bindings for both press and release of a given key?

river/command/map.zig
```diff
@@ -176,3 +182,36 @@ fn parseModifiers(allocator: *std.mem.Allocator, modifiers_str: []const u8, out:
     }
     return modifiers;
 }
+
+// TODO remove once https://github.com/ziglang/zig/issues/447 is implemented
```
ifreund, line 186: you can already declare an anonymous struct as the return type of the function if you like:
```zig
fn parseOptionalArgs(args: []const []const u8) struct {
    i: usize,
    release: bool,
} {
   ...
}
```
and then use an anon struct literal for the return: 
```zig
return .{ .i = 42, .release = false };
```
We'll likely want to make this more generic in order to avoid code duplication as more options are added to commands in the future though.

river/command/map.zig
```diff
@@ -176,3 +182,36 @@ fn parseModifiers(allocator: *std.mem.Allocator, modifiers_str: []const u8, out:
...
+        .release = false,
+    };
+
+    var i: usize = 0;
+    for (args) |arg| {
+        if (std.mem.eql(u8, arg, "-release")) {
+            parsed.release = true;
+            i += 1;
+        } else {
+            // Break if no argument more argument is recogniced
```
ifreund, line 210:
```suggestion
            // Break if the arg is not an option
```

ifreund (changes requested):

maringuu:
Sorry I wanted to change the title to [WIP] about an hour ago when I realised some things are not right so you reviewed some of my old code :/
I'll write again when I fixed your commends and pushed the new code.

doc/riverctl.1.scd
maringuu: Yes. Somehow I thought in the beginning that it would be a good idea to have it there.

river/Mapping.zig
maringuu: Yeah.
Changed them back.

river/command/map.zig
maringuu: Oh, I missed adding the check for release.

river/command/map.zig
maringuu: I'll remove the comment and leave it like it is. When other commands get arguments too we could thing of a more generic solution.

river/command/map.zig
maringuu: Changed.

maringuu:
Everything should be fixed now.

ifreund (commented):
Code looks fine, could you resolve that conflict in the man page so I can hit the merge button?

maringuu:
Done.

ifreund (approved):
Thanks!

PR #104, opened 2020-09-17
Optional args parsing

### LINKED ISSUE #102: Brainstorm parsing optional arguments
For #101 to be implemented we could think about a nice way to parse optional arguments. After some brainstorming I though of maybe using an `enum` and an option_mask which specifies options in the same way that tagmask specifies tags. This enum would be generated in the file which implements the command.
This enum could be generated from a nice descriptive list of options and their default value.
This could look something like this: 
`map.zig`
```zig
const OptionalArgs: enum(u32) = //Generated somehow.

const arglist = [_]struct {
    name: []const u8,
    default_value: bool, // Optionals can only be true or false
    name_in_enum: []u8, // The names of the entries in the enum have to come somewhere 
                    // Perhabs we could generate them from name
                    //but this would make it tricky to have commands that contain illegal characters for zig enum members
}{... // The args specified
}
```

Edit: The default value will be needed in the parsing functions which needs to determine what should be set when the option is not passed.
What are your thoughts and ideas about this?

maringuu:
I've implemented your suggestion from https://github.com/ifreund/river/issues/102#issuecomment-693645836.

There are several things left to be discussed and solved.
I've implemented options with an Int argument too to figure out a good way to parse the argument.
When merging this I will of course use this for `map -release` and remove the Int option as it is not needed.

### Default Options for bools
Since boolean options are either given or not given I'd like to make the default be false always and change them to true when the argument is given. 
A possibility would be to make it a compile error to have the default value set to false. This is what I implemented currently as it seems clean and explicit about the default beeing false.
Another idea would be to make giving the option toggle the default value. Code wise this would be the simplest but I really dislike this one as there is really no use for that imo.

### Enforce naming of arguments
Since we decided to always have arguments be in the form `-long-option` we could implement a check to enforce correct naming at compile time. Is there anything bad about this?

### Setting the offset
The last thing I'm not happy with is setting the offset. The offset is needed for the caller to know where to continue parsing.
If there was a way to run an else branch after a for loop even when the for loop breaks it would be nice to set it there.
There was a issue in the zig bugtracker somewhere but I can't seem to find it.

Closes #102

### DIFF
--- a/river/command.zig
+++ b/river/command.zig
@@ -142,3 +142,52 @@ pub fn errToMsg(err: Error) [:0]const u8 {
         Error.Other => unreachable,
     };
 }
+
+/// Parses a list of options specified by T
+/// args[0] must be the first option if any
+/// offset will be set to the index of the first entry in args that is not an option
+/// See test "optional args" for an example of T
+pub fn parse(args: []const []const u8, offset: *usize, comptime T: type) T {
+    var ret: T = T{};
+    // Make sure the format of T is valid
+    if (@typeInfo(T) != .Struct) @compileError("T must be a struct");
+    inline for (@typeInfo(T).Struct.fields) |field| {
+        if (@typeInfo(field.field_type) != .Bool)
+            @compileError("unsupported option type " ++ field.field_type);
+        if (field.default_value.?)
+            @compileError("bool options must be false by default");
+    }
+
+    offset.* = 0;
+    // Do the actual parsing of args
+    outer: for (args) |arg| {
+        inline for (@typeInfo(T).Struct.fields) |field| {
+            if (std.mem.eql(u8, arg, field.name)) {
+                @field(ret, field.name) = true;
+                offset.* += 1;
+                continue :outer;
+            }
+        }
+
+        // There are no options left
+        break;
+    }
+
+    return ret;
+}
+
+test "optional args" {
+    const testing = @import("std").testing;
+    const mem = @import("std").mem;
+
+    var offset: usize = 0;
+    const args = [_][]const u8{ "-bool-given", "not_an_option" };
+    const options = parse(&args, &offset, struct {
+        @"-bool-given": bool = false,
+        @"-bool-not-given": bool = false,
+    });
+
+    testing.expect(options.@"-bool-given" == true);
+    testing.expect(options.@"-bool-not-given" == false);
+    testing.expect(mem.eql(u8, args[offset], "not_an_option"));
+}
--- a/river/command/map.zig
+++ b/river/command/map.zig
@@ -19,6 +19,7 @@
 const std = @import("std");
 
 const c = @import("../c.zig");
+const command = @import("../command.zig");
 const util = @import("../util.zig");
 
 const Error = @import("../command.zig").Error;
@@ -51,9 +52,11 @@ pub fn map(
     args: []const []const u8,
     out: *?[]const u8,
 ) Error!void {
-    const optionals = parseOptionalArgs(args[1..]);
+    var offset: usize = undefined;
+    const options = command.parse(args[1..], &offset, struct {
+        @"-release": bool = false,
+    });
     // offset caused by optional arguments
-    const offset = optionals.i;
     if (args.len - offset < 5) return Error.NotEnoughArguments;
 
     const mode_id = try modeNameToId(allocator, seat, args[1 + offset], out);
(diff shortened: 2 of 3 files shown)


maringuu:
The newest commit somehow does not compile. I don't see why atm.

river/command.zig
```diff
@@ -117,3 +117,71 @@ pub fn errToMsg(err: Error) [:0]const u8 {
...
+        if (@typeInfo(field.field_type) != .Bool) continue;
+        if(field.default_value.?) @compileError("bool options must be false by default");
+
+    }
+    // TODO cleaner way to set offset
+    var i: usize = 0;
+    defer offset.* = i;
+
+    outer: for (args) |arg, idx| {
+        i = idx + 1;
```
ifreund, line 142: Pretty sure this is the why the tests are crashing, should just be `i = idx`

river/command.zig
```diff
@@ -117,3 +117,71 @@ pub fn errToMsg(err: Error) [:0]const u8 {
...
+            if (std.mem.eql(u8, arg, field.name)) {
+                found = true;
+                switch (@typeInfo(field.field_type)) {
+                    .Bool => {
+                        @field(ret, field.name) = true;
+                    },
+                    .Int => {
+                        if (args.len < idx + 2) return error.MissingArgument;
+                        @field(ret, field.name) = try std.fmt.parseInt(field.field_type, args[idx+1], 10);
+                        continue :outer;
```
ifreund, line 154: This is insufficient as the next iteration must be skipped as well, but it doesn't matter as we don't need to parse ints yet :)

ifreund (commented):
> ...and remove the Int option as it is not needed.

+1, I'm starting to question if there will be ever be a need for something beyond bools tbh.

> Since boolean options are either given or not given I'd like to make the default be false always and change them to true when the argument is given.

This sounds reasonable to me.

river/command.zig
maringuu: I changed it to just count the parsed arguments by adding one each time.

river/command.zig
maringuu: Yeah. Removed it.

maringuu:
I removed the int parsing now and changed `map.zig` to use this parsing.
In `command.zig` line 136 there is a todo left which is about enforcing the names of parameters to match a specific format.

river/command.zig
```diff
@@ -117,3 +117,59 @@ pub fn errToMsg(err: Error) [:0]const u8 {
...
+                found = true;
+                @field(ret, field.name) = true;
+            }
+        }
+        // stop parsing options when no options are left
+        if (!found) break;
+        i += 1;
+    }
+
+    offset.* = i;
```
ifreund, line 157: This loop can be written a little more cleanly:
```suggestion
    offset.* = 0;
    for (args) |arg| {
        inline for (@typeInfo(T).Struct.fields) |field| {
            if (std.mem.eql(u8, arg, field.name)) {
                @field(ret, field.name) = true;
                offset.* += 1;
                continue;
            }
        }
        // There are no options left
        break;
    }
```

river/command.zig
```diff
@@ -117,3 +117,59 @@ pub fn errToMsg(err: Error) [:0]const u8 {
...
+    if (@typeInfo(T) != .Struct) @compileError("T must be a struct");
+    inline for (@typeInfo(T).Struct.fields) |field| {
+        if (@typeInfo(field.field_type) != .Bool)
+            @compileError("unsupported option type " ++ field.field_type);
+        if (field.default_value.?)
+            @compileError("bool options must be false by default");
+        if (field.name[0] != '-')
+            // TODO Should we really have this check? If yes test ascii.isAlpha(field.name[1])
+            // Figure out how long arguments should be is -a a valid argument
+            @compileError("option names must start with a single - followed by an alphabetical cahr");
```
ifreund, line 138: I'd say just drop this check. Options not conforming to the naming convention will be fairly easy to catch in review.

river/command/map.zig
```diff
@@ -74,7 +77,9 @@ pub fn map(
     // Check if the mapping already exists
     const mode_mappings = &seat.input_manager.server.config.modes.items[mode_id].mappings;
     for (mode_mappings.items) |existant_mapping| {
-        if (existant_mapping.modifiers == modifiers and existant_mapping.keysym == keysym and existant_mapping.release == optionals.release) {
+        if (existant_mapping.modifiers == modifiers
+            and existant_mapping.keysym == keysym
+            and existant_mapping.release == @field(options, "-release")) {
```
ifreund, line 82: Pretty sure you can just do this here:
```suggestion
            and existant_mapping.release == options.@"-release") {
```
`@field()` is only needed if the name isn't static at comptime.

maringuu:
Sorry I was busy and didn't have time for this pr.
It should be finished now.

maringuu:
Oops i accidently closed it

maringuu:
I dont know how I keep closing this over and over again.

maringuu:
Rebased.

maringuu:
Rebased again. It is now used for `unmap` too.

maringuu:
This in neat and all but I've come to the conclusion that we dont yet need a fancy way to do parsing of arguments.

PR #107, opened 2020-10-01
Here it is.

SamHSmith:
Easy to put in. Easy to rip out.

### DIFF
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -95,15 +95,41 @@ pub fn deinit(self: Self) void {
     self.csd_filter.deinit();
 }
 
-pub var focused_view_opacity_delta : f32 = 0.0;
+// These can be set via riverctl
+pub var focus_minimum_update_time : f32 = 0.0;
+pub var focused_view_opacity_speed : f32 = 1.0;
+pub var unfocused_view_opacity_speed : f32 = 1.0;
+pub var focused_view_opacity_target : f32 = 1.0;
+pub var unfocused_view_opacity_target : f32 = 1.0;
+
 var _last_get_time : u64 = 0;
-pub var focused_view_opacity : f32 = 1.0;
-pub fn get_focused_view_opacity() f32 {
+pub var focused_view_opacity : f32 = 1.0; //this one too
+
+pub fn update_focused_view_opacity() void {
     var t : u64 = std.time.milliTimestamp();
     var d = @floatCast(f32, @intToFloat(f64, t - _last_get_time) / 1000.0);
     _last_get_time = t;
-    focused_view_opacity += d * focused_view_opacity_delta;
+    var move = d * 
+        std.math.copysign(f32, focused_view_opacity_speed, focused_view_opacity_target - focused_view_opacity);
+    focused_view_opacity += move;
+    if (move > 0.0 and focused_view_opacity > focused_view_opacity_target) { focused_view_opacity = focused_view_opacity_target; }
+    if (move < 0.0 and focused_view_opacity < focused_view_opacity_target) { focused_view_opacity = focused_view_opacity_target; }
     if (focused_view_opacity > 1.0) { focused_view_opacity = 1.0; }
     if (focused_view_opacity < 0.0) { focused_view_opacity = 0.0; }
-    return focused_view_opacity;
+}
+
+var _last_get_time2 : u64 = 0;                                         
+pub var unfocused_view_opacity : f32 = 1.0; //this one too
+
+pub fn update_unfocused_view_opacity() void {
+    var t : u64 = std.time.milliTimestamp();
+    var d = @floatCast(f32, @intToFloat(f64, t - _last_get_time2) / 1000.0);
+    _last_get_time2 = t;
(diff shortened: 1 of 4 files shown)


river/command/focused_view_opacity.zig
```diff
@@ -0,0 +1,56 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 Isaac Freund
```
Leon-Plickat, line 3: You should probably put _your_ name here.

river/command/focused_view_opacity.zig
```diff
@@ -0,0 +1,56 @@
...
+pub fn opacity(
+    allocator: *std.mem.Allocator,
+    seat: *Seat,
+    args: []const []const u8,
+    out: *?[]const u8,
+) Error!void {
+    if (args.len < 2) return Error.NotEnoughArguments;
+    if (args.len > 2) return Error.TooManyArguments;
+
+    const op = std.fmt.parseFloat(f32, args[1]);
```
Leon-Plickat, line 34: You probably should use `try`.

Leon-Plickat: Also I think river uses `f64` everywhere, but I am not sure...

river/command/focused_view_opacity.zig
```diff
@@ -0,0 +1,56 @@
...
+    if (op) |num| {
+        Config.focused_view_opacity = num;
+    } else |err| {
+        return Error.InvalidFloat;
+    }
```
Leon-Plickat, line 39: All this would not be necessary with `try`, see `river/command/mod_master_factor.zig` for reference, were `parseFloat()` is also used.

river/command/focused_view_opacity.zig
```diff
@@ -0,0 +1,56 @@
...
+pub fn delta(
+    allocator: *std.mem.Allocator,
+    seat: *Seat,
+    args: []const []const u8,
+    out: *?[]const u8,
+) Error!void {
+    if (args.len < 2) return Error.NotEnoughArguments;
+    if (args.len > 2) return Error.TooManyArguments;
+
+    const del = std.fmt.parseFloat(f32, args[1]);
```
Leon-Plickat, line 50: Same as above.

Also `del` sounds like "delete" and River uses `delta` in other places, so you should stick to convention here.

river/Config.zig
```diff
@@ -94,3 +94,16 @@ pub fn deinit(self: Self) void {
     self.float_filter.deinit();
     self.csd_filter.deinit();
 }
+
+pub var focused_view_opacity_delta : f32 = 0.0;
+var _last_get_time : u64 = 0;
+pub var focused_view_opacity : f32 = 1.0;
+pub fn get_focused_view_opacity() f32 {
+    var t : u64 = std.time.milliTimestamp();
```
Leon-Plickat, line 102: Really don't like this.

This function depends on river rendering multiple times a second. That will change in the future when river switches to damage based rendering.

The proper way would be to attach a timer to the event loop.

Leon-Plickat (changes requested):

river/command/focused_view_opacity.zig
SamHSmith: wlroots exposes alpha as f32. So I carried that through to everywhere.

Leon-Plickat: I see. In that case this makes sense.

river/Config.zig
SamHSmith: Timer feels way to heavy. I think a better solution would be to provide a is_opacity_changing() function. Then in the main loop we could damage the focused window if the opacity is changing. Then everything would still work and we only render multiple times a second when we actually are changing the opacity no?

Leon-Plickat: I do not think that is a good solution. Think of screens with a high refresh rate. A timer ensures that we change the opacity a specific amount of time.

ifreund: >  Timer feels way to heavy. I think a better solution would be to provide a is_opacity_changing() function. Then in the main loop we could damage the focused window if the opacity is changing. Then everything would still work and we only render multiple times a second when we actually are changing the opacity no?

As leon says, the proper way to do this is to use a timer. There is no "main loop," code is only run in reaction to events.

SamHSmith: Okay, I will move it too a timer. However, how do I make a timer that runs at the refresh rate of the current focused output?

Leon-Plickat: You don't. Animations like fading usually work by having to values: `delta_x` and `delta_t`. The first is the difference between the "steps", the second is the time between them. `delta_t` should not be equal to the refresh rate. That would be total overkill, way beyond human perception. I doubt more than six steps per second are needed.

Leon-Plickat: (You will want to look at the wl_event_loop documentation. AFAIK it has a nice function to insert a generic timer.)

ifreund:
Thanks for the PR, this has been superseded by #111 though.

PR #108, opened 2020-10-02
[WIP] draft river-layout-unstable-v1.xml protocol

Leon-Plickat:
TODO:
* [x] How to send view states?
* [x] How to change view dimensions?
* [ ] Implementation
   * [ ] River
   * [ ] Rivertile
       * [ ] `--inner-gaps` and `--outer-gaps` because gaps will be implemented in the layout client.
   * [x] Simple Proof of Concept client in C: [rilay](https://github.com/Leon-Plickat/rilay)
* [ ] Cleanup commits

---

With the `ack_layout_request` event, multiple layout clients can run side by side. River uses the one which send the ack first and ignores the others. Idea: maybe people want to implement different layouts as different clients.

The serial in every request / event is needed because a layout client may be tasked with generating layout for multiple monitors simultaneously.

---

Converting rivertile to a wayland client it probably trivial, but I am not sure yet how to approach this in river itself.

### DIFF
--- a/protocol/river-layout-unstable-v1.xml
+++ b/protocol/river-layout-unstable-v1.xml
@@ -1,0 +1,204 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="river_layout_unstable_v1">
+  <copyright>
+    Copyright © 2020 Leon Henrik Plickat
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+    MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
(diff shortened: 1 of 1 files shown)


Leon-Plickat:
I think I came up with a protocol that needs no additional roundtrip or any copied data.

I imagine it works like this:

* River sends a `layout_request` event to all layout clients
* River advertises all visible views (by index) on the output with a `advertise_view` event, after which `advertise_done` is send.
* If a client wants to generate a layout, it can send a `ack_layout_request` request. The first client which does so gains the "layout authority" and all other clients will simply be ignored.
* The client can then send `set_view_dimensions` requests to change the dimensions of a view, by index.
* When it is done, it sends the `layout_commit` request.

Leon-Plickat:
A simple layout client would work like this:

* Receive `layout_request`, send `ack_layout_request`
* In event handler function calculate layout, use `set_view_dimensions` for every possible view index, send `layout_commit`
* (Ignore anything else)

An advanced layout client would work like this:
* Receive `layout_request`, decide whether it want to gen. a layout, if yes send `ack_layout_request` and add an entry to some sort of "pending layout" list (by serial).
* Wait for every `advertise_view` event, possibly make some decisions based on their states.
* Wait for `advertise_done` event, use `set_view_dimensions` for every view index it wants to send, send `layout_commit`
* Possibly wait for `layout_status` and make some decisions based on that.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,154 @@
...
+        view_amount. However the views may be advertised in a random order.
+
+        This event is raised whether the layout request has been ack'ed or not.
+      </description>
+      <arg name="index" type="uint" summary="index of view in stack"/>
+      <arg name="tag_set" type="uint" summary="bit field, active tags of view"/>
+      <arg name="app_id" type="string" summary="view app-id"/>
+      <arg name="title" type="string" summary="view title"/>
+      <arg name="x" type="uint" summary="current x coordinate of view"/>
+      <arg name="y" type="uint" summary="current y coordinate of view"/>
```
ifreund, line 104: These should probably be `int`s not `uints`.

PR #111, opened 2020-10-04
#107, but with timers and double buffering

Leon-Plickat:
As the title states, this does what #107 does, but with timers. The way it is implemented also gives us a "fade in" effect on view map for free by setting the initial opacity to `0.0` (obviously configurable).

* [X] Implemented using per view opacity
* [X] (Optional) Fade effect between opacities
    * [X] Timers
    * [X] Avoids unnecessary transitions (example: no need to transition from 1.0 to 1.0)
    * [X] Animations can be disabled
    * [X] Each view has its own independent animation
    * [X] Supports changing target opacity mid-animation
    * [X] Animations are implemented in a clear and simple fashion with no magic
* [x] Opacity is double buffered
* [x] Everything is documented

I'll clean up the commits after the review.

---

For the animation I decided to use step-size instead of step-amount. This way the "speed" of the animation is constant instead of the time, which Feels Nicer™,  and is considerably easier to implement, especially w.r.t. changing the target opacity mid-animation. Also, looking at other project, this method of implementing fade effects appears to be more common.

### DIFF
--- a/contrib/config.sh
+++ b/contrib/config.sh
@@ -100,3 +100,5 @@ riverctl float-filter-add "popup"
 # Set app-ids of views which should use client side decorations
 riverctl csd-filter-add "gedit"
 
+# Set opacity and fade effect
+# riverctl opacity 1.0 0.75 0.0 0.1 20
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -66,6 +66,21 @@ csd_filter: std.ArrayList([]const u8),
 /// The selected focus_follows_cursor mode
 focus_follows_cursor: FocusFollowsCursorMode = .disabled,
 
+/// The opacity of the focused view
+view_opacity_focused: f32 = 1.0,
+
+/// The opacity of unfocused views
+view_opacity_unfocused: f32 = 1.0,
+
+/// The starting opacity of new views
+view_opacity_initial: f32 = 1.0,
+
+/// View opacity transition step
+view_opacity_delta: f32 = 1.0,
+
+/// Time between view opacity transition steps in msec
+view_opacity_delta_t: u31 = 20,
+
 pub fn init() !Self {
     var self = Self{
         .mode_to_id = std.StringHashMap(usize).init(util.gpa),
--- a/river/View.zig
+++ b/river/View.zig
@@ -1,6 +1,7 @@
 // This file is part of river, a dynamic tiling wayland compositor.
 //
 // Copyright 2020 Isaac Freund
+// Copyright 2020 Leon Henrik Plickat
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
@@ -65,6 +66,9 @@ const State = struct {
 
     float: bool = false,
     fullscreen: bool = false,
+
+    /// Opacity the view is transitioning to
+    target_opacity: f32,
 };
 
 const SavedBuffer = struct {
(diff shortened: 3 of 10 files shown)


river/View.zig
```diff
@@ -65,6 +66,12 @@ const State = struct {
 
     float: bool = false,
     fullscreen: bool = false,
+
+    /// The opacity of this view
+    opacity: f32 = 0.0,
```
ifreund, line 71: Don't use a default value if initializing this to something else.
```suggestion
    opacity: f32,
```

ifreund (changes requested):
Initial impressions playing around with the feature are good, it feels nice to have unfocused views slightly faded. 

Code could use a bit of cleanup, but the general approach is sound.

river/View.zig
```diff
@@ -379,3 +402,73 @@ pub fn unmap(self: *Self) void {
 
     root.startTransaction();
 }
+
+/// Change the opacity of a view by config.view_opacity_delta.
+/// If the target opacity was reached, return true.
+fn incrementOpacity(self: *Self) bool {
+    // TODO damage view when implementing damage based rendering
+    const config = self.output.root.server.config;
```
ifreund, line 410: Not sure if zig is smart enough to elide this copy yet.
```suggestion
    const config = &self.output.root.server.config;
```

river/View.zig
```diff
@@ -379,3 +402,73 @@ pub fn unmap(self: *Self) void {
...
+
+/// Destroy a views opacity timer
+fn killOpacityTimer(self: *Self) void {
+    if (c.wl_event_source_remove(self.opacity_timer) < 0) unreachable;
+    self.opacity_timer = null;
+}
+
+/// Set the timeout on a views opacity timer
+fn armOpacityTimer(self: *Self) void {
+    const delta_t = @intCast(c_int, self.output.root.server.config.view_opacity_delta_t);
```
ifreund, line 430:
```suggestion
    const delta_t = self.output.root.server.config.view_opacity_delta_t;
```

river/View.zig
```diff
@@ -379,3 +402,73 @@ pub fn unmap(self: *Self) void {
...
+}
+
+/// Create an opacity timer for a view and arm it
+fn attachOpacityTimer(self: *Self) void {
+    const server = self.output.root.server;
+    self.opacity_timer = c.wl_event_loop_add_timer(
+        c.wl_display_get_event_loop(server.wl_display),
+        handleOpacityTimer,
+        self,
+    ) orelse return;
```
ifreund, line 455: Should probably log an error here

river/View.zig
```diff
@@ -379,3 +402,73 @@ pub fn unmap(self: *Self) void {
...
+    self.armOpacityTimer();
+}
+
+/// Start an opacity transition
+pub fn startOpacityTransition(self: *Self, target_opacity: f32) void {
+    self.pending.target_opacity = target_opacity;
+}
```
ifreund, line 462: I'm not really a fan of one line functions like this. IMO they don't really pull their weight in terms of readability as I always end up going to the definition to see what it actually does.

river/XdgToplevel.zig
```diff
@@ -251,8 +251,12 @@ fn handleCommit(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
             view.pending_serial = null;
             if (view.shouldTrackConfigure())
                 view.output.root.notifyConfigured()
-            else
+            else {
+                var view_tags_changed = view.pending.tags != view.current.tags;
```
ifreund, line 255:
```suggestion
                const view_tags_changed = view.pending.tags != view.current.tags;
```

river/Config.zig
```diff
@@ -66,6 +66,21 @@ csd_filter: std.ArrayList([]const u8),
...
+view_opacity_focused: f32 = 1.0,
+
+/// The opacity of unfocused views
+view_opacity_unfocused: f32 = 1.0,
+
+/// The starting opacity of new views
+view_opacity_initial: f32 = 1.0,
+
+/// View opacity transition step
+view_opacity_delta: f32 = 0.1,
```
ifreund, line 79:
```suggestion
view_opacity_delta: f32 = 1.0,
```

contrib/config.sh
```diff
@@ -100,3 +100,5 @@ riverctl float-filter-add "popup"
 # Set app-ids of views which should use client side decorations
 riverctl csd-filter-add "gedit"
 
+# Set opacity and fade effect
+riverctl opacity 1.0 1.0 0.0 0.1 20
```
ifreund, line 104: Let's have an example with animation but commented out by default.
```suggestion
#riverctl opacity 1.0 0.75 0.0 0.1 20
```

PR #112, opened 2020-10-04
Use a wayland protocol for communicating layouts

Leon-Plickat:
TODO:
* [X] Protocol
* [x] River implementation
  * [x] "float-mode" (when no layout is in use, like when an output has no namespace yet or there is no layout with a matching namespace)
    * [X] Keep track of ~~if view dimensions on an output are set by a layout client~~ active (meaning _last used_ or _pending_) layout per output.
    * [x] Do not automatically set views to floating when moved/resize manually (with mouse or keyboard or command) when in "float-mode". This would not be expected by users and is confusing since the change is only visible once the views are layed out again. So instead let's just act like dwm, where all moving and resizing of non-floating windows in "float-mode" is forgotten once the next layout is active. This will also prevent all potential bugs and unexpected behaviour which could happen when a layout is demanded and one of the views is moved by the user before the layout is committed.
  * [x] man page
    * [X] Document default options of outputs
  * [X] Block output transaction when layout demand is in progress
* [X] Rivertile implementation
  * [X] river-layout
  * [x] river-options
  * [x] Gaps
    * [x] use river-options  
  * [x] man page
* [x] Simple reference client implementation in contrib
  * [X] Demonstrate river-layout
  * [x] Demonstrate using river-option
* [x] New keybinds for default config
* [x] transaction

---

~~With the `ack_layout_request` event, multiple layout clients can run side by side. River uses the one which send the ack first and ignores the others. Idea: maybe people want to implement different layouts as different clients.~~

The serial in every request / event is needed because a layout client may be tasked with generating layout for multiple outputs simultaneously.


Leon-Plickat:
We also need to think about what happens when no layout client is running.

My current idea is to get rid of the built-in "full" layout and instead place new windows at 0,0 (or maybe center), let them have their requested size, and don't touch existing windows. That probably simplifies things and has the interesting side-effect that river could be used as a floating compositor.

ifreund:
> We also need to think about what happens when no layout client is running.
> 
> My current idea is to get rid of the built-in "full" layout and instead place new windows at 0,0 (or maybe center), let them have their requested size, and don't touch existing windows. That probably simplifies things and has the interesting side-effect that river could be used as a floating compositor.

That sounds reasonable to me. I'm fine with river basically being a minimal floating compositor by default with no layout client running.

ifreund:
Commenting so we don't let it slide: we need to address the problem of switching between multiple layouts before committing to this protocol fully. A common use case would be using the default `rivertile` layout generator alongside a more specialized custom layout generator.

Leon-Plickat:
Actually there are **two** problems to solve here: How to switch between different layout generator clients and how to get arbitrary information from the user to a running client. The importance of the latter one can be demonstrated by an example: A user may want to switch the master position of rivertile (and running four instances of rivertile at the same time is a bit stupid IMO).

A simple solution to both problems is just letting the user kill the current layout generator and exec a new one (or the same one but with different args). That however is not a very clean solution and should be avoided if possible IMO.

I believe the first problem must be solved entirely in the compositor alone without any interaction with the client.

The first problem may be solved with namespaces: When creating a `river_layout` object, a client would need to register an arbitrary namespace string. When requesting a layout, river listens to the client which first `ack`s the request which also has the correct namespace. A user could then switch namespaces with `riverctl layout <namespace>`. This approach allows multiple layout clients to run simultaneously, which I think is important to allows different layout across different outputs. *(Note that I think the namespace should be different from the layout name, as the latter is pretty-printed data explicitly provided for the user and may change depending on different circumstances: A client may put things like the amount of views in its master area into its name string, or use somethin abstract, like the dwm-typical `[]=` .)*

For the second problem I see two approaches: **1)** We add additional parameters to the layout requests, like `direction` (which are obviously also only optional just like the current ones), or **2)** we add another arbitrary string, which is set by the user at runtime and send to the client upon a layout request. I personally prefer the first option.

ifreund:
> When creating a river_layout object, a client would need to register an arbitrary namespace string. When requesting a layout, river listens to the client which first acks the request which also has the correct namespace. A user could then switch namespaces with riverctl layout <namespace>

Why don't we just send the request to the first client to register in target namespace? then we don't waste the user's cpu time waking up all the layout clients that will be ignored anyways.

Leon-Plickat:
Yeah that sounds reasonable.

Leon-Plickat:
I have been over this again and have a few thoughts:

* Instead of the `advertise_*` events, wouldn't it be better to send the view states in a `wl_array` of a special struct (like `river_layout_view_state`) we define in the protocol? That would make layout client depending on the additional information considerably easier to write, as all logic would happen in response to a single event (the layout request).
* The additional information should not include the current dimensions and position of the view. Layouts could otherwise depend on the current layout, which would be circular. Also those values are basically arbitrary, so layout logic could become opaque to the user, which we should avoid.
* We now have namespaces and intend to only send the layout request to a single client. If we explicitly note this in the protocol, we can drop `ack_layout_request` as well as the "layout authority" concept, which would seriously simplify the protocol.

ifreund:
>     * Instead of the `advertise_*` events, wouldn't it be better to send the view states in a `wl_array` of a special struct (like `river_layout_view_state`) we define in the protocol? That would make layout client depending on the additional information considerably easier to write, as all logic would happen in response to a single event (the layout request).

Hard nack, this would make the protocol dependent on the C ABI and there is no way for the scanner to scan arbitrary struct definitions out of the protocol code (for good reason). What we could and probably should do is provide example/skeleton code in a few popular languages to get people started, maybe even make it into a small library if it makes sense.

>     * The additional information should not include the current dimensions and position of the view. Layouts could otherwise depend on the current layout, which would be circular. Also those values are basically arbitrary, so layout logic could become opaque to the user, which we should avoid.

Totally agree there.

>     * We now have namespaces and intend to only send the layout request to a single client. If we explicitly note this in the protocol, we can drop `ack_layout_request` as well as the "layout authority" concept, which would seriously simplify the protocol.

This seems like a reasonable approach to take. How should we handle the case where the client requests a namespace that is already in use then? I suppose we could make that a protocol error but that seems harsh as clients have no way of knowing what namespaces are in use.

Leon-Plickat:
> Hard nack, this would make the protocol dependent on the C ABI and there is no way for the scanner to scan arbitrary struct definitions out of the protocol code (for good reason). What we could and probably should do is provide example/skeleton code in a few popular languages to get people started, maybe even make it into a small library if it makes sense.

Yeah, I wasn't sure if it is possible, which is why I came up with the events in the first place.

I prefer the skeleton approach. A library would be a bit more involved to support.

> This seems like a reasonable approach to take. How should we handle the case where the client requests a namespace that is already in use then? I suppose we could make that a protocol error but that seems harsh as clients have no way of knowing what namespaces are in use.

A protocol error is definitely too harsh. Also client could get different results (protocol error vs works as expected) when doing the exact same thing, which isn't a good thing IMO.

Maybe it should be possible for a client to change its namespace whenever it wants (instead of just allowing it to be set at connection) and river would then send an "invalid namespace" (or a more generic "connection refused") event if the namespace is already used or unsuitable for other reasons. I don't think we can get around this communication step if we want to  handle this cleanly.

Leon-Plickat:
You are right, allowing clients to change their namespace at any time is probably racy. So maybe we don't allow that and when the namespace is invalid, the client can't do anything other than destroying the `river_layout` object. If it wants to try again, it has to bind a new one.

Leon-Plickat:
* Rebased and squashed.

* Removed old broken contrib layouts and added a new one.
  * Todo: More contrib layouts in different languages, like rust (zig probably not needed since rivertile exists).

* Started porting rivertile.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+    <request name="set_view_dimensions">
+      <description summary="propose view dimensions">
+        A client may use this to propose position and dimensions of one of the
+        set of views for which a layout has been requested.
+
+        It is not necessary to propose position and dimensions for every view.
+        If the client did not make a proposal of these for a view, they will be
+        determined by the compositor.
+
+        This request may be send before the view of the index has been
```
ifreund, line 149:
```suggestion
        This request may be sent before the view of the index has been
```

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+        advertised by the compositor.
+
+        Dimensions are in output local logical coordinates per XDG-output
+        protocol extensions.
+
+        The serial is used to identify the layout request this request is a
+        response to.
+      </description>
+      <arg name="serial" type="uint" summary="serial of layout request"/>
+      <arg name="index" type="uint" summary="index of view in layout"/>
```
ifreund, line 159: We don't strictly need the index argument, we could simply rely on the order. We could make it a protocol error to send too many or too few `set_view_dimensions` requests.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+        set of views for which a layout has been requested.
+
+        It is not necessary to propose position and dimensions for every view.
+        If the client did not make a proposal of these for a view, they will be
+        determined by the compositor.
+
+        This request may be send before the view of the index has been
+        advertised by the compositor.
+
+        Dimensions are in output local logical coordinates per XDG-output
```
ifreund, line 152:
```suggestion
        Dimensions are in output local logical coordinates per xdg-output
```

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+    </event>
+
+    <request name="set_view_dimensions">
+      <description summary="propose view dimensions">
+        A client may use this to propose position and dimensions of one of the
+        set of views for which a layout has been requested.
+
+        It is not necessary to propose position and dimensions for every view.
+        If the client did not make a proposal of these for a view, they will be
+        determined by the compositor.
```
ifreund, line 147: What's your reasoning for allowing this? It seems to me it would be better to start off more strict and loosen up if needed.

ifreund (commented):
This is looking very close to ready, there's still a bit of cleanup that can be done though.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+        have set a namespace.
+
+        Setting the namespace may fail, for example when the namespace is
+        already in use by a different client. In that case the invalid_namespace
+        event is raised and the client can only destroy the river_layout object.
+      </description>
+      <arg name="namespace" type="string" summary="namespace of the layout client"/>
+    </request>
+
+    <event name="invalid_namespace">
```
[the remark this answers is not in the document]
Leon-Plickat: I wanted to use this event also when a client tries to change its namespace when it already has one.

protocol/river-layout-unstable-v1.xml
Leon-Plickat: With the index, logic of more advanced layout clients is simplified because they can send view dimensions in any order and don't need to keep a list.

protocol/river-layout-unstable-v1.xml
Leon-Plickat: Sure.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+
+        The serial of this event is used to identify subsequent events and
+        request as belonging to this layout request. Beware that the client
+        might need to handle multiple layout requests at the same time.
+      </description>
+      <arg name="view_amount" type="uint" summary="amount of views which are part of the layout"/>
+      <arg name="master_amount" type="uint" summary="proposed views in the master area"/>
+      <arg name="master_factor" type="fixed" summary="proposed master area factor"/>
+      <arg name="x" type="int" summary="x coordinate of useable area on output"/>
+      <arg name="y" type="int" summary="y coordinate of useable area on output"/>
```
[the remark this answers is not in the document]
Leon-Plickat: Sure.

protocol/river-layout-unstable-v1.xml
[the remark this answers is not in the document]
ifreund: I see. That should be a fatal protocol error though but also won't be necessary with the layout-manager approach.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+      <description summary="destroy the river_layout object">
+        This request indicates that the client will not use the river_layout
+        object any more.
+      </description>
+    </request>
+
+    <request name="set_namespace">
+      <description summary="set the namespace of the layout client">
+        This request is used by the client to inform the server about its
+        namespace. Clients may only set their namespace once.
```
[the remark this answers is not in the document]
Leon-Plickat: On it.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,194 @@
...
+      <arg name="height" type="uint" summary="height of view"/>
+    </request>
+
+    <request name="commit">
+      <description summary="commit a layout request">
+        The client may use this to signal the compositor that it has finished
+        proposing a layout.
+
+        Sending this event irrevocably marks the layout as finished. The client
+        may not send any other responses to the layout request.
```
ifreund, line 172:
```suggestion
        may not send any other responses to the layout request, doing so is a protocol error.
```
and create/add an entry to the `error` enum.

protocol/river-layout-unstable-v1.xml
ifreund: That sounds fine to me. We need protocol errors for setting an out of bounds index and for leaving an index unset then.

Leon-Plickat:
I think we have to rename "layout request" to something else, because technically it is not a request but rather an event. The current name is a  bit confusing.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,205 @@
...
+  </description>
+
+  <interface name="zriver_layout_manager_v1" version="1">
+    <description summary="manage river layout objects">
+      A global factory for river_layout objects.
+    </description>
+
+    <enum name="error">
+      <entry name="invalid_serial" value="0" summary="serial does not belong to an active layout demand">
+      <entry name="invalid_index" value="0" summary="index does not belong to a view in the list">
```
ifreund, line 52:
```suggestion
      <entry name="invalid_index" value="1" summary="index does not belong to a view in the list">
```

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,205 @@
...
+    <request name="get_river_layout">
+      <description summary="create a river_layout object">
+        This creates a new river_layout object for the given wl_output.
+
+        All layout related communication is done through this interface.
+
+        The namespace is used by the compositor to decide which river_layout
+        object will receive layout demands for the output. Namespaces must be
+        globally unique. If the namespace is already in use by a different
+        river_layout object, this request will fail.
```
ifreund, line 72: Requests cannot fail in wayland aside from a protocol error which is fatal and terminates the client. We need to send an event on the `river_layout` object to handle this case:
```suggestion
        globally unique. If the namespace is already in use by a different
        river_layout object, the server will send a namspace_in_use event on the river_layout object immediately after creation and ignore all requests on the object aside from the destroy request.
```

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,216 @@
...
+      <arg name="serial" type="uint" summary="serial of layout demand"/>
+    </event>
+
+    <request name="set_view_dimensions">
+      <description summary="propose view dimensions">
+        A client may use this to propose position and dimensions for one of the
+        set of views for which a layout has been demanded.
+
+        A client must propose position and dimensions for the entire set of
+        views, otherwise the layout will be ignored.
```
ifreund, line 170: Last change I see for the protocol:
```suggestion
        A client must propose position and dimensions for the entire set of
        views, leaving out an index is a protocol error.
```

river/Layout.zig
```diff
@@ -0,0 +1,196 @@
...
+
+        // TODO we need a list which has a static size, so nodes can be added at arbitrary indices
+        view_dimensions: std.TailQueue(ViewDimensions) = .{},
+
+        const ViewDimensions = struct {
+            x: i32, y: i32, w: u32, h: u32, index: u32
+        };
+
+        pub fn init(self: *LayoutRequest, output: *Output) void {
+            self.serial = getSerial(); // TODO probably just get the time
```
ifreund, line 55: you can use `wl.Server.getSerial()`/`wl.Server.nextSerial()` here.

contrib/layout.c
```diff
@@ -0,0 +1,294 @@
...
+	output->output = wl_registry_bind(registry, name, &wl_output_interface, 3);
+	output->context = context;
+	output->layout = NULL;
+
+
+	/* If we already have the river_layout_manager, we can get a
+	 * river_layout for this output.
+	 */
+	if ( context->layout_manager != NULL )
+		configure_output(output);
```
ifreund, line 182: No reason to do this here, I'd just create all the objects at once after all globals have been advertised. Less code is better!

contrib/layout.c
```diff
@@ -0,0 +1,294 @@
...
+{
+	/* Oh no, the namespace we choose is already used by another client!
+	 * All we can do now is destroy the river_layout object. Because we are
+	 * lazy, we just abort and let our cleanup mechanism destroy it. A more
+	 * sophisticated client could instead destroy only the one single
+	 * affected river_layout object and recover from this mishap. Writing
+	 * such a client is left as an exercise for the reader.
+	 */
+	struct Context *context = (struct Context *)data;
+	fputs("Invalid namespace.\n", stderr);
```
ifreund, line 142:
```suggestion
	fputs("Namespace already in use.\n", stderr);
```

contrib/layout.c
Leon-Plickat: This is needed for when outputs are added later.

Leon-Plickat: When we just create the `river_layout` objects once after the registry roundtrip, outputs added while we are in the main loop won't get a `river_layout`. We really do need to do this twice.

ifreund: Ah of course, don't mind me.

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,218 @@
...
+      <description summary="destroy the river_layout object">
+        This request indicates that the client will not use the river_layout
+        object any more.
+      </description>
+    </request>
+
+    <event name="namespace_in_use">
+      <description summary="namespace already in use">
+        The requested namespace is already used by another river_layout object.
+        After raising this event, the server will ignore all request on the
```
ifreund, line 100:
```suggestion
        After raising this event, the server will ignore all requests on the
```

protocol/river-layout-unstable-v1.xml
```diff
@@ -0,0 +1,218 @@
...
+        widgets such as panels.
+
+        Note that the main area related values are simply proposals and can
+        safely be ignored by the client.
+
+        The serial of this event is used to identify subsequent events and
+        request as belonging to this layout demand. Beware that the client
+        might need to handle multiple layout demands at the same time.
+
+        A client does not need to respond to this event.
```
ifreund, line 121: I think this should be clarified as good clients will respond to every event unless preemption occurs.
```suggestion
        The server will ignore responses to all but the most recent layout command. Thus, clients should only respond to the most recent layout_demand received. If a newer layout_demand is received before the client has finished responding to an old demand, the client should abort work on the old demand as any further work would be wasted.
```

protocol/river-layout-unstable-v1.xml
ifreund: Hmm, I'm wavering on this. It makes the server code much more complex and doesn't simplify things greatly even for "advanced" clients. Such "advanced" clients likely already need to allocate an array to hold the information from advertise_view requests, I don't see needing to allocate a second to be a huge burden. They could even probably reuse the same array with a bit of cleverness.

river/Root.zig
```diff
@@ -260,6 +260,11 @@ pub fn startTransaction(self: *Self) void {
     // Iterate over all views of all outputs
     var output_it = self.outputs.first;
     while (output_it) |output_node| : (output_it = output_node.next) {
+        // If a layout demand is currently in progress for this output,
+        // skip transaction on this output to avoid imperfect frames.
+        // We take Waylands motto, "every frame is perfect", seriously.
+        if (output_node.data.layout_demand != null) continue;
```
ifreund, line 266: This should preempt the current layout demand and make a new one

river/LayoutManager.zig
```diff
@@ -0,0 +1,300 @@
...
+
+    /// Arbitrary timeout
+    const timeout: u32 = 1000;
+
+    const ViewDimensions = struct {
+        x: i32, y: i32, w: u32, h: u32, index: u32
+    };
+
+    pub fn init(self: *LayoutDemand) void {
+        const output = @fieldParentPtr(Output, "layout_demand", self);
```
ifreund, line 51: you can't use `@fieldParentPtr()` here as `Output` stores a pointer to a `LayoutDemand` not the struct itself

river/Output.zig
```diff
@@ -149,157 +153,50 @@ pub fn sendViewTags(self: Self) void {
...
-/// and completed.
+/// Try to find a layout for this output and send it a layout demand.
+///
+/// If no suitable layout has been found, all views will simply retain their
+/// current dimensions. So without any active layouts, river will function
+/// like a simple floating WM.
+///
+///  If a layout has been found, the changes of view dimensions are async
+///  and can not be expected to apply with the next transaction. Therefore
+///  the layout handler will start transaction once the layout is done.
```
ifreund, line 169: If there is a layout client with the proper namespace, we should not start the transaction until after the client has answered our layout demand.

ifreund: That is, the order of events should look something like this:

1. startTransaction() is called
2. layout_demand event is sent to the layout client
3. layout client responds with dimensions for all views
4. configure events are sent to all views needing them
5. after all views have acked the configure and committed, commitTransaction() is called.

Leon-Plickat: I am pretty sure that is what the code does, the comment is just not up-to-date yet.

river/Root.zig
Leon-Plickat: why?

Leon-Plickat:
I have now completely purged the word "index" from the protocol and made it clear that views can not be addressed out of order.

Leon-Plickat:
We should probably discuss whether we want to add the foreign toplevel handle to the view advertise event now or in a later update.

Also I think the new river control protocol should have some sort of `arrangeOutput` request which triggers a `output.arrangeViews()` on an output. That way, a layout client can also change the layout based on other factors (both from other Wayland protocol extensions as well as external). Maybe we use this request also to set the layout namespace, and just re-arrange if it is passed NULL.

ifreund:
> We should probably discuss whether we want to add the foreign toplevel handle to the view advertise event now or in a later update.

I'd say later update, lets keep things simple for now.

> Also I think the new river control protocol should have some sort of `arrangeOutput` request which triggers a `output.arrangeViews()` on an output. That way, a layout client can also change the layout based on other factors (both from other Wayland protocol extensions as well as external). Maybe we use this request also to set the layout namespace, and just re-arrange if it is passed NULL.

This sounds reasonable, let's wait for a use-case to arise though.

river/LayoutManager.zig
```diff
@@ -0,0 +1,356 @@
...
+
+            // We receive this event when the client wants to mark the proposed layout
+            // of the layout demand matching the serial as done.
+            .commit => |req| {
+                // Are we a valid layout?
+                if (self.namespace == null) {
+                    layout.postError(
+                        .no_namespace,
+                        "after receiving the namespace_in_use event, all requests other than destroy are illegal",
+                    );
```
ifreund, line 233: This can't be a protocol error as it would be racy. We should just state that all requests other than destroy are ignored.

river/Output.zig
```diff
@@ -56,19 +67,23 @@ usable_box: Box,
 views: ViewStack(View) = .{},
 
 /// The double-buffered state of the output.
-current: State = State{ .tags = 1 << 0 },
-pending: State = State{ .tags = 1 << 0 },
+current: State = State{ .tags = 1 << 0, .layout = false },
+pending: State = State{ .tags = 1 << 0, .layout = false },
```
ifreund, line 71: Using a default value here would be more idiomatic. (should probably have done that for `tags`)
```suggestion
pending: State = State{ .tags = 1 << 0, .layout = false },
```

river/XdgToplevel.zig
```diff
@@ -170,7 +170,7 @@ fn handleMap(listener: *wl.Listener(*wlr.XdgSurface), xdg_surface: *wlr.XdgSurfa
 
     view.surface = self.xdg_surface.surface;
 
-    // Use the view's "natural" size centered on the output as the default
+    // Use the views "natural" size centered on the output as the default
```
ifreund, line 173: The apostrophe was correct.
```suggestion
    // Use the view's "natural" size centered on the output as the default
```

river/XdgToplevel.zig
```diff
@@ -179,6 +179,10 @@ fn handleMap(listener: *wl.Listener(*wlr.XdgSurface), xdg_surface: *wlr.XdgSurfa
     view.float_box.y = std.math.max(0, @divTrunc(@intCast(i32, view.output.usable_box.height) -
         @intCast(i32, view.float_box.height), 2));
 
+    // Also use the views  "natural" size as the initial regular dimensions,
```
ifreund, line 182:
```suggestion
    // Also use the view's "natural" size as the initial regular dimensions,
```

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+const View = @import("View.zig");
+const ViewStack = @import("view_stack.zig").ViewStack;
+
+const Error = error{BindFailed};
+
+/// This object holds the server side data for an ongoing layout demand
+pub const LayoutDemand = struct {
+    serial: u32 = 0,
+    views: u32 = 0,
+    output: *Output = undefined,
```
ifreund, line 41: These are manually set below in `init()` so the default values are superfluous.
```suggestion
    serial: u32,
    views: u32,
    output: *Output,
```

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+
+    /// Timeout timer
+    timer: ?*wl.EventSource = null,
+
+    /// Proposed view dimensions
+    view_dimensions: std.TailQueue(Box) = .{},
+
+    /// Arbitrary timeout
+    const timeout: u32 = 1000;
```
ifreund, line 50: Pretty sure this can just be left a `comptime_int`
```suggestion
    const timeout = 1000;
```

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+pub const LayoutDemand = struct {
+    serial: u32 = 0,
+    views: u32 = 0,
+    output: *Output = undefined,
+
+    /// Timeout timer
+    timer: ?*wl.EventSource = null,
+
+    /// Proposed view dimensions
+    view_dimensions: std.TailQueue(Box) = .{},
```
ifreund, line 47: I don't see any reason this needs to be a linked list. Why not just an `std.ArrayList(Box)`?

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+    pub fn pushViewDimensions(self: *LayoutDemand, x: i32, y: i32, width: u32, height: u32) !void {
+        const node = try util.gpa.create(std.TailQueue(Box).Node);
+
+        // Here we apply the offset to align the coords with the origin of the
+        // usable area and shrink the dimensions to accomodate the border size.
+        node.data = .{
+            .x = x + self.output.usable_box.x + @intCast(i32, self.output.root.server.config.border_width),
+            .y = y + self.output.usable_box.y + @intCast(i32, self.output.root.server.config.border_width),
+            .width = width - 2 * self.output.root.server.config.border_width,
+            .height = height - 2 * self.output.root.server.config.border_width,
```
ifreund, line 114: This will underflow if width is small and border width is high. We need to handle this case.

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+
+        return null;
+    }
+};
+
+/// This object encapsulates a zriver.LayoutV1 object bound by a client
+pub const Layout = struct {
+    layout: *zriver.LayoutV1 = undefined,
+    namespace: ?[]const u8 = null,
+    output: *Output = undefined,
```
ifreund, line 136: Default values don't really make sense here either as all of these fields must be set to something else in `init()`
```suggestion
    layout: *zriver.LayoutV1,
    namespace: ?[]const u8,
    output: *Output,
```

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+        // and free the namespace buffer, marking the Layout object as invalid.
+        var it = output.layouts.first;
+        while (it) |layout_node| : (it = layout_node.next) {
+            if (std.mem.eql(u8, self.namespace.?, layout_node.data.namespace.?)) {
+                self.layout.sendNamespaceInUse();
+                util.gpa.free(self.namespace.?);
+                self.namespace = null;
+                return;
+            }
+        }
```
ifreund, line 159: Not sure if you saw my comment on IRC, but I think enforcing globally unique namespaces where the same client may register the same namespace on multiple outputs would be the most intuitive way to handle this. This would mean iterating over all layout objects on all outputs and only allowing a duplicate if `layout.getClient()` returns the same `wl.Client` pointer.

river/Output.zig
```diff
@@ -56,19 +67,23 @@ usable_box: Box,
...
 
-/// Current layout of the output. If it is "full", river will use the full
-/// layout. Otherwise river assumes it contains a string which, when executed
-/// with sh, will result in a layout.
-layout: []const u8,
+/// Current layout-namespace of the output.
+layout_namespace: ?[]const u8 = null,
+
+/// The currently active LayoutDemand
+layout_demand: ?*LayoutManager.LayoutDemand = null,
```
ifreund, line 83: Is there a reason this needs to be a pointer? It seems to me that we could get rid of the indirection and allocation here:
```suggestion
layout_demand: ?LayoutManager.LayoutDemand = null,
```

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+            self.output.usable_box.height,
+            self.output.layout_demand.?.serial,
+        );
+
+        // And finally we advertise all visible views
+        var it = ViewStack(View).iter(self.output.views.first, .forward, self.output.pending.tags, Output.arrangeFilter);
+        while (it.next()) |view| {
+            self.layout.sendAdvertiseView(
+                view.pending.tags,
+                view.getAppId() orelse "[no-appid]",
```
ifreund, line 194: Hmm, should we just add `allow-null=true` in the protocol?

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+                self.output.layout_demand.?.serial,
+            );
+        }
+        self.layout.sendAdvertiseDone(self.output.layout_demand.?.serial);
+    }
+
+    fn handleRequest(layout: *zriver.LayoutV1, request: zriver.LayoutV1.Request, self: *Layout) void {
+        switch (request) {
+            .destroy => {
```
ifreund, line 203: This should do nothing but call `layout.destroy()`. The actual destruction should then happen in the destroy function passed to `setHandler()`. See this section of the wlroots contributing guide for other good rules to follow when implementing protocols server-side: https://github.com/swaywm/wlroots/blob/master/CONTRIBUTING.md#wayland-protocol-implementation

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+    fn handleRequest(layout_manager: *zriver.LayoutManagerV1, request: zriver.LayoutManagerV1.Request, self: *LayoutManager) void {
+        switch (request) {
+            .destroy => {
+                self.deinit();
+
+                // Remove this client from the global clients list
+                const node = @fieldParentPtr(std.TailQueue(LayoutManager).Node, "data", self);
+                self.parent.layout_managers.remove(node);
+                util.gpa.destroy(node);
+            },
```
ifreund, line 320: Same as above, this should just call `layout_manager.destroy()` and free the memory in the destroy handler set with `setHandler()`

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+                    return;
+                };
+                output.layouts.append(node);
+            },
+        }
+    }
+};
+
+global: *wl.Global,
+server_destroy: wl.Listener(*wl.Server) = undefined,
```
ifreund, line 347:
```suggestion
server_destroy: wl.Listener(*wl.Server) = wl.Listener(*wl.Server).init(handleServerDestroy),
```
(and remove the `setNotify()` call below)

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+
+fn handleServerDestroy(listener: *wl.Listener(*wl.Server), wl_server: *wl.Server) void {
+    const self = @fieldParentPtr(Self, "server_destroy", listener);
+    self.global.destroy();
+
+    // Cleanup all connected clients
+    while (self.layout_managers.pop()) |node| {
+        node.data.deinit();
+        util.gpa.destroy(node);
+    }
```
ifreund, line 367: Don't need this once you create a destroy handler for the resource as libwayland will destroy all resources automatically when the display is terminated.

river/LayoutManager.zig
```diff
@@ -0,0 +1,378 @@
...
+                // TODO layout name?
+            },
+        }
+    }
+};
+
+/// This object encapsulates a zriver.LayoutManagerV1 object bound by a client
+const LayoutManager = struct {
+    parent: *Self,
+    layout_manager: *zriver.LayoutManagerV1,
```
ifreund, line 294: I'm not sure splitting this off into another struct is really worth it. The implementations of river-control and river-status simply do everything in the toplevel object. I don't think you even need to store a list of `*zriver.LayoutManagerV1` pointers.

ifreund (commented):
Here's some in depth code review for the server side implementation. I haven't had a chance to take a good look at rivertile yet though I will get to that soon.

I'd also recommend moving the `Layout` and `LayoutDemand` structs to their own files.

river/Layout.zig
```diff
@@ -0,0 +1,182 @@
...
+const ViewStack = @import("view_stack.zig").ViewStack;
+const LayoutDemand = @import("LayoutDemand.zig");
+
+layout: *zriver.LayoutV1,
+namespace: ?[]const u8,
+output: *Output,
+
+pub fn init(self: *Self, namespace: [*:0]const u8, output: *Output, client: *wl.Client, version: u32, id: u32) !void {
+    self.output = output;
+    self.namespace = try util.gpa.dupe(u8, std.mem.span(namespace));
```
ifreund, line 42: this gets leaked if the next line returns an error. You could use errdefer here or postpone allocation till later and use `wl.Client.postNoMemory()` on failure

river/Layout.zig
```diff
@@ -0,0 +1,182 @@
...
+    log.debug(
+        .layout_manager,
+        "starting layout demand '{}' on output '{}'",
+        .{ self.namespace, self.output.wlr_output.name },
+    );
+
+    // First we create the Layoutdemand object
+    var demand: LayoutDemand = .{};
+    demand.init(self.output, views);
+    self.output.layout_demand = demand;
```
ifreund, line 88: We don't store pointers to the layout demand anywhere, so we can use the more idiomatic
```suggestion
    self.output.layout_demand = LayoutDemand.init(self.output, views);
```

river/LayoutDemand.zig
```diff
@@ -0,0 +1,150 @@
...
+views: u32 = 0,
+output: *Output = undefined,
+
+/// Timeout timer
+timer: ?*wl.EventSource = null,
+
+/// Proposed view dimensions
+view_dimensions: std.ArrayList(Box) = undefined,
+
+pub fn init(self: *Self, output: *Output, views: u32) void {
```
ifreund, line 51: We don't need to store pointers to the layout demand anywhere, so we can use a more idiomatic signature for init:
```suggestion
pub fn init(output: *Output, views: u32) Self {
```

river/LayoutDemand.zig
```diff
@@ -0,0 +1,150 @@
...
+const ViewStack = @import("view_stack.zig").ViewStack;
+
+const Error = error{ViewDimensionMismatch};
+
+/// Arbitrary timeout
+const timeout = 1000;
+
+serial: u32 = 0,
+views: u32 = 0,
+output: *Output = undefined,
```
ifreund, line 43:
```suggestion
serial: u32,
views: u32,
output: *Output,
```

river/LayoutDemand.zig
```diff
@@ -0,0 +1,150 @@
...
+
+/// Timeout timer
+timer: ?*wl.EventSource = null,
+
+/// Proposed view dimensions
+view_dimensions: std.ArrayList(Box) = undefined,
```
ifreund, line 49:
```suggestion
view_dimensions: std.ArrayList(Box),
```

river/LayoutDemand.zig
```diff
@@ -0,0 +1,150 @@
...
+        else
+            height,
+    };
+}
+
+/// Apply the proposed layout to the output
+pub fn apply(self: *Self) !void {
+    const output = self.output;
+
+    // Whether the layout demand succeeds or fials, we need to finish the layout demand
```
ifreund, line 122:
```suggestion
    // Whether the layout demand succeeds or fails, we need to finish the layout demand
```

rivertile/main.zig
```diff
@@ -14,129 +14,322 @@
...
+        std.os.exit(1);
+    }
+
+    context.configureAllOutputs();
+    defer context.destroyAllOutputs();
+
+    while (context.running) {
+        _ = display.dispatch() catch {
+            context.running = false;
+        };
```
ifreund, line 317: Let's not exit cleanly if there's an error.
```suggestion
        _ = try display.dispatch();
```

rivertile/main.zig
```diff
@@ -14,129 +14,322 @@
...
+fn registryListener(registry: *wl.Registry, event: wl.Registry.Event, context: *Context) void {
+    switch (event) {
+        .global => |global| {
+            if (std.cstr.cmp(global.interface, zriver.LayoutManagerV1.getInterface().name) == 0) {
+                context.layout_manager = registry.bind(global.name, zriver.LayoutManagerV1, 1) catch return;
+            } else if (std.cstr.cmp(global.interface, wl.Output.getInterface().name) == 0) {
+                context.addOutput(registry, global.name) catch {
+                    std.debug.warn("Failed to bind output.\n", .{});
+                    context.running = false;
+                };
```
ifreund, line 330: This doesn't handle the case where an output is advertised before the river layout manager properly. Currently `addOutput()` will return an error and we will exit.

rivertile/main.zig
Leon-Plickat: That makes rivertile always error out when river closes, which I think is not good behaviour, considering this will be the one main way rivertile will be stopped.

rivertile/main.zig
Leon-Plickat: I am pretty sure it does. If the configuration fails due to the river layout manager not being available, it will not error out. I have edited the code to make this more clear.

Leon-Plickat:
Things left to do / consider:

* What to do with the layout name? Since we currently don't have a way to inform status tracking clients of it, I left it out for now. Should be added later though.
* River crashes on exit. Might not be related to this PR, haven't tested master yet.
* Transactions: If I block`root.commitTransaction()` for outputs with an active layout demand, river crashes when a view is destroyed. If I don't block, then we break "every frame is perfect", as for example newly spawned views will be visible with their initial dimensions for a short amount of time and when destroying views a "hole" in the layout will be visible for a short amount of time.

Leon-Plickat:
And if we want any other layout-related values send from the compositor to the client, such as perhaps padding size, now is the best time to add them.

Maybe we could also do what dwms `sfact` patch does: Give every single view a configurable value which is passed to the layout, which could be used to change the relative sizes of individual views. I don't have a use case for this, but we should consider it now rather than later.

ifreund:
>     * What to do with the layout name? Since we currently don't have a way to inform status tracking clients of it, I left it out for now. Should be added later though.

Leaving it out is fine for now. I don't anticipate any problems working it into a revised status protocol.

>     * River crashes on exit. Might not be related to this PR, haven't tested master yet.

It doesn't crash (for me at least) on master. 

>     * Transactions: If I block`root.commitTransaction()` for outputs with an active layout demand, river crashes when a view is destroyed. If I don't block, then we break "every frame is perfect", as for example newly spawned views will be visible with their initial dimensions for a short amount of time and when destroying views a "hole" in the layout will be visible for a short amount of time.

There should be a way to make it not crash, I'll take a look.

ifreund:
> And if we want any other layout-related values send from the compositor to the client, such as perhaps padding size, now is the best time to add them.

`river-options` was just pushed to master, so there's no need for anything more in this protocol as options can be used to pass arbitrary values to the client. We should however add some options defined by river by default such as `view_padding`, `outer_padding`, `main_count`, and `main_factor` which all layouts may expect to be defined.

> Maybe we could also do what dwms `sfact` patch does: Give every single view a configurable value which is passed to the layout, which could be used to change the relative sizes of individual views. I don't have a use case for this, but we should consider it now rather than later.

This doesn't seem very useful to me while also sounding quite complex. I don't think I want this feature.

example/init
```diff
@@ -103,13 +93,10 @@ riverctl map normal $mod Space toggle-float
...
-riverctl map normal $mod Up layout rivertile top
-riverctl map normal $mod Right layout rivertile right
-riverctl map normal $mod Down layout rivertile down
-riverctl map normal $mod Left layout rivertile left
-
-# Mod+S to change to Full layout
-riverctl map normal $mod S layout full
+riverctl map normal $mod Up layout  tile-top
+riverctl map normal $mod Right layout tile-right
+riverctl map normal $mod Down layout tile-down
```
ifreund, line 98:
```suggestion
riverctl map normal $mod Down layout tile-bottom
```

Leon-Plickat:
Mini-brain-dump: Eventually we should check at the time of applying a layout whether some views are hidden behind others. Those views should not be rendered and their damage ignored.

Leon-Plickat:
@ifreund  You force pushed an outdated version of this branch, overwriting a ton of code related to the river-options integration, basically most of the things I did this year. Since I just pulled your branch, I do not have that code anymore.

novakne:
I think it should be more clear that you need to start a layout client with like `riverctl spawn rivertile` before you can use it. I know you can see it in the `init` exemple but it would be better with something about this in one of the man pages imo.

Maybe adding an exemple in `rivertile` man page like this:
``` markdown
# EXEMPLES
Start rivertile:
`riverctl spawn rivertile`

Then you can use river-options to choose the desired layout
e.g:
`riverctl set-option -focused-output layout tile-left`
```

Or maybe in another man page  with a more general exemple

snakedye:
Could the protocol also provide the index of the focused window? I think it's the only thing missing to do proper "manual" window management.

Leon-Plickat:
That information is already provided indirectly through the order of the advertise view events.

snakedye:
How so?

Leon-Plickat:
Advertise view events are in the exact order of the view list.

snakedye:
> It is guaranteed that every view in the layout will be exactly once, in the exact order of the view list.

I do get this but even if I have the list how could I determine the index of the focused view? If I have 10 instances of foot, how do i determine which one is new or which one was previously focused if they all have the same app_id?

Leon-Plickat:
You don't. The index would not help here either.

snakedye:
May I know why?

Leon-Plickat:
Because when we allow layouts clients to know what view is currently focused, they might base they might base their layout on that. And if they do, the user expects the layout to update when they change the focus. And that gets messy.

snakedye:
They're already are window managers as in they determine the position of windows.

It is also already doable to create a view_index option and to increment and decrement it each time you `focus next` or `focus previous` but there could be a clean way to do it.

> And if they do, the user expects the layout to update when they change the focus. And that gets messy.

Since since a layout generator can have it's own layout logic and options, the user should also expect each one of them to behave differently.

Leon-Plickat:
To be frank I am not particularly interested in splitting hairs over the word "window manager". Layout generators are not window managers in the traditional sense (meaning like on X). The idea has always been for them to provide a static layout: They say river were to place windows, but are not involved in anything beyond that.

Leon-Plickat:
Yes, it theoretically is possible to do what you want. But I am of the opinion that the increased complexity is not worth it.

snakedye:
The idea was not for the layout to change when I change the focus of the view, i don't think sending a layout event each time it happens is particularly useful. It was the provide the index of the view when river would usually ask of a layout.

I don't really care when the focus is when nothing is happening, it's when the view_amount increases or decreases is so the layout generator knows where to insert or remove views. The layouts are generated is still "static" but the client is more flexible.

crypt17:
on a remi related matter is there a way to adjust the window order on creation

At the moment using awesome I have any newly  created window set to the first slave window using an awesome signal. 

This would be a nice feature to have if there was some way to subscribe to events such as window creation and then adjust the window stack location before it is presented  on screen.

If this is already possible I would be interested in any pointers on where to look

Leon-Plickat:
@crypt17 That sounds more like something we'd implement directly in river. See the `attach-mode` configuration option. Currently it only has `top` or `bottom` as possible states, but adding `after-main` would not be too hard.

Leon-Plickat:
@snakedye the layout protocol is designed with the intention that if a layout is based on a set of parameters, it gets updated when any of these change. Just providing the index of the focused view breaks this.

But this is a good reminder that I need to remove the view-tags from the advertise view events.

snakedye:
@Leon-Plickat It is beyond this PR but if the layout protocol doesn't provide the view index, I think the status protocol could and it would be fitting there. The layout protocol could provide the full title of views (with the app-id) so I could match it with the one I get from from the status protocol. 

This could also be viable although I have some "layout conflict" with the status protocol. For example, when I switch tags, the layout of the previously focused tag is generated before the one of the focused. My event loop might be at fault but I suspect it might be because by the time I filter all events the status protocol didn't send anything, so the previous tag layout is generated. After that I get send a status event and the appropriate layout is generated. This causes a small visual glitch because windows are resized in a split second.

Leon-Plickat:
ToDo: rethink order of layout demand event and output status event to make layouts depending on the currently active tags more seemless.

Ordering could also be left as is now, if the status event is send before the transaction is complete. A layout generator can wait to generate and commit a layout for the status event to be received.

snakedye:
Thanks @Leon-Plickat ! I tried building the branch to try it out but it fails. Wlroots 0.13.0 is not on Arch's community repositories yet so would it explain the errors?

novakne:
@snakedye yes you need wlroots 0.13

snakedye:
@Leon-Plickat I just installed wlroot 0.13.0 and can attest the patch works :+1: .

ifreund:
@Leon-Plickat I don't think 6f632f7 makes any sense. The view may be moved to a different tag while fullscreen or other views may be added to the view's tag, both of which require the view to be rearranged. What was the motivation for this change?

protocol/river-layout-v1.xml
```diff
@@ -0,0 +1,220 @@
...
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the river_layout_manager object">
+        This request indicates that the client will not use the
+        river_layout_manager object any more. Objects that have been created
+        through this instance are not affected.
+      </description>
+    </request>
+
+    <request name="get_river_layout">
```
ifreund, line 56: I think the "river" is redundant here and that we should drop it.
```suggestion
    <request name="get_layout">
```

Leon-Plickat:
@ifreund wouldn't those conditions trigger an arrange regardless? It will affect the fullscreened view because its new dimensions are saved in `pre_fullscreen_box` and will be used as soon as it leaves fullscreen.

The motivation, in addition to fixing #243, is avoiding unnecessary arranges. Without the patch, river always arranges views when leaving fullscreen, which does not make sense since fullscreen should not be affecting the layout in any way.

ifreund:
Ah I see, I missed that `pre_fullscreen_box` was updated in every arrange while the view was in fullscreen. Given that behavior, I find the name a little misleading. 

As for whether this is a good idea or not, I'm not entirely convinced. Generally, the less state we store the better as out-of-sync/stale state is a large source of bugs. I don't think a single extra arrange when transitioning a view out of fullscreen (a relatively rare event) is a significant issue. As for #243, we could restore views to their `float_box` on exiting fullscreen if there is no active layout.

protocol/river-layout-v1.xml
```diff
@@ -0,0 +1,220 @@
...
+    </request>
+
+    <event name="namespace_in_use">
+      <description summary="namespace already in use">
+        The requested namespace is already used by another river_layout object.
+        After receiving this event, the client should destroy the river_layout
+        object. Any other request will be ignored.
+      </description>
+    </event>
+
```
YerinAlexey, line 103: Why this is a separate event instead of an error?

river/LayoutDemand.zig
```diff
@@ -0,0 +1,138 @@
...
+/// Number of views for which dimensions have not been pushed.
+/// This will go negative if the client pushes too many dimensions.
+views: i32,
+/// Timeout timer
+timer: *wl.EventSource,
+/// Proposed view dimensions
+view_boxen: []Box,
+
+pub fn init(layout: *Layout, views: u32) !Self {
+    // Attach a timout timer
```
YerinAlexey, line 52:
```suggestion
    // Attach a timeout timer
```

river/LayoutDemand.zig
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/// Destroy the LayoutDemand on timeout.
+/// All further responses to the event will simply be ignored.
+fn handleTimeout(layout: *Layout) callconv(.C) c_int {
+    log.notice(
+        "layout demand for layout '{}' on output '{}' timed out",
+        .{ layout.namespace, layout.output.wlr_output.name },
+    );
+    layout.output.layout_demand.?.deinit();
```
YerinAlexey, line 78: Are you sure that `layout_demand` is non-null at this point?

river/LayoutDemand.zig
```diff
@@ -0,0 +1,138 @@
...
+        );
+        return Error.ViewDimensionMismatch;
+    }
+
+    // Apply proposed layout to views
+    var it = ViewStack(View).iter(output.views.first, .forward, output.pending.tags, Output.arrangeFilter);
+    var i: u32 = 0;
+    while (it.next()) |view| : (i += 1) {
+        if (view.pending.fullscreen) {
+            view.pre_fullscreen_box = self.view_boxen[i];
```
YerinAlexey, line 130: Isn't that going to include border size in fullscreen views?

protocol/river-layout-v1.xml
ifreund: errors are fatal, and the client can't possible know that the namespace was already in use

river/LayoutDemand.zig
ifreund: yep

river/main.zig
```diff
@@ -126,39 +127,36 @@ pub fn main() anyerror!void {
...
+        log_server.info("running startup command '{}'", .{cmd});
         const child_args = [_:null]?[*:0]const u8{ "/bin/sh", "-c", cmd, null };
-        const pid = try std.os.fork();
+        const pid = try os.fork();
         if (pid == 0) {
-            if (std.os.system.sigprocmask(std.os.SIG_SETMASK, &std.os.empty_sigset, null) < 0) unreachable;
-            std.os.execveZ("/bin/sh", &child_args, std.c.environ) catch c._exit(1);
+            if (c.setsid() < 0) unreachable;
+            if (os.system.sigprocmask(os.SIG_SETMASK, &os.empty_sigset, null) < 0) unreachable;
+            os.execveZ("/bin/sh", &child_args, std.c.environ) catch c._exit(1);
```
YerinAlexey, line 150: Since, this file is also touched. Maybe it makes sense to look up `sh` in user's `$PATH` as was discussed on IRC?

ifreund: I'd rather do that change in a separate PR, this one is large enough. Also I have yet to learn of a system that both can build river and does not have `/bin/sh`.

YerinAlexey (changes requested):
Looked at this patch again

river/main.zig
YerinAlexey: NixOS

YerinAlexey: > I'd rather do that change in a separate PR, this one is large enough.

Makes sense, but maybe then this patch don't need to touch this section? It doesn't seem to be related to layouts in any way.

> Also I have yet to learn of a system that both can build river and does not have `/bin/sh`.

Pretty sure that's NixOS

protocol/river-layout-v1.xml
YerinAlexey: But errors aren't so fatal that they cause clients to drop, are they?

ifreund:
```
12:53 <yyp> https://github.com/ifreund/river/pull/112#discussion_r615375314 - do I not understand Wayland errors? I thought they are just a different response to a request indicating that something went wrong, do they do anything more?
12:57 <ifreund> yes, they disconnect the client
13:02 <yyp> Ok, makes sense. But do we really need to keep the layout client connected when it's namespace is taken?
13:02 <yyp> s/it's/its
13:04 <ifreund> we need to allow clients to try again with a different namespace or report an error as they see fit
13:05 <ifreund> wayland errors should only be used for protocol violations
13:05 <ifreund> and protocol violations should never happen in a bug free client/compositor
```

river/main.zig
ifreund: > Makes sense, but maybe then this patch don't need to touch this section? It doesn't seem to be related to layouts in any way.

This is related, because I needed these changes to make using a service manager (runit) as my startup process ergonomic. This service manager is responsible for running `rivertile` among other things as `rivertile` now needs to run continuously as a separte process.

> Pretty sure that's NixOS

Pretty sure NixOS has `/bin/sh`, otherwise sway's `exec` commands wouldn't work.

Leon-Plickat:
> Ah I see, I missed that `pre_fullscreen_box` was updated in every arrange while the view was in fullscreen. Given that behavior, I find the name a little misleading.

true

> As for whether this is a good idea or not, I'm not entirely convinced. Generally, the less state we store the better as out-of-sync/stale state is a large source of bugs. I don't think a single extra arrange when transitioning a view out of fullscreen (a relatively rare event) is a significant issue. As for #243, we could restore views to their `float_box` on exiting fullscreen if there is no active layout.

That would mean that a views position when initially set to floating while a layout is active depends on whether it has been fullscreened earlier while no layout was active. Floating and fullscreen are from a UX perspective totally unrelated, so they should not form any causality chain in my opinion.

ifreund:
> That would mean that a views position when initially set to floating while a layout is active depends on whether it has been fullscreened earlier while no layout was active. Floating and fullscreen are from a UX perspective totally unrelated, so they should not form any causality chain in my opinion. Layout state being involved makes this even more awkward.

I wasn't suggesting that we should set the `float_box` when making a view fullscreen, only that we could use whatever the `float_box` happens to be as a reasonable size/position to give a view on making it not fullscreen if there is no layout.

river/main.zig
Leon-Plickat: @yyp Also note that this PR contains in fact multiple commits ("patches"). While the history is not final yet (some commits probably should be squashed), this PR will definitely have multiple commits. This PR containing other changes that are required by or depend on the layout implementation but are not directly related is fine. As long as commits themselves are strictly on-topic, the git history will be clean regardless.

Also getting `sh` from `$PATH` wasn't *really* discussed on IRC AFAIK. It was just mind dumb when I was wondering about #263. I have since come to the conclusion that this would not fix the issue, so I no longer consider doing that a good idea. Despite `/bin/sh` apparently not being a defined standard, it exists on effectively any machine. Now you could argue that `/bin/sh` isn't guaranteed to be a POSIX compatible shell and I'd agree, but then it should explicitly configurable, not implicitly snatched from the users `$PATH`. And that really belongs into another PR, to keep the discussions on-topic.

YerinAlexey: > > Makes sense, but maybe then this patch don't need to touch this section? It doesn't seem to be related to layouts in any way.
> 
> This is related, because I needed these changes to make using a service manager (runit) as my startup process ergonomic. This service manager is responsible for running `rivertile` among other things as `rivertile` now needs to run continuously as a separte process.

Fair enough

> > Pretty sure that's NixOS
> 
> Pretty sure NixOS has `/bin/sh`, otherwise sway's `exec` commands wouldn't work.

Maybe. But I still think that hard-coding `/bin/sh` is a bit of a weird move.

Leon-Plickat:
Oh, yeah that's better. Though it would mean that after leaving fullscreen all views would snap the their `float_box`, no matter what their previous box actually was, if no layout is active.

protocol/river-layout-v1.xml
Leon-Plickat: > But do we really need to keep the layout client connected when it's namespace is taken?

Yes. A client could very well have multiple layouts are some other functionality important to the user. We should not interrupt that just because some other client was faster at registering  a layout with a specific name.

ifreund:
> Oh, yeah that's better. Though it would mean that after leaving fullscreen all views would snap the their `float_box`, no matter what their previous box actually was, if no layout is active.

Hmm, we can leave views that weren't fullscreen at their previous dimensions, but then it might be strange that only the fullscreened view ends up in its `float_box` dimensions.

Maybe the approach you implemented is the best way to go, though we should `s/pre_fullscreen_box/post_fullscreen_box/`.

Leon-Plickat:
Changed the name.

Leon-Plickat:
Anyway, I would like to "feature freeze" this PR now and only include your upcoming options change and strictly layout related bug fixes.

What do you think about the git history? I think some commits maybe should be squashed.

river/View.zig
```diff
@@ -117,6 +117,10 @@ saved_buffers: std.ArrayList(SavedBuffer),
 /// view returns to floating mode.
 float_box: Box = undefined,
 
+/// Pre-fullscreen dimensions of the view, saved so that they can be restored if
+/// the view returns from fullscreen without needing to rearrange the output.
+post_fullscreen_box: Box = undefined,
```
ifreund, line 122:
```suggestion
/// While a view is in fullscreen, it is still arranged if a layout is active but the resulting dimensions are stored here instead of being applied to the view's state. This allows us to avoid an arrange when the view returns from fullscreen and for more intuitive behavior if there is no active layout for the output.
post_fullscreen_box: Box = undefined,
```

ifreund:
> Anyway, I would like to "feature freeze" this PR now and only include your upcoming options change and strictly layout related bug fixes.
> 
> What do you think about the git history? I think some commits maybe should be squashed.

I'll do a pass over the commits just before landing this in master. I've got a few things to debug with the options changes before I push them. Would you mind updating the example C layout for the new options protocol? If you want to get started before I push you can find the xml here: https://0x0.st/-A9l.xml

Leon-Plickat:
Great, I'll try to find some time tomorrow for the contrib layout.

ifreund:
Alright, the initial implementation of `river-options-v2` is done and seems to work well. One minor pain point is that users wanting to configure the options of e.g. `rivertile` before starting river tile must declare said options themselves. I'm not quite sure what, if anything, we should do about this though, so I don't think it needs to block the merge.

Barring any bugs I'm unaware of, the current merge blockers are updating the contrib layout in C for river-options-v2 and reviewing the documentation/example init script.

YerinAlexey (approved):

snakedye:
@ifreund Why should I destroy an undeclared option?

image

To give context, I use this option to send commands to my client. It is cleared (I set it's value to null) after the command is interpreted so I'm notified of following one.

ifreund:
> @ifreund Why should I destroy an undeclared option?
> 
> image
> 
> To give context, I use this option to send commands to my client. It is cleared (I set it's value to null) after the command is interpreted so I'm notified of following one.

The protocol has changed, options must be declared explicitly with one of the `declare_foo_option` requests and may no longer be implicitly declared through a handle to an undeclared option. Please give the new protocol a read and let me know on IRC if you have any more questions.

ifreund (approved):
Very nice work on this @Leon-Plickat!

PR #115, opened 2020-10-07
Implement moving and resizing views via keyboard

Leon-Plickat:
This PR introduces ~~two~~ *three* simple commands to move and resize views via the keyboard.

* `move {up,down,left,right} <delta>` to move a view by `delta` in the specified direction.

* `snap {up,down,left,right}` to snap a view to the specified screen edge.

* `resize {horizontal,vertical} <delta>` to resize a view by `delta` in the specified orientation.

Since all three actions remove the view from the layout, the view will be set to floating.

---

In the default config, moving is `Super+Alt+{h,j,k,l}` and snapping is `Super+Alt+Control+{h,j,k,l}`. Resizing ~~which I'll implement sometime in the future will be~~ *is* `Super+Alt+Shift+{h,j,k,l}`.

Went ahead and implemented resizing as well. Works like it does in sway: The views center remains stationary.

All three operations will not allow the view to escape the output, exactly like when moving with the moose. Resizing respects view constraint.

### DIFF
--- a/river/command/move.zig
+++ b/river/command/move.zig
@@ -62,10 +62,9 @@ pub fn snap(
     const direction = std.meta.stringToEnum(PhysicalDirection, args[1]) orelse
         return Error.InvalidPhysicalDirection;
 
-    const view = get_view(seat) orelse return;
-    const output_box = get_output_dimensions(view);
     const view = getView(seat) orelse return;
     const border_width = @intCast(i32, view.output.root.server.config.border_width);
+    const output_box = view.output.getEffectiveResolution();
     switch (direction) {
         .up => view.pending.box.y = border_width,
         .down => view.pending.box.y =
@@ -91,10 +90,9 @@ pub fn resize(
     const orientation = std.meta.stringToEnum(Orientation, args[1]) orelse
         return Error.InvalidOrientation;
 
-    const view = get_view(seat) orelse return;
-    const output_box = get_output_dimensions(view);
     const view = getView(seat) orelse return;
     const border_width = @intCast(i32, view.output.root.server.config.border_width);
+    const output_box = view.output.getEffectiveResolution();
     switch (orientation) {
         .horizontal => {
             var real_delta: i32 = @intCast(i32, view.pending.box.width);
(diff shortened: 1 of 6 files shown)


river/command/move.zig
```diff
@@ -0,0 +1,196 @@
...
+    if (args.len > 3) return Error.TooManyArguments;
+
+    const delta = try std.fmt.parseInt(i32, args[2], 10);
+    const orientation = std.meta.stringToEnum(Orientation, args[1]) orelse
+        return Error.InvalidOrientation;
+
+    const view = get_view(seat) orelse return;
+    const output_box = get_output_dimensions(view);
+    const border_width = @intCast(i32, view.output.root.server.config.border_width);
+    switch (orientation) {
```
ifreund, line 114: It'd be nice if we could unify these two code paths. Should be doable with pointers to `width/x` or `height/y`.

river/command/move.zig
```diff
@@ -0,0 +1,196 @@
...
+        },
+    }
+
+    apply(view);
+}
+
+fn apply(view: *View) void {
+    // Set the view to floating but keep the position and dimensions
+    view.pending.float = true;
+    view.float_box = view.pending.box;
```
ifreund, line 167: This should be set when switching away from not to floating.
```suggestion
```

river/command/move.zig
```diff
@@ -0,0 +1,196 @@
...
+    // Do not touch fullscreen views
+    if (view.pending.fullscreen) return null;
+
+    // Do not touch views which are the target of a cursor action
+    if (seat.input_manager.isCursorActionTarget(view)) return null;
+
+    return view;
+}
+
+fn get_output_dimensions(view: *View) Box {
```
ifreund, line 185: There are a lot of places we are doing this, prehaps an `Output.effectiveResolution()` function would be appropriate.

river/command/move.zig
```diff
@@ -0,0 +1,196 @@
...
+
+fn apply(view: *View) void {
+    // Set the view to floating but keep the position and dimensions
+    view.pending.float = true;
+    view.float_box = view.pending.box;
+
+    view.applyPending();
+}
+
+fn get_view(seat: *Seat) ?*View {
```
ifreund, line 172:
```suggestion
fn getView(seat: *Seat) ?*View {
```

ifreund (commented):
These will be nice to have

river/command/move.zig
Leon-Plickat: That is needed because otherwise the view will not keep the dimensions.

river/command/move.zig
Leon-Plickat: The effective resolution is now stored in the output.

ifreund (approved):
LGTM, Thanks!

PR #116, opened 2020-10-08
Always set fullscreen views to fully opaque

Leon-Plickat:
Although per XDG spec we technically could, we do not draw anything interesting to watch behind  fullscreened views. As such, server side transparency makes no sense for these views.

With this PR, fullscreened views always get set to an opacity of 1.0 and once they return to their previous state the correct opacity based on their focus state.


ifreund (approved):
LGTM, Thanks!

PR #118, opened 2020-10-09
Relative Pointer Stuff

Bonicgamer:
This adds the relative pointer thing.

I put everything where it needs to be and it should be good I think.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -108,15 +108,33 @@ const Mode = union(enum) {
         passthrough(self, event.time_msec);
     }
 
-    fn processMotion(self: *Self, device: *c.wlr_input_device, time: u32, delta_x: f64, delta_y: f64) void {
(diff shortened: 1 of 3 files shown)


river/Cursor.zig
```diff
@@ -108,15 +108,33 @@ const Mode = union(enum) {
...
+                c.wlr_relative_pointer_manager_v1_send_relative_motion(
+                    self.seat.input_manager.wlr_relative_pointer_manager,
+                    self.seat.wlr_seat,
+                    time,
+                    delta_x,
+                    delta_y,
+                    unaccel_dx,
+                    unaccel_dy,
+                );
                 c.wlr_cursor_move(self.wlr_cursor, device, delta_x, delta_y);
```
ifreund, line 125: This is a fair bit of duplication between the passthrough and down branches, how about we move it to an if before the switch statement? 
```zig
if (self.mode == .passthrough or self.mode == .down) {
    /// send rel. motion and cursor move
}
```

Bonicgamer:
Closing in favor of #120 since it includes Relative Pointer.

PR #120, opened 2020-10-15
Pointer Constraints

Bonicgamer:
This PR includes both pointer constraint and motion relative.

As of now this does ***not*** work with wlroot's pointer constraints example for some reason, and ***does*** work with games like Minecraft.

Todo:
- [ ] Fix not working with [wlroots pointer constraints example](https://github.com/swaywm/wlroots/blob/master/examples/pointer-constraints.c)

Edit: Forgot to mention most of the code is copied from [sway's cursor.c file](https://github.com/swaywm/sway/blob/master/sway/input/cursor.c) and modified to fix compile errors, so if you were thinking to yourself: "Hey! This code is copied from sway!", it was.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -121,22 +121,24 @@ const Mode = union(enum) {
         );
         var dx: f64 = delta_x;
         var dy: f64 = delta_y;
-        if (self.active_constraint != null and device.type == @intToEnum(c.wlr_input_device_type, c.WLR_INPUT_DEVICE_POINTER)) {
-            var sx: f64 = undefined;
-            var sy: f64 = undefined;
-            var sx_confined: f64 = undefined;
-            var sy_confined: f64 = undefined;
+        if (self.active_constraint) |active_constraint| {
+            if (device.type == .WLR_INPUT_DEVICE_POINTER) {
+                var sx: f64 = undefined;
+                var sy: f64 = undefined;
+                var sx_confined: f64 = undefined;
+                var sy_confined: f64 = undefined;
 
-            const surface: ?*c.wlr_surface = self.surfaceAt(self.wlr_cursor.x, self.wlr_cursor.y, &sx, &sy,);
+                const surface: ?*c.wlr_surface = self.surfaceAt(self.wlr_cursor.x, self.wlr_cursor.y, &sx, &sy,);
 
-            if (self.active_constraint.?.surface != surface) {
-                return;
+                if (active_constraint.surface != surface) {
+                    return;
+                }
+                if (!c.wlr_region_confine(&self.confine, sx, sy, sx + dx, sy + dy, &sx_confined, &sy_confined,)) {
+                    return;
+                }
+                dx = sx_confined - sx;
+                dy = sy_confined - sy;
             }
-            if (!c.wlr_region_confine(&self.confine, sx, sy, sx + dx, sy + dy, &sx_confined, &sy_confined,)) {
-                return;
-            }
-            dx = sx_confined - sx;
-            dy = sy_confined - sy;
         }
 
         switch (self.mode) {
@@ -249,12 +251,6 @@ const Mode = union(enum) {
         }
     }
 };
-const struct_pointer_constraint = struct {
-    cursor: *Self,
-    constraint: ?*c.wlr_pointer_constraint_v1,
-    set_region: c.wl_listener,
-    destroy: c.wl_listener,
-};
 
 const default_size = 24;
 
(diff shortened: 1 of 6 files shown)


river/Cursor.zig
```diff
@@ -611,3 +779,24 @@ fn viewSurfaceAt(output: Output, ox: f64, oy: f64, sx: *f64, sy: *f64) ?*c.wlr_s
...
+
+        const view: ?*View = View.fromWlrSurface(constraint.?.surface);
+        const cur = view.?.current;
+
+        const lx: f64 = sx + @intToFloat(f64, cur.box.x + view.?.surface_box.x);
+        const ly: f64 = sy + @intToFloat(f64, cur.box.y + view.?.surface_box.y);
+
+        const asdf = c.wlr_cursor_warp(self.wlr_cursor, null, lx, ly);
+
+        // Warp the pointer as well, so that on the next pointer rebase we don't
```
ifreund, line 798: Unlike sway, river doesn't have a pointer rebase function

river/Cursor.zig
```diff
@@ -350,6 +454,37 @@ pub fn handleViewUnmap(self: *Self, view: *View) void {
     }
 }
 
+fn checkConstraintRegion(self: *Self) void {
+    const constraint: ?*c.wlr_pointer_constraint_v1 = self.active_constraint;
+    var region: *c.pixman_region32_t = &constraint.?.region;
+    const view: ?*View = View.fromWlrSurface(constraint.?.surface);
```
ifreund, line 460: Could be something like:
```zig
if (View.fromWlrSurface(constraint.?.surface)) |view| {
    // ...
}
```

ifreund (changes requested):
Not sure why the wlroots example doesn't work yet. Here's some initial code review though. Also, please run `zig fmt` on the code.

river/Cursor.zig
```diff
@@ -241,6 +280,68 @@ listen_motion_absolute: c.wl_listener = undefined,
...
+            // has finished committing, which means that warping won't work properly,
+            // since this code will be run *after* the focus has been set.
+            // That is why we duplicate the code here.
+            if (c.pixman_region32_not_empty(&constraint.?.current.region) > 0) {
+                const tst = c.pixman_region32_intersect(&constraint.?.region,
+                    &constraint.?.surface.*.input_region, &constraint.?.current.region);
+            } else {
+                const tst = c.pixman_region32_copy(&constraint.?.region,
+                    &constraint.?.surface.*.input_region);
+            }
```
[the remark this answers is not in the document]
Bonicgamer: Yeah. I forgot to mention that most of the code is copied from sway. I added it just in case it is necessary (like the comment from sway's code says), but I'll look to see if there is a way around it.

river/Cursor.zig
Bonicgamer: Yeah, more copied code from sway (Well, comment... Why did I copy a comment?).

river/Cursor.zig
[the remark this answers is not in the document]
Leon-Plickat: I am not sure 1:1 copying and translating what sway does is the right thing to do here. While GPL can technically swallow MIT, you'd have to include the copyright notice of everyone who ever worked on that part of sway, explicitly only for that part of the code, or you'd be violating sways license (I am not sure if `Copyright (c) the Sway authors` works, but `copied from sway` definitely does not).

It may be better to re-implement this feature, especially since the differences between river and sway are not just superficial but also concern the internal stuff.

ifreund:
> Edit: Forgot to mention most of the code is copied from sway's cursor.c file and modified to fix compile errors, so if you were thinking to yourself: "Hey! This code is copied from sway!", it was.

I really can't accept code that has been directly copied from sway. Doing so would require including their license in release tarballs and mentioning their copyright. I don't think that's worth it, especially when the resulting code will be better when written from scratch in Zig with river's design choices taken into account.

Bonicgamer:
> > Edit: Forgot to mention most of the code is copied from sway's cursor.c file and modified to fix compile errors, so if you were thinking to yourself: "Hey! This code is copied from sway!", it was.
> 
> I really can't accept code that has been directly copied from sway. Doing so would require including their license in release tarballs and mentioning their copyright. I don't think that's worth it, especially when the resulting code will be better when written from scratch in Zig with river's design choices taken into account.

Oh well. I guess I'll wait for [zig-wlroots](https://github.com/swaywm/zig-wlroots) and then try again without coping sway.
But until then I'm going to use this for myself because it took me a while to make and I want to play Minecraft.

ifreund:
Thank you for your work on this, it is valuable as a prototype even though it won't be merged. Enjoy the minecraft :)

PR #122, opened 2020-10-17
Implement XDG Toplevel move and resize requests.

Leon-Plickat:
Pretty simple patch, since the code that does the actual work is already there, it just needs to be called. Using `Cursor.Mode` is a bit awkward though, since its `self` wants a `Cursor`.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -33,7 +33,7 @@ const Seat = @import("Seat.zig");
 const View = @import("View.zig");
 const ViewStack = @import("view_stack.zig").ViewStack;
 
-const Mode = union(enum) {
+pub const Mode = union(enum) {
     passthrough: void,
     down: *View,
     move: *View,
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -45,7 +45,7 @@ const Mode = union(enum) {
     },
 
     /// Enter move or resize mode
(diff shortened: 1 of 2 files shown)


river/InputManager.zig
```diff
@@ -151,3 +151,14 @@ fn handleNewInput(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
     // TODO: suport multiple seats
     self.defaultSeat().addDevice(device);
 }
+
+/// Returns a seat matching a wlr_seat.
+pub fn seatFromWlrSeat(self: *Self, wlr_seat: *c.wlr_seat) ?*Seat {
```
ifreund, line 156: We store a pointer to the river `Seat` in the `data` field of the `wlr_seat`, this function shouldn't be needed.

ifreund (commented):
> Using Cursor.Mode is a bit awkward though, since its self wants a Cursor.

Yeah, this makes me realize that the functions should probably be members of `Cursor` not `Cursor.Mode`. Feel free to move them in a separate commit if you like.

Leon-Plickat:
Moved the functions from `Cursor.Mode` to `Cursor`; Now they are straight forward to use.

ifreund (approved):
LGTM, thanks!

PR #123, opened 2020-10-17
Bugfix: Update view.float_box when setting a floating view to fullscreen

Leon-Plickat:
Currently,  if a floating view is moved via the mouse, then enters fullscreen, then leaves fullscreen, it returns to its original, pre-move, position/dimensions. This fixes that.


ifreund (approved):
LGTM, thanks!

PR #124, opened 2020-10-18
Some mapping things.

Leon-Plickat:
**First patch:** This adds the possibility to use the modifier string "None" (evals to 0) to create mappings without modifiers. This is simply more user friendly then the current way of doing this.

---

**Second patch:** This removes the map checking for translated keysyms. I am pretty sure the raw keysyms cover every possible mapping, so it's simply not necessary.

### DIFF
--- a/river/command/map.zig
+++ b/river/command/map.zig
@@ -29,7 +29,7 @@ const modifier_names = [_]struct {
     name: []const u8,
     modifier: u32,
 }{
-    .{ .name = "", .modifier = 0 },
+    .{ .name = "None", .modifier = 0 },
     .{ .name = "Shift", .modifier = c.WLR_MODIFIER_SHIFT },
     .{ .name = "Lock", .modifier = c.WLR_MODIFIER_CAPS },
     .{ .name = "Control", .modifier = c.WLR_MODIFIER_CTRL },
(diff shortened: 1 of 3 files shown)


river/command/map.zig
```diff
@@ -30,6 +30,7 @@ const modifier_names = [_]struct {
     modifier: u32,
 }{
     .{ .name = "", .modifier = 0 },
+    .{ .name = "none", .modifier = 0 },
```
ifreund, line 33: Should this be `None` for consistency with the other names from xkbcommon?

Leon-Plickat: Probably.

river/command/map.zig
```diff
@@ -30,6 +30,7 @@ const modifier_names = [_]struct {
     modifier: u32,
 }{
     .{ .name = "", .modifier = 0 },
```
ifreund, line 32: Sorry I didn't catch this in the first round, but we really shouldn't have two ways to do the same thing.
```suggestion
    .{ .name = "", .modifier = 0 },
```

Leon-Plickat: Agreed.

ifreund (approved):
Thanks!

PR #127, opened 2020-10-21
Make XwaylandView.zig a little more like XdgToplevel.zig

Bonicgamer:
This makes XwaylandView.zig more like XdgToplevel.zig mostly in the "making windows float" area, which fixes an issue I was having with Kicad where if you open the 'Edit Value' window a second time, Kicad would freeze.

### DIFF
--- a/river/XwaylandView.zig
+++ b/river/XwaylandView.zig
@@ -165,6 +165,28 @@ fn handleMap(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
     view.float_box.y = std.math.max(0, @divTrunc(@intCast(i32, view.output.usable_box.height) -
         @intCast(i32, view.float_box.height), 2));
 
+    const size_hints = self.wlr_xwayland_surface.size_hints;
+    const has_fixed_size = size_hints.*.min_width != 0 and size_hints.*.min_height != 0 and
+        (size_hints.*.min_width == size_hints.*.max_width or size_hints.*.min_height == size_hints.*.max_height);
+    const app_id: [*:0]const u8 = if (self.wlr_xwayland_surface.class) |id| id else "NULL";
+
+    if (self.wlr_xwayland_surface.parent != null or has_fixed_size) {
+        // If the toplevel has a parent or has a fixed size make it float
+        view.current.float = true;
+        view.pending.float = true;
+        view.pending.box = view.float_box;
+    } else {
+        // Make views with app_ids listed in the float filter float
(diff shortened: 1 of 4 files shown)


river/XwaylandView.zig
```diff
@@ -184,7 +205,7 @@ fn handleCommit(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
     const self = @fieldParentPtr(Self, "listen_commit", listener.?);
     const view = self.view;
 
-    view.surface_box = Box{
+    const new_box = Box{
```
ifreund, line 208: Why change this? This is now assigned in both branches.

river/XwaylandView.zig
```diff
@@ -193,12 +214,21 @@ fn handleCommit(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
-        if (!view.pending.float and !view.pending.fullscreen)
+        if (view.shouldTrackConfigure())
             view.output.root.notifyConfigured()
-        else
+        else {
+            const view_tags_changed = view.pending.tags != view.current.tags;
             view.current = view.pending;
+            view.commitOpacityTransition();
+            if (view_tags_changed) view.output.sendViewTags();
+        }
```
ifreund, line 230: This is exactly the same as what is in `XdgToplevel.zig`, we should probably extract it to a new function in `View.zig`

river/XwaylandView.zig
```diff
@@ -165,6 +165,27 @@ fn handleMap(listener: ?*c.wl_listener, data: ?*c_void) callconv(.C) void {
...
         @intCast(i32, view.float_box.height), 2));
 
+    const size_hints = self.wlr_xwayland_surface.size_hints;
+    const has_fixed_size = size_hints.*.min_width != 0 and size_hints.*.min_height != 0 and
+        (size_hints.*.min_width == size_hints.*.max_width or size_hints.*.min_height == size_hints.*.max_height);
+    const app_id: [*:0]const u8 = if (self.wlr_xwayland_surface.class) |id| id else "NULL";
+
+    if (self.wlr_xwayland_surface.parent != null or has_fixed_size) {
+        // If the toplevel has a parent or has a fixed size make it float
+        view.pending.float = true;
```
ifreund, line 175: I believe we should set `view.current.float = true` here as well.

ifreund (commented):
Looks reasonable, just a few things to clean up.

ifreund (approved):
LGTM, thanks!

PR #128, opened 2020-10-21
map,map-pointer: Make mode argument a list of modes

maringuu:
With the introduction of the "locked" mode I see myself copy pasting a lot in the config file for keybindings that should be in both the normal mode and the locked mode.
I figured that it would be a good idea to be able to specify a mapping for multiple modes.

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -134,9 +134,10 @@ that tag 1 through 9 are visible.
 	but let the currently focused window in focus.
 	When set to _strict_ this is not the case. The focus will be updated on every cursor movement.
 
-*map* [-release] _mode_ _modifiers_ _key_ _command_
-	_mode_ is either "normal" (the default mode), "locked" (the mode entered when
-	an input inhibitor such as a lock screen is active) or a mode created with *declare-mode*.
+*map* [-release] _modes_ _modifiers_ _key_ _command_
+	_modes_ is a comma seperated list whose members are one of "normal" (the default mode),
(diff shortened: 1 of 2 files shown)


Leon-Plickat:
While I am not opposed to this, I wonder how much sense it actually makes. What keybinds does one actually want on a lockscreen? Media, volume, brightness, so probably no more than 10(-ish)?

For such a low amount of keybinds, why not simply use a loop in your config script?

```sh
for mode in normal locked
do
        # Put all duplicate binds here
        riverctl map $mode <whatever>
done
````

**Comment by maringuu** [comment]
The loop is a good idea.

If you had multiple modes and many keybindings that you want to have across a different subset of the modes this would make sense.

```
riverctl declare-mode mode1
riverctl declare-mode mode2
riverctl declare-mode mode3
riverctl declare-mode mode4

map mode1,mode2 <foo>
map mode2,mode3 <bar>
map mode1,mode2,mode3,mode4 <foobar>
```

But I can't think of any real use case for that many modes and bindings shared accross modes. Maybe you could want to exit from any mode but that's only one binding.

ifreund:
I don't see a strong use case for this, a for loop over modes in the startup script seems like a fine solution for the few cases where it makes sense for bindings to be shared across modes. Perhaps we should add an example doing this to the `config.sh` in `contrib`?

Leon-Plickat:
> Perhaps we should add an example doing this to the config.sh in contrib?

Probably for the media keys (like `XF86AudioRaiseVolume`), then we also have an example for the `None` modifier and media keys.

PR #129, opened 2020-10-22
Add mediakey maps for locked and normal mode to example config

Leon-Plickat:
The example config now additionally demonstrates how to use a shell loop
to create a mapping in multiple modes, how mediakeys can be mapped and
how to use the "None" modifier.

---

As discussed in #128


Leon-Plickat:
I am not exactly sure what programs would make good examples for these mappings, so I just included the ones I use myself.

ifreund (approved):
LGTM, thanks!

Issue #12, opened 2020-04-27
Suggestions for new window commands.


lazy-dolphin:
- commands to move a window up and down the stack
- commands to spawn a window in a location next or previous to the currently focused window in the stack where the window to spawn is given as an argument.

ifreund:
> commands to move a window up and down the stack

This seems like a very reasonable thing to add.

> commands to spawn a window in a location next or previous to the currently focused window in the stack where the window to spawn is given as an argument.

I think it makes most sense to implement this as an option determining where new views should be spawned. Taking inspiration from [dwm's patches](https://dwm.suckless.org/patches/), a list of possible values for an `attach` option might be:

- `top` (current behavior)
- `bottom` (spawn at the bottom of the stack)
- `above` (spawn above the current focus)
- `below` (spawn below the current focus)
- `aside` (spawn as the first non-master view)

lazy-dolphin:
The reason for me suggesting separate spawn commands/arguments is that it is much more flexible than a setting to spawn all windows in a particular position. For instance, a use case for this would be mapping two different bindings to `spawning a terminal as a slave window` and `spawning a terminal as a master window`.

ifreund:
this could also be done with
```sh
riverctl map Mod4 X spawn 'riverctl set-option attach top && alacritty'
riverctl map Mod4 Y spawn 'riverctl set-option attach aside && alacritty'
```
```

ifreund:
`attach top|bottom` and `swap next|previous` have been implement. Possibly implementing more attach modes can be discussed in a new issue if necessary

benbuch:
I can't figure out how to achieve this? How would I create a window next to master window? lets say am having firefox as the master window and now i want to spawn a new foot window and have the focus on this new foot window?

benbuch:
that only works ones. the patch lazy-dolpin refers to https://dwm.suckless.org/patches/attachaside/ will keep spawning windows next to master and not just in the bottom of the stack

novakne:
There is only the option to spawn to the top or the bottom of the stack in river

PR #130, opened 2020-10-22
Introduce importance command

maringuu:
The `importance` command simply swaps the focused node with the node above/below in the view stack.
I am open to other name suggestions.

I introduced a rather complicated looking `ViewStack.swap` function.
I'm not sure if this is good or not since it is only used in one place.
What are your thought about this?

----
Addresses one part of #12

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -72,6 +72,12 @@ used to control and configure river.
 	_shell_command_ if you do not want special characters to get
 	interpreted by your shell before the command gets passed to _/bin/sh_.
 
+*swap* *next*|*previous*
+	Swap the focused window with the next/previous visible non-floating window.
+	When the focused view is the first view there is no previous view.
+	In this case *previous* swaps with the last view.
+	*next* behaves analogous.
+
 *toggle-float*
 	If the focused view is floating, make it tiled. If it is tiled, make
 	it floating.
--- a/river/command/swap.zig
+++ b/river/command/swap.zig
@@ -1,0 +1,72 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 Marten Ringwelski
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const Error = @import("../command.zig").Error;
+const Direction = @import("../command.zig").Direction;
+const Seat = @import("../Seat.zig");
+const View = @import("../View.zig");
+const ViewStack = @import("../view_stack.zig").ViewStack;
+
+/// Swap the currently focused view with either the view higher or lower in the visible stack
+pub fn swap(
+    allocator: *std.mem.Allocator,
+    seat: *Seat,
+    args: []const []const u8,
+    out: *?[]const u8,
+) Error!void {
+    if (args.len < 2) return Error.NotEnoughArguments;
+    if (args.len > 2) return Error.TooManyArguments;
+
+    if (seat.focused != .view)
+        return;
+
+    // Filter out everything that is not part of the current layout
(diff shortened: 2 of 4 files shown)


doc/riverctl.1.scd
```diff
@@ -76,6 +76,9 @@ used to control and configure river.
 	If the focused view is floating, make it tiled. If it is tiled, make
 	it floating.
 
+*importance* *up*|*down*
+	Move the focused view up or down in the view stack.
```
Leon-Plickat, line 80: Name suggestion: `swap`

Needs better documentation. While the sentence is not incorrect, it is better to describe what happens *on screen* rather than what happens in an internal data structure (which is totally opaque to the user). I propose "Swap the focused window with the next/previous visible non-floating window".

river/command/importance.zig
```diff
@@ -0,0 +1,64 @@
...
+    if (args.len < 2) return Error.NotEnoughArguments;
+    if (args.len > 2) return Error.TooManyArguments;
+
+    if (seat.focused != .view)
+        return;
+
+    // Filter out everything that is not part of the current layout
+    if (seat.focused.view.pending.float or seat.focused.view.pending.fullscreen) return;
+
+    const direction = std.meta.stringToEnum(enum { up, down }, args[1]) orelse return Error.UnknownOption;
```
Leon-Plickat, line 41: It is probably better to use `Command.Direction` here instead of the anon. struct; It is more idiomatic and has a specifif error.

river/command/importance.zig
```diff
@@ -0,0 +1,64 @@
...
+
+    const focused_node = @fieldParentPtr(ViewStack(View).Node, "view", seat.focused.view);
+    const output = seat.focused_output;
+    var it = if (direction == .up) ViewStack(View).iter(focused_node, .reverse, output.pending.tags, filter)
+        else if (direction == .down) ViewStack(View).iter(focused_node, .forward, output.pending.tags, filter)
+        else unreachable;
```
Leon-Plickat, line 47: `direction` can only be `.up` or `.down`, so only a single conditional is required.

Also I think you should use `output.current.tags` here.

river/view_stack.zig
```diff
@@ -108,6 +110,50 @@ pub fn ViewStack(comptime T: type) type {
             }
         }
 
+        /// Swap the nodes one and two.
+        /// pointers to Node.T will point to the same data as before
+        pub fn swap(self: *Self, one: *Node, two: *Node) void {
```
Leon-Plickat, line 115: Style: `one` and `two` are not very good variable names IMO. For such functions, `a` and `b` are more common.

Leon-Plickat (commented):
Overall probably something I'd find useful.

Although it would be nice if it would wrap.

doc/riverctl.1.scd
ifreund: `swap` works for me.

river/command/importance.zig
maringuu: I took inspiration from `zoom.zig` since both `swap` and `zoom` manipulate the viewstack.
For zoom pending instead of current is used everywhere.

When we use the current state here should we also use it in the filter function?

doc/riverctl.1.scd
maringuu: > Although it would be nice if it would wrap.

Added wraping

maringuu (commented):
Accidently clicked somewhere and started this review. Cant find the abort button...

river/command/swap.zig
```diff
@@ -0,0 +1,72 @@
...
+
+    // Filter out everything that is not part of the current layout
+    if (seat.focused.view.pending.float or seat.focused.view.pending.fullscreen) return;
+
+    const direction = std.meta.stringToEnum(Direction, args[1]) orelse return Error.InvalidDirection;
+
+    const focused_node = @fieldParentPtr(ViewStack(View).Node, "view", seat.focused.view);
+    const output = seat.focused_output;
+    var it = if (direction == .next) ViewStack(View).iter(focused_node, .forward, output.pending.tags, filter)
+        else ViewStack(View).iter(focused_node, .reverse, output.pending.tags, filter);
```
ifreund, line 47: This would be more readable imo, same for the next two lines. 
```suggestion
    var it = ViewStack(View).iter(
        focused_node,
        if (direction == .next) .forward else .reverse,
        output.pending.tags,
        filter,
    );
```

river/view_stack.zig
```diff
@@ -15,6 +15,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+const std = @import("std");
+
```
ifreund, line 19: This isn't used...
```suggestion
```

river/command/swap.zig
maringuu: That looks a lot better. I am really not used to zigs ability to place everything everywhere.

ifreund:
Thanks!

PR #131, opened 2020-10-23
Mapping.zig: Save allocator in struct

maringuu:
Is there a reason why the allocator is not saved in the struct?
Right now it would be possible to call `deinit` with another allocator than the one used for `init` which would result in undefined behavior.

### DIFF
--- a/river/Mapping.zig
+++ b/river/Mapping.zig
@@ -20,6 +20,7 @@ const Self = @This();
 const std = @import("std");
 
 const c = @import("c.zig");
+const util = @import("util.zig");
 
 keysym: c.xkb_keysym_t,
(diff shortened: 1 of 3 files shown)


ifreund (changes requested):
There is a reason: storing the allocator pointer wastes 4 bytes per mapping. It may seem small but stuff like this adds up across the codebase. If you want to make it "safer" perhaps we should hardcode it to `util.gpa`

maringuu:
Yeah, hardcoding sounds better then.
I updated the commit.

ifreund:
Thanks, edited slightly and landed in 16c8752de2a0ff1a2c11914e7fc969e490556a8d

PR #132, opened 2020-10-24
Implement unmap and unmap-pointer

maringuu:
Edit:
I thought this was discussed somewhere but I can't remember where.
An obvious usecase is remapping keys in a running session.

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -200,6 +200,10 @@ that tag 1 through 9 are visible.
 	Removes the mapping defined by the arguments *-release*, *modifiers* and *key* from *mode*.
 	See *map* for an explanation of the arguments.
 
+*unmap-pointer* _mode_ _modifiers_ _button_
+	Removes the mapping defined by the arguments *modifiers* and *button* from *mode*.
+	See *map-pointer* for an explanation of the arguments.
+
 *view-padding* _pixels_
 	Set the padding around the edge of each view to _pixels_.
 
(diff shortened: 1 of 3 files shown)


river/command/map.zig
```diff
@@ -211,3 +218,76 @@ fn parseOptionalArgs(args: []const []const u8) OptionalArgsContainer {
...
+            break i;
+        }
+    } else {
+        out.* = try std.fmt.allocPrint(
+            allocator,
+            "there is no mapping for modifiers '{}' and keysym '{}'",
+            .{ args[2 + offset], args[3 + offset] },
+        );
+        return Error.Other;
+    };
```
ifreund, line 254: Could we de-duplicate this check across `map`/`unmap` and `map-pointer`/`unmap-pointer`?

maringuu: Done.
It didn't deduplicate that much because the error messages are different.

maringuu:
Rebased.

doc/riverctl.1.scd
```diff
@@ -196,6 +196,14 @@ that tag 1 through 9 are visible.
 *outer-padding* _pixels_
 	Set the padding around the edge of the screen to _pixels_.
 
+*unmap* [-release] _mode_ _modifiers_ _key_
```
maringuu, line 199: I have mixed feelings about `-release`.
It looks like an optional argument to `unmap` (well it kind of is) but if we had other arguments that do not specify the mapping to unmap further then it would look wired.
On the plus side the syntax is the same as with `map`.

What are your thoughts?

ifreund: I'm ok with how this is as it parallels `map` nicely. I also can't think of a reasonable alternative, do you have something else in mind?

maringuu: No I don't. lets keep it as it is!

ifreund (approved):
LGTM, Thanks!

PR #136, opened 2020-10-27
Implement wlr-output-power-management-unstable-v1 protocol

maringuu:
This is my first protocol implementation of any kind.

I need help with one little thing which is that when disabling and enabling afterwards I get an error.
Can you point me to anything I should investigate?
```
info: (server) initializing
debug: (server) new output eDP-1
debug: (server) new output DP-1
info: (server) running...
debug: (server) new xdg_toplevel
debug: (server) view 'Alacritty' mapped
debug: (transaction) started transaction with 1 pending configure(s)
debug: (server) Disabling dmps for output eDP-1
00:00:08.531 [ERROR] [types/wlr_output.c:567] Basic output test failed
debug: (server) Enabling dmps for output eDP-1
info: (server) shutting down
debug: (server) output 'eDP-1' destroyed
debug: (server) output 'DP-1' destroyed
```

For reference here is sways code: https://github.com/swaywm/sway/blob/0cb9282aeebddeed0b5259f153f98a45311053a5/sway/desktop/output.c#L1036

For testing I hacked together a little wayland client which will turn off the first output it finds and will turn it back on three seconds later. https://git.sr.ht/~maringuu/wlr-dpms

Edit: It seems that I "try to commit a buffer to a disabled output"

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -153,6 +153,7 @@ const ScanProtocolsStep = struct {
         const protocol_dir_paths = [_][]const []const u8{
             &[_][]const u8{ protocol_dir, "stable/xdg-shell/xdg-shell.xml" },
             &[_][]const u8{ "protocol", "wlr-layer-shell-unstable-v1.xml" },
+            &[_][]const u8{ "protocol", "wlr-output-power-management-unstable-v1.xml" },
             &[_][]const u8{ "protocol", "river-control-unstable-v1.xml" },
             &[_][]const u8{ "protocol", "river-status-unstable-v1.xml" },
         };
@@ -160,6 +161,7 @@ const ScanProtocolsStep = struct {
         const server_protocols = [_][]const u8{
             "xdg-shell",
             "wlr-layer-shell-unstable-v1",
+            "wlr-output-power-management-unstable-v1",
             "river-control-unstable-v1",
             "river-status-unstable-v1",
         };
--- a/protocol/wlr-output-power-management-unstable-v1.xml
+++ b/protocol/wlr-output-power-management-unstable-v1.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_output_power_management_unstable_v1">
+  <copyright>
+    Copyright © 2019 Purism SPC
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice (including the next
+    paragraph) shall be included in all copies or substantial portions of the
+    Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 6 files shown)


ifreund:
Haven't taken a close look at this, but perhaps we now need a check if the output is enabled in the frame handler? You could use gdb and run river in a nested session to see which `wlr_output_commit` call is causing this error.

maringuu:
Thanks for your input!
It seems like you are right and we need a check whether the output is enabled in the frame handler. I just tested it and the error dissapears.
Also I found out that debugging in a nested session is useless because a wayland output can't be disabled.
Debugging a session that renders to drm is not usefull either because no frame events should be send. But also I can't attach gdb to river because I use alpine and gcompat.


The commit that fails is the one in the `render.zig`. The commit in `handleOutputPowerManagement` succeeds.
According to the wlroots docs an disabled output will not send frame events.
What I don't understand is why a frame event is send after disabling the output succeeded and was commited.

I'll investigate further the next days when I have time.

maringuu:
I am now convinced that this is a wlroots bug.
The check in `handleFrame` should not have to be there since no frame events should be send when the output is disabled.

I opened a PR at wlroots to fix this.
https://github.com/swaywm/wlroots/pull/2456

maringuu:
The wlroots bug is fixed now.
I guess we'll wait for the next wlroots release then.

ifreund:
The next wlroots release should come in the next week or two based on when wayfire is ready, so waiting for that would be reasonable. We could also add the check that will be made unnecessary by that release with a `TODO(wlroots): ...` comment to be removed upon updating.

Nice work finding and fixing the issue in wlroots!

maringuu:
> We could also add the check that will be made unnecessary by that release with a `TODO(wlroots): ...` comment to be removed upon updating.

Good idea! Pushed it just now and also rebased.

> Nice work finding and fixing the issue in wlroots!

Thanks!

ifreund (approved):
LGTM, Thanks!

Leon-Plickat:
@dkwo something like this: https://sr.ht/~leon_plickat/wlopm/

dkwo:
@Leon-Plickat Is there a way to do DPMS (ie. blank the screen but keep workspaces as-is), other than turn off output?
See also #571

Leon-Plickat:
@dkwo Yes, using the tool I linked to. It will turn off the output, not destroy it.

dkwo:
Just to make sure I understand:
is this tool working differently than wlr-randr, which I'm currently using?

dkwo:
@Leon-Plickat Thanks, indeed it works, where `wlr-randr` was not working!
I wrote a template to build this on Void linux, does it look fine to you?
(Perhaps at some point tag a release, so one could package it there :)
```
# Template file for 'wlopm'
pkgname=wlopm
version=git20220318
revision=1
_commit=27b7ccd1f1f2f5f2986508890805934c3e7614ee
wrksrc=${pkgname}-${_commit}
build_style=gnu-makefile
makedepends="wayland-devel"
short_desc="Wayland output power management"
maintainer="dkwo ..."
license="GPL-3.0-or-later"
homepage="https://git.sr.ht/~leon_plickat/wlopm"
distfiles="https://git.sr.ht/~leon_plickat/${pkgname}/archive/${_commit}.tar.gz"
checksum=3f96c7382e2aaa4d5fdcdec12a2ad81eb623915dd8207a9b2d614a644c0815d1
make_check=no # No test suite
```

dkwo:
@Leon-Plickat Btw, it would be nice to have both options in the same tool: right now i'm using wlr-randr if I need to destroy the output (e.g. in a multi-monitor setup, where I don't want the laptop monitor), and wlopm if I want to preserve it (e.g. in the same setup, to do dpms/turn off screen).

PR #137, opened 2020-10-30
Remove most of the hack which works around @cImport and flexible arrays

maringuu:
While working on implementing the virtual pointer protocol, I noticed that most of the hack is not needed.

The problem is the `wlr/render/wlr_renderer.h`.
None of  `wlr/backend.h`, `wlr/backend/noop.h` and `wlr/backend/multi.h` include `wlr/render/wlr_renderer.h` so it should be safe to `@cInclude` them and remove the wrapper functions.

### DIFF
--- a/include/bindings.c
+++ b/include/bindings.c
@@ -1,7 +1,5 @@
 #define WLR_USE_UNSTABLE
 #include <wlr/backend.h>
-#include <wlr/backend/noop.h>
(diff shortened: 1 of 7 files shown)


ifreund (approved):
LGTM, can't wait to completely get rid of these when I finally get river switched over to [zig-wlroots](https://github.com/swaywm/zig-wlroots)

PR #138, opened 2020-10-30
Implement virtual_keyboard_unstable_v1 protocol

maringuu:
Can be tested with [wayvnc](https://github.com/any1/wayvnc). The tldr is install it and run `wayvnc 0.0.0.0`. 

You might notice that the keyboard input works fine but the rendering does not update. I don't know yet why this happens but I am quite sure that this problem is unrelated to this PR.

Edit: wayvnc also needs virtual_pointer_unstable_v1.
I inplemented it in https://github.com/ifreund/river/pull/139.

### DIFF
--- a/river/InputManager.zig
+++ b/river/InputManager.zig
@@ -1,6 +1,7 @@
 // This file is part of river, a dynamic tiling wayland compositor.
 //
 // Copyright 2020 Isaac Freund
+// Copyright 2020 Marten Ringwelski
 //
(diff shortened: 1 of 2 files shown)


ifreund:
Thanks for implementing this! Github doesn't seem to be smart enough to figure out that this is merged (773c5ff687742f3b56132ce9e7a2f21627e33714), probably since I had to resolve a few conflicts with the virtual_pointer PR.

Issue #49, opened 2020-06-12
river-status: send input mode


ifreund:
We should send the current input mode when the mode changes as an event on the `river_seat_status` object.

ifreund:
Implemented in d657dc791bb4c771c9429a2d03ab021f1cb8075c

PR #140, opened 2020-10-30
Updates to river-status protocol

Leon-Plickat:
1. Send `wl_output.done` after all `river_output_status` properties have been sent to allow atomic changes. (Allows clients like status bars to only render on `wl_output.done`, which reduces complexity and is more efficient).

2. Send more information in `river_seat_status.focused_view`, which may be useful to clients.

3. Send event when layout changed. This allows clients like status bars to display the name of the layout.


ifreund:
We should probably have an event relaying the current keybind mode to clients as well (#49).

PR #142, opened 2020-11-01
code: Move wlr_output_layout_add from Output.init to Root.addOutput

maringuu:
This is needed for my upcoming PR to implement `wlr_output_management_unstable_v1` (Which should be opened tomorrow).
I'm not quite sure if my argumentation is right so feel free to ignore this until I opend the PR for output management.

The difference this PR makes is that now we can listen to `wlr_output_layout.change` and whenever this change is the addition of an output we now can be sure that the output is a fully initialized member of `Root.outputs`.

Currently the output is added to `Root.outputs` only after it is added to the `wlr_output_layout` which makes it hard for me to iterate over all outputs.
While typing I this noticed that I could also iterate over `wlr_output_layout.outputs`. But this feels wrong to me.

Also this introduces some some duplication of the call to `wlr_output_layout_add_auto` because it now needs to be called manually. Imo this is fine since the noop_output is a special case anyway.

What are your thoughts?

### DIFF
--- a/river/Root.zig
+++ b/river/Root.zig
@@ -102,6 +102,12 @@ pub fn addOutput(self: *Self, wlr_output: *c.wlr_output) void {
     };
     self.outputs.append(node);
 
+    // Add the new output to the layout. The add_auto function arranges outputs
+    // from left-to-right in the order they appear. A more sophisticated
+    // compositor would let the user configure the arrangement of outputs in the
+    // layout. This automatically creates an output global on the wl_display.
+    c.wlr_output_layout_add_auto(self.wlr_output_layout, wlr_output);
+
     // if we previously had no real outputs, move focus from the noop output
     // to the new one.
     if (self.outputs.len == 1) {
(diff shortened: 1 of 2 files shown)


river/Root.zig
```diff
@@ -74,6 +74,7 @@ pub fn init(self: *Self, server: *Server) !void {
 
     const noop_wlr_output = c.wlr_noop_add_output(server.noop_backend) orelse return error.OutOfMemory;
     try self.noop_output.init(self, noop_wlr_output);
+    c.wlr_output_layout_add_auto(self.wlr_output_layout, noop_wlr_output);
```
ifreund, line 77: We shouldn't add the noop output to the layout. It is an implementation detail and should never be advertised.

ifreund (approved):
The reasoning makes sense to me, the `wlr_output_layout` is a member of `Root` anyways

ifreund:
Thanks!

PR #144, opened 2020-11-02
code: Replace deprecated std.fmt.trim with std.mem.trim

maringuu:
Related: https://github.com/ifreund/zig-wayland/pull/10


ifreund (approved):
Thanks!

PR #149, opened 2020-11-10
Repeat Rate Command

### LINKED ISSUE #70: Make repeat delay and repeat rate configurable
I'm pretty sure this is planned at some point, but couldn't find a tracking issue. Adding one, since I'm very interested in this.

Bonicgamer:
This adds a 'set-repeat' command which sets the repeat rate and delay.

Most of the code is copied from River itself (not Sway this time), but one day I promise I'm going to make something without coping code.

Closes #70

### DIFF
--- a/river/command/set_repeat.zig
+++ b/river/command/set_repeat.zig
@@ -1,6 +1,6 @@
 // This file is part of river, a dynamic tiling wayland compositor.
 //
-// Copyright 2020 Isaac Freund
+// Copyright 2020 Bonicgamer
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
@@ -22,7 +22,7 @@ const c = @import("../c.zig");
 const Error = @import("../command.zig").Error;
 const Seat = @import("../Seat.zig");
 
-/// Close the focused view, if any.
+/// Set the repeat rate and delay for all keyboards.
 pub fn setRepeat(
     allocator: *std.mem.Allocator,
     seat: *Seat,
@@ -35,6 +35,8 @@ pub fn setRepeat(
     const rate = try std.fmt.parseInt(i32, args[1], 10);
     const delay = try std.fmt.parseInt(i32, args[2], 10);
 
-    const keyboard: *c.wlr_keyboard = c.wlr_seat_get_keyboard(seat.wlr_seat);
-    c.wlr_keyboard_set_repeat_info(keyboard, rate, delay);
+    var it = seat.keyboards.first;
+    while (it) |node| : (it = node.next) {
+        c.wlr_keyboard_set_repeat_info(node.data.wlr_keyboard, rate, delay);
+    }
 }
--- a/river/command/set_repeat.zig
+++ b/river/command/set_repeat.zig
@@ -1,6 +1,6 @@
 // This file is part of river, a dynamic tiling wayland compositor.
 //
-// Copyright 2020 Bonicgamer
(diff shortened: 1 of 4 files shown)


river/command/set_repeat.zig
```diff
@@ -0,0 +1,40 @@
...
+    out: *?[]const u8,
+) Error!void {
+    if (args.len < 3) return Error.NotEnoughArguments;
+    if (args.len > 3) return Error.TooManyArguments;
+
+    const rate = try std.fmt.parseInt(i32, args[1], 10);
+    const delay = try std.fmt.parseInt(i32, args[2], 10);
+
+    const keyboard: *c.wlr_keyboard = c.wlr_seat_get_keyboard(seat.wlr_seat);
+    c.wlr_keyboard_set_repeat_info(keyboard, rate, delay);
```
ifreund, line 39: It would probably be more correct to iterate over `Seat.keyboards` and set the repeat rate for all keyboards. Ideally we should allow targeting specific keyboards by identifier in the future but that ties into more general input config and we don't yet have the infrastructure for that.

Leon-Plickat:
Overall a good commit.

However I see one big problem:  This PR just naively configures the rates for the wlr_keyboard of the singular seat and therefore for all physical keyboards. While this does work, I think this goes in the wrong direction, as per-device / per-seat configuration is important IMO.

Also IIRC we came up with the idea of having input configuration being handled by some external program, just like output configuration. Either way, river should either have *all* or *none* of the possible input configurations.

(Also interesting to see that wlroots handles key repeats. I was under the impression a compositor has to manually hook up a timer, but it makes sense I guess.)

river/command/set_repeat.zig
```diff
@@ -0,0 +1,40 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const c = @import("../c.zig");
+
+const Error = @import("../command.zig").Error;
+const Seat = @import("../Seat.zig");
+
+/// Close the focused view, if any.
```
Leon-Plickat, line 25: This comment is wrong.

river/command/set_repeat.zig
```diff
@@ -0,0 +1,40 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 Isaac Freund
```
Leon-Plickat, line 3: **Your** name goes here.

ifreund:
>(Also interesting to see that wlroots handles key repeats. I was under the impression a compositor has to manually hook up a timer, but it makes sense I guess.)

This is for sending the `repeat_info` event to client so they know how to implement key repeat.

river/command/set_repeat.zig
```diff
@@ -0,0 +1,42 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 Bonicgamer
```
ifreund, line 3: Feel free to add yourself to AUTHORS if you would like credit, just pushed the changes I proposed on irc to master so this should be:
```suggestion
// Copyright 2020 The River Developers
```

Bonicgamer: I've decided after sitting in front of my computer screen for about 30 minutes that I'm going to not add myself to the AUTHORS file since most of the code was copied from River.

Once when I make something without copying code then I will add myself to the AUTHORS file.

ifreund (approved):
Thanks for being patient, Looks good to me!

Issue #147, opened 2020-11-04
Default location for startup script


daurnimator:
Add a default location to look for startup scripts, e.g. `~/.config/river/startup`
Possibly falling back to `/etc/river/startup`.

ifreund:
If we do this we should probably respect `$XDG_CONFIG_HOME` as well. I'm thinking that this is an acceptable level of additional complexity over just having the command line flag but I'll have to think a bit on what I want the name to be.

Leon-Plickat:
In that case I recommend explicitly not using any extensions like `.sh` for the default path, people may use something other than a shell script.

ifreund:
> In that case I recommend explicitly not using any extensions like `.sh` for the default path, people may use something other than a shell script.

Agreed, my favorite so far is `~/.config/river/init`.

ifreund:
Implemented in #152

Issue #146, opened 2020-11-04
Install example config to /etc/river/init


daurnimator:
To help new users get started it would be helpful to ship an example config which users could use to derive their own.

Leon-Plickat:
Errr... isn't `contrib/config.sh` exactly that?

ifreund:
`contrib` usually is for stuff that should not be included when packaging a program. This issue can be closed when we move the example config from contrib to doc and install it to `/etc/river/whatever` in our build.zig.

ifreund:
Implemented in c7bc47d6d4ca11ad0cc3a0216b8e45d232fc30fa

PR #152, opened 2020-11-17
Add default config location

Leon-Plickat:
As discussed in #147.

With this patch, if a config path is not provided by user, river tries `$XDG_CONFIG_HOME/river/init` and `/etc/river/init`.

#### TODO
* [x] Find out how `std.os.access()` works (definitely not like the C version).
* [x] Find out why zig complains about freeing `startup_command`.

### DIFF
--- a/river/main.zig
+++ b/river/main.zig
@@ -32,8 +32,33 @@ const usage: []const u8 =
     \\
 ;
 
+fn testConfigPath(comptime fmt: []const u8, args: anytype) std.fmt.AllocPrintError!?[:0]const u8 {
+    const path = try std.fmt.allocPrintZ(util.gpa, fmt, args);
+    std.os.access(path, std.os.X_OK) catch {
+        util.gpa.free(path);
+        return null;
+    };
+    return path;
+}
+
+fn getStartupCommand() std.fmt.AllocPrintError!?[:0]const u8 {
+    if (std.os.getenv("XDG_CONFIG_HOME")) |xdg_config_home| {
+        if (try testConfigPath("{}/river/init", .{xdg_config_home})) |path| {
+            return path;
+        }
+    } else if (std.os.getenv("HOME")) |home| {
+        if (try testConfigPath("{}/.config/river/init", .{home})) |path| {
+            return path;
+        }
+    }
+    if (try testConfigPath("/etc/river/init", .{})) |path| {
+        return path;
+    }
+    return null;
+}
+
 pub fn main() anyerror!void {
-    var startup_command: ?[*:0]const u8 = null;
+    var startup_command: ?[:0]const u8 = null;
     {
         var it = std.process.args();
         // Skip our name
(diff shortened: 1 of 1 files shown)


river/main.zig
```diff
@@ -32,7 +32,21 @@ const usage: []const u8 =
     \\
 ;
 
+fn getStartupCommand(allocator: *std.mem.Allocator) anyerror!?[*:0]const u8 {
+    if (std.os.getenv("XDG_CONFIG_HOME")) |xdg_config_home| {
+        const path = try std.fmt.allocPrintZ(allocator, "{}/river/init", .{xdg_config_home});
+        if (try std.os.access(path, 1)) return path; // 1 == X_OK
```
ifreund, line 38:
```suggestion
        if (try std.os.access(path, std.os.X_OK)) return path;
```

ifreund (commented):
For using `os.access()` you probably want to get rid of the ifs and just do 
```zig
std.os.access(foo, std.os.X_OK) catch return null;
```

Leon-Plickat:
Almost there.

Only thing left to do is the type. If I use a terminated string (`[*:0]const u8` or `[:0]const u8`) for the path (`startup_command`), then zig complains about freeing it, but if I use regular `[]const u8`, then it can't be integrated into `child_args` because it is not terminated.

river/main.zig
```diff
@@ -32,8 +32,42 @@ const usage: []const u8 =
...
+    if (try testConfigPath(allocator, "/etc/river/init", .{})) |path| {
+        return path;
+    }
+    return null;
+}
+
 pub fn main() anyerror!void {
-    var startup_command: ?[*:0]const u8 = null;
+    var allocator_instance = std.heap.GeneralPurposeAllocator(.{}){};
+    const gpa = &allocator_instance.allocator;
```
ifreund, line 69: really we should just use `util.gpa` here like the rest of the code base.

ifreund (commented):
I think `[:0]const u8` is the type you want here, though I don't see why zig would complain about freeing it. What's the error that you get?

Leon-Plickat:
I figured it out: Zig did not knew that `startup_command` could not be `null` at that point, so I had to explicitly tell it that via `free(startup_command orelse unreachable)`.

Leon-Plickat:
Ready for review.

river/main.zig
```diff
@@ -32,8 +32,36 @@ const usage: []const u8 =
...
+            return path;
+        }
+    } else if (std.os.getenv("HOME")) |home| {
+        if (try testConfigPath("{}/.config/river/init", .{home})) |path| {
+            return path;
+        }
+    }
+    if (try testConfigPath("/usr/local/etc/river/init", .{})) |path| {
+        return path;
+    }
```
ifreund, line 56: I don't see a reason we should look here. Do you have a strong reason why we should do so?
```suggestion
```

river/main.zig
```diff
@@ -82,6 +124,7 @@ pub fn main() anyerror!void {
             if (std.os.system.sigprocmask(std.os.SIG_SETMASK, &std.os.empty_sigset, null) < 0) unreachable;
             std.os.execveZ("/bin/sh", &child_args, std.c.environ) catch c._exit(1);
         }
+        util.gpa.free(startup_command orelse unreachable);
```
ifreund, line 127:
```suggestion
        util.gpa.free(cmd);
```

river/main.zig
Leon-Plickat: Convention. "Manually" installed things go to `/usr/local/`, which becomes relevant if we decide to ship a default config (per #146, f.e.).

ifreund: I think manually installing stuff system wide is an antipattern and people should use `-/.local/` for that use case. (I should probably amend the build command example in the readme I realize).

ifreund: The exact system dir is something we may want to make configurable at build time in the future though, I'm sure there's at least one distro out there that doesn't use `/etc/`

Leon-Plickat: Sounds reasonable. I'll remove the `/usr/local/` part for now.

ifreund (approved):
Thanks!

PR #155, opened 2020-11-19
code: Set wlroots loglevel based on -l flag

### LINKED ISSUE #154: Add an option to set wlroots log level
Sometimes during development it can be useful to set wlroots logging to `WLR_DEBUG` or any other log level.
In wlroots we have 4 log levels and in zig we have 7 so probably it would be good to add add change the `-l` argument to accept two values. Maybe in the format `river_log_level[:optional_wlroots_log_level]`.

The default would be to just log `WLR_ERROR`

Any opinions on this?

https://github.com/swaywm/wlroots/blob/6284af121fd1427eaefb043ea03363c969f2df3f/include/wlr/util/log.h#L25-L30

maringuu:
As discussed in #154

Closes #154

### DIFF
--- a/river/main.zig
+++ b/river/main.zig
@@ -93,7 +93,14 @@ pub fn main() anyerror!void {
         }
     }
 
-    c.wlr_log_init(.WLR_ERROR, null);
+    c.wlr_log_init(
+        switch (log.level) {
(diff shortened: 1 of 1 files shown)


ifreund (approved):
LGTM, Thanks!

PR #165, opened 2020-12-06
focus-follow-cursor: Change output focus when needed

maringuu:
I think this is the expected behavior. I added a note in the man page to make it clear.

Also I simplified the code.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -584,25 +584,15 @@ fn passthrough(self: *Self, time: u32) void {
 
             c.wlr_seat_pointer_notify_enter(self.seat.wlr_seat, wlr_surface, sx, sy);
             c.wlr_seat_pointer_notify_motion(self.seat.wlr_seat, time, sx, sy);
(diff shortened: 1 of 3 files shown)


river/Cursor.zig
```diff
@@ -584,25 +584,15 @@ fn passthrough(self: *Self, time: u32) void {
...
-                    .strict => {
-                        self.seat.focus(view);
-                        root.startTransaction();
-                    },
+
+            const follow_mode = config.focus_follows_cursor;
+            if (follow_mode == .strict or (follow_mode == .normal and focus_change)) {
+                if (View.fromWlrSurface(wlr_surface)) |view| {
+                    self.seat.focus(view);
+                    self.seat.focusOutput(view.output);
```
ifreund, line 592: Can you add an early return in this function if the passed output is already focused?

maringuu: Good catch.

ifreund (approved):
LGTM, Thanks!

PR #170, opened 2020-12-12
command: Implement spawn-tagmask

### LINKED ISSUE #169: Scratchpad
These are two things I heavily used in sway. It would be nice if there was a way to do them with river.

## Sticky views
~A sticky view is a view that is always visible. Commonly sticky views are floating.~
~My idea was to give sticky views a specific tag.
Lets say I have tags with values `0b001`and `0b010`. To implement sticky views I introduced a tag `0b100`. All sticky views would go to this tag.
My old config~
```
# The 0b is just for better explanation
riverctl map $mod 1 set-focused-tags 0b001
riverctl map $mod 2 set-focused-tags 0b010
```
~would then become~
```
riverctl map $mod 1 set-focused-tags 0b101
riverctl map $mod 2 set-focused-tags 0b110

# Make view sticky
riverctl map $mod S set-view-tags 0b100
```

~The problem is that all applications that are spawned inherit the current view tags and thus would also be sticky~

Edit: I thought the wrong way around. This is already possible.

## Ranger as an application opener
When I used sway I always opened a terminal with `ranger` in it. This terminal would go to the scratchpad.
When I wanted to open a pdf on my _current_ workspace I showed the scratchpad, opened the pdf and then hid my terminal with ranger.

When implementing a simelar behavior in river I face the same problem as with sticky views. Newly spawned windows would always be part of the scratchpad.

# Proposed solution
Both usecases are solved by being able to set the tagmasks that new views will have.
Something like `riverctl set-spawn-tags` and `riverctl toggle-spawn-tags`.

What are your thoughts?

maringuu:
I find it quite useful to be able to filter out some tags when an new view is opened.
My primary usecase is to have scratchpad like behaviour.

Closes #169

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -104,6 +104,8 @@ take a normal base 10 number as their argument but the semantics are best
 understood in binary. The binary number 000000001 represents a set containing
 only tag 1 while 100001101 represents a set containing tags 1, 3, 4, and 9.
 
+When a view spawns it will be assigned the currently focused tags.
+
 At least one tag must always be focused and each view must be assigned at
 least one tag. Operations that would violate either of these requirements
 are ignored by river.
@@ -115,6 +117,13 @@ are ignored by river.
 *set-view-tags* _tags_
 	Assign the currently focused view the tags corresponding to the set
 	bits of _tags_.
+
+*spawn-tagmask* _tagmask_
+	Specify a _tagmask_ to filter the tags assigned to a newly spawned view.
+	E.g. when 000011111 is focused and _tagmask_ is 111110000 a view will
+	be assigned 000010000.
+	When no tags would remain after filtering the _tagmask_ is ignored.
+
 
 *toggle-focused-tags* _tags_
 	Toggle visibility of views with tags corresponding to the set bits
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -73,6 +73,9 @@ layout: []const u8,
 /// Determines where new views will be attached to the view stack.
 attach_mode: AttachMode = .top,
 
+/// Bitmask that whitelists tags for newly spawned views
+spawn_tagmask: u32 = (1 << 32) - 1,
+
 /// List of status tracking objects relaying changes to this output to clients.
 status_trackers: std.SinglyLinkedList(OutputStatus) = .{},
 
--- a/river/Server.zig
+++ b/river/Server.zig
@@ -177,7 +177,7 @@ fn handleNewXdgSurface(listener: *wl.Listener(*wlr.XdgSurface), xdg_surface: *wl
         xdg_surface.resource.postNoMemory();
         return;
     };
-    node.view.init(output, output.current.tags, xdg_surface);
+    node.view.init(output, getNewViewTags(output), xdg_surface);
 }
 
 /// This event is raised when the layer_shell recieves a new surface from a client.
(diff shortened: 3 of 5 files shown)


ifreund:
Hmm, the use case for scratchpad like is definitely valid but I'm not sure if this is the way we want to support it. I think adding a `-tags` flag to spawn might be more intuitive as it would avoid adding more global state.

maringuu:
I'm not sure if this is the right way to do this. It seems counter intuitive for me that with `-tags` it would be possible to specify tags for a command that may not be anything graphical at all.
Also then it would not be possible to specify tags for windows that are not opened through `spawn`.
(Now that I write it down I realise that I `set-spawn-tagmask` is bad wording from me. The man page descibes what I intended it to do.)

Leon-Plickat:
I feel like this is a needed feature, but that this implementation is a bit too limited. Maybe the right way to do it would be another filter list, where the user can configure arbitrary tags for views based on their app-id.

Leon-Plickat:
IIUC, the idea of this patch, on a technical level, is to make arbitrary tags unavailable to clients when they are mapped. Wouldn't it then make more sense to have a black-list instead of a white-list? Instead of a tagmask of allowed tags, I think it would be more idiomatic to  set a tagmask of denied tags.

ifreund:
> I'm not sure if this is the right way to do this. It seems counter intuitive for me that with `-tags` it would be possible to specify tags for a command that may not be anything graphical at all.

I agree, there is no reason to run non-graphical programs through `riverctl spawn`. In fact, the only point of having a spawn command in the first place is to be able to implement behavior such as the proposed `-tags` flag or handle the case of the user switching tags right after spawning something but before the spawned program maps a toplevel.

> Also then it would not be possible to specify tags for windows that are not opened through `spawn`.

Good point, I suppose we would need a global tagmask anyways to address this. I agree with @Leon-Plickat that a blacklist would make more sense for this use-case though, it could function through a simple subtraction.

maringuu:
> Wouldn't it then make more sense to have a black-list instead of a white-list?

Yes!

> Maybe the right way to do it would be another filter list

Good idea, but as you said that is another usecase.

> I agree, there is no reason to run non-graphical programs through riverctl spawn

I'm confused. Some non-gui things I spawn with `riverctl map [..] spawn` are `mako`, `grim` and `pamixer`.
My point was that if we had something like `riverctl spawn -tags $tags` you could have `riverctl map $mod F12 spawn -tags grim` in your config which does not make sense.


I changed the PR to implement `set-view-tag-blacklist` (I'm not that happy with naming...) to implement the blacklist behavior.

Leon-Plickat:
> (I'm not that happy with naming...) 

What about `set-view-tag-block-mask`? I think "mask" fits better than "list" since tags are a bitfield.

ifreund:
Hmm, naming stuff is really hard. I think "mask" is more accurate than list here, but "mask" always means a whitelist not a blacklist in the context of bitfields. Perhaps your original approach would indeed be the most intuitive way to do this. I'd prefer to leave off the "set" prefix though in cases like this where there is no ambiguity. `spawn-tags-mask` is my current favorite.  As an aside, I think we should stop using the term "tagmask" in our documentation and just say "tags" or "tag bitfield" as it's a little misleading.

maringuu:
Thanks for your help.

>  As an aside, I think we should stop using the term "tagmask" in our documentation and just say "tags" or "tag bitfield" as it's a little misleading.

That sounds reasonable. I prefer `tags`.

>  Perhaps your original approach would indeed be the most intuitive way to do this.

Probably. `spawn-tag-allowed-mask` would also work imo.

ifreund:
I've updated the docs in 5f4ba06 to clarify the terminology surrounding tags. Unfortunately this seems to have caused some conflicts. `spawn-tag-allowed-mask` isn't a bad option but it's a little verbose and `allowed` is already implied by `mask` IMO.

Could you rebase this, switch it back to being a whitelist, and name the command either `spawn-tag-mask` or `spawn-tagmask`? I should finally be able to merge it then.

maringuu:
Finally had time to finish this.

* Rebased
* Renamed to `spawn-tagmask`

On thing I noted is that as spawn-tagmasks are per output simply putting `riverctl spawn-tagmask $tagmaskForEveryOutput` will of course just be applied to the output that is focused on startup. When we have hooks or something this can be improved (just like setting the layout for every output)

ifreund:
Made some opinionated changes to the documentation and rebased to master locally. Looks like Github is too dumb to realize this though :/

Thanks for the contribution!

PR #176, opened 2020-12-23
river-control: rewrite as v2

ifreund:
Instead of passing the raw argv to the compositor for in-compositor
parsing encode all valid commands as requests in the protocol. This
moves a significant amount of error handling and input validation
out of the compositor process.

Draft PR in case anyone wants to read/comment on the protocol while I work on the implementation

### DIFF
--- a/protocol/river-control-unstable-v1.xml
+++ b/protocol/river-control-unstable-v1.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<protocol name="river_control_unstable_v1">
+<protocol name="river_control_unstable_v2">
   <copyright>
     Copyright 2020 The River Developers
 
@@ -16,17 +16,9 @@
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
   </copyright>
 
-  <interface name="zriver_control_v1" version="1">
+  <interface name="zriver_control_v2" version="1">
     <description summary="run compositor commands">
-      This interface allows clients to run compositor commands and receive a
-      success/failure response with output or a failure message respectively.
-
-      Each command is built up in a series of add_argument requests and
-      executed with a run_command request. The first argument is the command
-      to be run.
-
-      A complete list of commands should be made available in the man page of
-      the compositor.
+      This interface allows clients to control compositor behavior.
     </description>
 
     <request name="destroy" type="destructor">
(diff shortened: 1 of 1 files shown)


protocol/river-control-unstable-v1.xml
```diff
@@ -37,49 +29,421 @@
...
+        null, instead remove the mapping if it exists.
+
+        A complete list of XKB key names for the key argument can usually be
+        found at /usr/include/xkbcommon/xkbcommon-keysyms.h.
+      </description>
+      <arg name="mode" type="string"/>
+      <arg name="modifiers" type="uint" enum="modifiers"/>
+      <arg name="key" type="string" summary="XKB key name"/>
+      <arg name="state" type="uint" enum="wl_keyboard.key_state"/>
+      <arg name="command" type="string" allow-null="true"/>
```
Leon-Plickat, line 303: Just from reading the protocol it is not clear what `command` is supposed to be. Most people would probably interpret it as a shell command. Also it should be documented somewhere what strings make river do what.

protocol/river-control-unstable-v1.xml
```diff
@@ -37,49 +29,421 @@
...
+      </description>
+      <arg name="app_id" type="string" summary="the app_id to filter"/>
+    </request>
+
+    <request name="background_color">
+      <description summary="set the background color"/>
+      <arg name="red" type="fixed"/>
+      <arg name="green" type="fixed"/>
+      <arg name="blue" type="fixed"/>
+      <arg name="alpha" type="fixed"/>
```
Leon-Plickat, line 229: I don't think alpha makes much sense for the background colour.

protocol/river-control-unstable-v1.xml
```diff
@@ -37,49 +29,421 @@
...
+    <request name="layout">
+      <description summary="set the layout command">
+        Provide a command which river will use for generating the layout
+        of non-floating windows on the currently focused output. See
+        *river-layouts*(7) for details on the expected formatting of the
+        output of layout commands. Alternatively, "full" can be given
+        instead of a command to cause river to use its single internal layout,
+        in which windows span the entire width and height of the output.
+      </description>
+      <arg name="seat" type="object" interface="wl_seat"/>
```
Leon-Plickat, line 191: Instead of getting the focused output of the seat, maybe this request should just have an `output` parameter. A client can get the focused output with `river-status-v1`.

Leon-Plickat (commented):
Overall this seems fine.

Although if we want river to be really scriptable then in the long run we may want to think about a protocol that allows operations on arbitrary views, not just the focused one.

protocol/river-control-unstable-v1.xml
ifreund: It's trivial to support and has an effect in a nested context so why not continue supporting it (we already do).

protocol/river-control-unstable-v1.xml
ifreund: It is just a shell command, though I realize I should probably have documented that better. In order to remove all parsing code from the compositor, my current plan is to have riverctl pass e.g. `riverctl focus-view next` as the command argument to map when invoked with `riverctl map normal Mod4 J focus-next`.

protocol/river-control-unstable-v1.xml
ifreund: > Although if we want river to be really scriptable then in the long run we may want to think about a protocol that allows operations on arbitrary views, not just the focused one.

You make a good point, this current draft is still very close to a direct translation of the previous command interface. We do have an opportunity to make this a lot more flexible while still exposing the same command line interface through riverctl. I'll have to give some more thought towards how to best handle addressing of views.

Leon-Plickat:
> [...] I'll have to give some more thought towards how to best handle addressing of views.

RE: our discussion on IRC a few days ago about making river-control an extension to foreign-toplevel.

While that definitely makes this protocol considerably more powerful, this approach is probably not good for what this protocol will be used most:  Manipulating the focused view.

Your idea is to stop parsing binds in river and always execute `riverctl`. That would mean that for a lot of actions, `riverctl` would be getting the toplevel list and finding the focused one. That's a lot of wasted cycles IMO.

Maybe we need some way to very simply get the foreign handle of only the currently focused view.

> my current plan is to have riverctl pass e.g. riverctl focus-view next as the command argument to map when invoked with riverctl map normal Mod4 J focus-next.

I also have a feeling this will not work when multi-seat support is involved. How is `riverctl` supposed to know what seat issued the command?

ifreund:
> Maybe we need some way to very simply get the foreign handle of only the currently focused view.

Alternatively, we could make passing NULL as the target toplevel handle mean that the currently focused toplevel should be used.

ifreund:
```
19:05 <leon-p> well, we really would just need the next river-control to be able to change the output and tags of arbitrary views /atomically/.
19:05 <leon-p> Maybe hold all changes until the client sends "done"
19:05 <leon-p> or why not "riverToplevel.commit" or something like that
19:06 <ifreund> hmm, a commit request for river control would be a good idea yeah
19:07 <leon-p> would require Views to have a third state though which is then copied to pending on commit. But I doubt thats a problem
19:08 <ifreund> either that or a way to represent commands such that they could be queued up
19:09 <ifreund> which would also get rid of that roundtrip issue with my current v2 approach for river-control
19:09 <leon-p> yeah a command queue would probably work better for changing outputs
19:09 <ifreund> it would be much more robust yeah
19:10 <leon-p> I think riverctl should by default then commit directly after adding the command to the queue, but have an option to omit the committing.
19:11 <ifreund> that seems like a reasonable way to expose such functionality with the CLI, though getting the protocol needs to come first
19:12 <ifreund> s/protocol/protocol done/
```

YerinAlexey (approved):

PR #180, opened 2021-01-01
build: fix #179 - add `default_config_path` build option

### LINKED ISSUE #179: default init script doesn't run when installing to a different prefix
I built river and installed to ~/.local. When I ran it, I noticed none of the keybinds were set. Looks like river looks for `init` in `XDG_CONFIG_HOME`, `HOME`, then `/etc`. Maybe the install prefix should be added to the places to look, or it should replace the `/etc` fallback. I'd be happy to have a go at implementing this and submit a PR.

### Steps

1. `zig build install --prefix=$HOME/.local` or anywhere else
2. run `~/.local/bin/river`
3. try using any keybind

AlexJMohr:
Instead of hardcoding `/etc/river/init`, the config file will now be installed to `{b.install_prefix}/etc/river/init`. The path is passed to the river executable via `river.addBuildOption()`

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -27,13 +27,16 @@ pub fn build(b: *zbs.Builder) !void {
 
     const examples = b.option(bool, "examples", "Set to true to build examples") orelse false;
 
-    // Sigh, why are the conventions inconsistent like this.
     const resolved_prefix = try std.fs.path.resolve(b.allocator, &[_][]const u8{b.install_prefix.?});
-    if (std.mem.eql(u8, resolved_prefix, "/usr")) {
-        b.installFile("example/init", "../etc/river/init");
-    } else {
-        b.installFile("example/init", "etc/river/init");
-    }
+    const rel_config_path = if (std.mem.eql(u8, resolved_prefix, "/usr"))
+        "../etc/river/init"
+    else
+        "etc/river/init";
+    b.installFile("example/init", rel_config_path);
+    const default_config_path = try std.fs.path.resolve(
+        b.allocator,
+        &[_][]const u8{ resolved_prefix, rel_config_path },
+    );
 
     const scanner = ScanProtocolsStep.create(b);
     scanner.addSystemProtocol("stable/xdg-shell/xdg-shell.xml");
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -27,13 +27,11 @@ pub fn build(b: *zbs.Builder) !void {
...
-    // Sigh, why are the conventions inconsistent like this.
-    const resolved_prefix = try std.fs.path.resolve(b.allocator, &[_][]const u8{b.install_prefix.?});
-    if (std.mem.eql(u8, resolved_prefix, "/usr")) {
-        b.installFile("example/init", "../etc/river/init");
-    } else {
-        b.installFile("example/init", "etc/river/init");
-    }
+    const default_config_path = try std.fs.path.resolve(
+        b.allocator,
+        &[_][]const u8{ b.install_prefix.?, "etc/river/init" },
```
ifreund, line 32: We still need the logic to install to `/etc/river/init` if the prefix is `/usr`.

ifreund (changes requested):
Looks good, just one thing to fix.

Also could you use something a little more descriptive than "Fix #179" as the first line of the commit? "build: derive default config path from install prefix" would better explain what this patch does without needing to check github.

build.zig
AlexJMohr: Changed the commit message. What is the purpose of moving out of `/usr` if river can find it there via `default_config_path`?

build.zig
```diff
@@ -27,13 +27,15 @@ pub fn build(b: *zbs.Builder) !void {
...
-        b.installFile("example/init", "etc/river/init");
-    }
+    const default_config_path = try std.fs.path.resolve(
+        b.allocator,
+        &[_][]const u8{
+            if (std.mem.eql(u8, resolved_prefix, "/usr")) "/" else resolved_prefix,
+            "etc/river/init",
+        },
+    );
+    b.installFile("example/init", default_config_path);
```
ifreund, line 38: According to the doc comment on `Builder.installFile()` the second argument should be a path relative to the install prefix. This currently if `Builder.dest_dir` is `null` but will break if `DESTDIR` is set in the environment. This is important for packaging so we should really use a relative path here.

ifreund: https://github.com/ziglang/zig/pull/7646

ifreund (approved):
LGTM, Thanks!

PR #208, opened 2021-01-21
Xwayland: Floating window size

pmkap:
Use the X size hints max- and min width/height to determine the size of a floating window. First time I use Zig so I am not sure if this could be done nicer.


ifreund (commented):
I don't think using the maximum size as the default make sense. I've just pushed a related commit to master (5e09b853f70b4c78186bcd2d5865843ed6c0db3d), could you try that and see if it solves whatever issue you were having?

pmkap:
Indeed, this solves the issues I had with some popup windows. Thanks!

ifreund:
> Indeed, this solves the issues I had with some popup windows. Thanks!

No problem, thanks for your patience!

PR #209, opened 2021-02-01
Add output-title default option

Leon-Plickat:
Outputs now have a default option, "output-title". If this changes, the
outputs title is set to the option value.

---

Also lays the ground-work for any later default options.

### DIFF
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -35,6 +35,7 @@ const View = @import("View.zig");
 const ViewStack = @import("view_stack.zig").ViewStack;
 const AttachMode = @import("view_stack.zig").AttachMode;
 const OutputStatus = @import("OutputStatus.zig");
+const Option = @import("Option.zig");
 
 const State = struct {
     /// A bit field of focused tags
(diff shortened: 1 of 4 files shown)


river/Output.zig
```diff
@@ -545,3 +552,29 @@ pub fn getEffectiveResolution(self: *Self) struct { width: u32, height: u32 } {
         .height = @intCast(u32, height),
     };
 }
+
+pub fn setTitle(self: *Self, title: [*:0]const u8) void {
+    if (self.wlr_output.isWl()) {
+        self.wlr_output.wlSetTitle(title);
+    } else if (self.wlr_output.isWl()) {
```
ifreund, line 559:
```suggestion
    } else if (self.wlr_output.isX11()) {
```

ifreund (commented):
just a bit of bikeshedding, I tested in river nested in sway nested in river and it worked. I wonder if we should append the output name to the value of the option when setting the title though as this is useful information.

ifreund:
I added the output name to the default value, renamed the option from `output-title` to `output_title`, and landed this in 7029a5cd3e90ed5c23e8e5039745f92caeca4a99.

I think using snake_case over kebab-case is the right convention to set for option names in order to distinguish them from commands.

PR #212, opened 2021-02-05
Use std.log instead of log.zig

foxfromabyss:
Replaces log.zig with std.log.

I am not sure if the way that I did it makes sense, so just let me know if top level scope declaration, where possible, is actually wrong.


foxfromabyss:
I also couldn't come up with a better commit message sadly :(

river/View.zig
```diff
@@ -441,13 +442,13 @@ pub fn map(self: *Self) void {
 
     self.pending.target_opacity = self.output.root.server.config.opacity.unfocused;
 
-    log.debug(.server, "view '{}' mapped", .{self.getTitle()});
+    std.log.scoped(.server).debug("view '{}' mapped", .{self.getTitle()});
```
ifreund, line 445: let's just put this messages and the others in this file that currently have the server scope in the view scope

river/main.zig
```diff
@@ -21,11 +21,13 @@ const wlr = @import("wlroots");
 const build_options = @import("build_options");
 
 const c = @import("c.zig");
-const log = @import("log.zig");
 const util = @import("util.zig");
 
 const Server = @import("Server.zig");
 
+const log = std.log.scoped(.server);
+var log_level: std.log.Level = .warn;
```
ifreund, line 29: Making this `var` doesn't affect whether or not the log level is runtime configurable. What we want to do here is set this variable to `.debug` and define a custom log function in this file which preforms the filtering by log level at runtime. The log function in river's current  `log.zig` should be suitable.

ifreund (changes requested):
Thanks for working on this, don't forget that we can delete river's current `log.zig` with these changes!

river/main.zig
foxfromabyss: Thanks for the feedback!
Could you elaborate on that?
I can't seem to find the reference point for `log.zig`, to see the exported functions or anything, sadly.
What do I search to find the reference?
New to Zig, and all that

ifreund: The `std.log` implementation will check for a log function defined in the root source file (`river/main.zig` in this case) and call that instead of its default implementation ([code here](https://github.com/ziglang/zig/blob/783cb980ab137d6a99c20f007d70b4191d4a63c4/lib/std/log.zig#L126)). 

We need to define such a function to make the log level configurable at runtime by passing the `-l` flag to river. Copying the log function from river's current `log.zig` to `main.zig` and making it `pub` should be sufficient.

foxfromabyss: Took some time to figure it out, thank you very much for the explanation

ifreund (approved):
LGTM, Thanks!

PR #221, opened 2021-02-14
Implement pointer_gestures_v1

maringuu:
Depends on https://github.com/swaywm/zig-wlroots/pull/10

Pinching can be tested with evince and swiping can be tested with the GNOME Web.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -40,6 +40,7 @@ pub fn build(b: *zbs.Builder) !void {
 
     const scanner = ScanProtocolsStep.create(b);
     scanner.addSystemProtocol("stable/xdg-shell/xdg-shell.xml");
+    scanner.addSystemProtocol("unstable/pointer-gestures/pointer-gestures-unstable-v1.xml");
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -41,6 +41,7 @@ pub fn build(b: *zbs.Builder) !void {
     const scanner = ScanProtocolsStep.create(b);
     scanner.addSystemProtocol("stable/xdg-shell/xdg-shell.xml");
     scanner.addSystemProtocol("unstable/xdg-output/xdg-output-unstable-v1.xml");
+    scanner.addProtocolPath("protocol/pointer-gestures-unstable-v1.xml");
```
ifreund, line 44: This protocol should be available from the system wayland-protocols package
```suggestion
    scanner.addSystemProtocol("unstable/pointer-gestures-unstable-v1.xml");
```

ifreund (approved):
LGTM, Thanks!

PR #223, opened 2021-02-17
Add distance threshold for move/resize request of tiled XDG toplevels

Leon-Plickat:
Back when I implemented move/resize requests for XDG toplevel I also accidentally introduced a now pretty common and annoying issue: Views immediately enter the move/resize mode and were therefore are set to floating when the request is made, which would interfere with normal usage (example: clicking anywhere in the mpv window and it will immediately start floating).

This patch  introduces an arbitrary threshold before the requested move/resize mode is entered. This is only the case for tiled views; Floating views will enter the mode immediately. The issue mentioned above is solved by this addition.

This is indeed yet another topic where a "client-side vs server-side" discussion could ignite, but the current state is that there is only one client-side implementation of such a threshold I am aware of (GTK header bars). All other clients leave it up to the server. This means GTK header bars now have two thresholds, the CS and SS one, but since both are small and we only apply the SS one for tiled views it is not really noticeable, so I am personally fine with it. We could also make our SS threshold smaller if need be.

FWIW kwin apparently also has a threshold for "snapped" views, although theirs appears to be much smaller.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -46,9 +46,25 @@ const Mode = union(enum) {
         offset_x: i32,
         offset_y: i32,
     },
+    moveThreshold: struct {
+        view: *View,
+        start_x: f64,
+        start_y: f64,
+        current_x: f64,
+        current_y: f64,
+    },
+    resizeThreshold: struct {
+        view: *View,
+        start_x: f64,
+        start_y: f64,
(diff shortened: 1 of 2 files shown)


YerinAlexey (approved):

Leon-Plickat:
Superseded by #224

PR #233, opened 2021-03-05
Implement idle-inhibit-unstable-v1

### LINKED ISSUE #199: Implement idle-inhibit-unstable-v1
Waybar keeps crashing when clicking the idle inhibitor module. On sway it seems to work so that leads me to believe it is a problem with river not implementing the [idle inhibit protocol](https://github.com/wayland-project/wayland-protocols/blob/master/unstable/idle-inhibit/idle-inhibit-unstable-v1.xml).

I'm fine if you decide to not support the protocol, but could there be a little note about it in the wiki, especially since waybar is a recommended software?

maringuu:
Depends on https://github.com/swaywm/zig-wlroots/pull/12

Closes #199

### DIFF
--- a/river/Server.zig
+++ b/river/Server.zig
@@ -125,6 +125,7 @@ pub fn init(self: *Self) !void {
     _ = try wlr.DataControlManagerV1.create(self.wl_server);
     _ = try wlr.ExportDmabufManagerV1.create(self.wl_server);
     _ = try wlr.GammaControlManagerV1.create(self.wl_server);
+    _ = try wlr.IdleInhibitManagerV1.create(self.wl_server);
     _ = try wlr.ScreencopyManagerV1.create(self.wl_server);
     _ = try wlr.Viewporter.create(self.wl_server);
 }


river/InputManager.zig
```diff
@@ -37,6 +37,7 @@ server: *Server,
 new_input: wl.Listener(*wlr.InputDevice) = wl.Listener(*wlr.InputDevice).init(handleNewInput),
 
 idle: *wlr.Idle,
+idle_inhibit_manager: *wlr.IdleInhibitManagerV1,
```
ifreund, line 40: There's no need to store a pointer to this as it's never used and will free itself when the `wl.Server` is destroyed.

maringuu: Alright. I moved the creation to `Server.zig` as I just saw that all the other wlroots interfaces that free themselves sit there.

ifreund (commented):
Just took a minute to read to the protocol and what wlroots provides. Just creating the `IdleInhibitManager` isn't really enough to support this protocol. Wlroots only manages creation/destruction of the protocol objects for us and does not implement any of the actual idle inhibition semantics.

From the protocol the semantics are:
```
      An idle inhibitor prevents the output that the associated surface is
      visible on from being set to a state where it is not visually usable due
      to lack of user interaction (e.g. blanked, dimmed, locked, set to power
      save, etc.)  Any screensaver processes are also blocked from displaying.

      If the surface is destroyed, unmapped, becomes occluded, loses
      visibility, or otherwise becomes not visually relevant for the user, the
      idle inhibitor will not be honored by the compositor; if the surface
      subsequently regains visibility the inhibitor takes effect once again.
      Likewise, the inhibitor isn't honored if the system was already idled at
      the time the inhibitor was established, although if the system later
      de-idles and re-idles the inhibitor will take effect.
```

maringuu:
Whoops. I failed to read that.

maringuu:
Sadly I don't know when I have time to implement this properly.

PR #236, opened 2021-03-13
completions: add zsh completion for riverctl

novakne:
I think I didn't forget any command, it works for all commands and some sub-commands for when there is a choice like `next previous`

### DIFF
--- a/completions/zsh/_riverctl
+++ b/completions/zsh/_riverctl
@@ -1,0 +1,99 @@
+#compdef riverctl
+
+
+_riverctl() {
+
+  _next_prev() { _alternative 'arguments:args:(next previous)' }
+  _orientations() { _alternative 'arguments:args:(up down left right)' }
+  _hor_ver() { _alternative 'arguments:args:(horizontal vertcial)' }
+  _attach() { _alternative 'arguments:args:(top bottom)' }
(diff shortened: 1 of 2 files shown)


completions/zsh/_riverctl
```diff
@@ -0,0 +1,134 @@
...
+      |background-color\
+      |border-color-focused\
+      |border-color-unfocused\
+      |border-width\
+      |opacity\
+      |outer-padding\
+      |set-repeat\
+      |view-padding\
+      |xcursor-theme\
+      |outputs)
```
YerinAlexey, line 106: I think this could be simplified with `_cmds`

novakne: Oh yeah it works, didn't think about it, just
``` bash
  case "$words[1]" in
    _cmds) return 0 ;;
```

YerinAlexey: > ```shell
>   case "$words[1]" in
>     _cmds) return 0 ;;
> ```

Wait, that's illegal

I would expect something like this:

```shell
case "$words[1]" in
  # snip
  *)
    return 0
  ;;
esac
```

novakne: Yeah it's better

YerinAlexey (approved):
LGTM!

ifreund (approved):
Thanks!

PR #237, opened 2021-03-13
Add bash completion for riverctl

Leon-Plickat:


YerinAlexey (approved):

build.zig
```diff
@@ -25,6 +25,18 @@ pub fn build(b: *zbs.Builder) !void {
...
+        bool,
+        "bash-completion",
+        "Set to true to install bash completion for riverctl. Defaults to true if bash is installed.",
+    ) orelse bash_found: {
+        _ = b.findProgram(&[_][]const u8{"bash"}, &[_][]const u8{}) catch |err| switch (err) {
+            error.FileNotFound => break :bash_found false,
+            else => return err,
+        };
+        break :bash_found true;
+    };
```
ifreund, line 38: I don't think this logic is worth it, we don't require bash as a dependency to generate or install the completions.
```suggestion
    ) orelse true;
```

Leon-Plickat: Makes sense.

ifreund (approved):
LGTM, Thanks!

PR #246, opened 2021-03-26
Fix #153 (screen freezing issues)

### LINKED ISSUE #153: wlr_output_commit sometimes start to fail a lot
Tested on [current master](cbd4a2807bcfa74cc87c36d7ec20cdf92a7c77fc) the rendering "freezes".

My setup consists of two outputs. One 1080p laptop display and an external 4k output connected via displayport.

I did the following:
1. Open river
2. Spawn a bunch of terminals (foot)
3. Change the layout a couple of times to cause calls to `root.startTransaction`
4. Now changing the layout did not result in a visual change
5. Change focus to my other output and spawn a terminal. I didn't see any visual change here either
6. Change VT back and fourth
7. Now the rendering works as expected

I can kind of reliably reproduce this. The steps above lead to this most of the time.

Here are the logs: https://paste.sr.ht/~maringuu/bb42efa93eeed27b3d8d53258c907ee9f2771f87

Here are some other logs where the rendering also froze: 
https://paste.sr.ht/~maringuu/862ccc42f315baeeadffcd20fe667ad604027689

Edit:
I figured out a simpler way to reproduce. It works very reliable for me.
1. Open river
2. Open at least two terminals (foot)
3. Do `riverctl swap next` a bunch of times until it freezes

Q-I-U:
This should fix rendering "freezes" reported in #153
Tested on FreeBSD 14-current

### DIFF
--- a/river/render.zig
+++ b/river/render.zig
@@ -143,8 +143,12 @@ pub fn renderOutput(output: *Output) void {
     }
 
     // TODO: handle failure
-    output.wlr_output.commit() catch
+    if (output.wlr_output.commit()) |_| {
+        return;
+    } else |_| {
+        renderOutput(output);
         log.err("output commit failed for {}", .{output.wlr_output.name});
+    }
 }
 
 fn renderFilter(view: *View, filter_tags: u32) bool {
--- a/river/render.zig
+++ b/river/render.zig
@@ -143,7 +143,7 @@ pub fn renderOutput(output: *Output) void {
     }
 
     output.wlr_output.commit() catch {
-        log.err("output commit failed for {}, retrying", .{output.wlr_output.name});
+        log.err("output commit failed for {}, retrying", .{std.mem.spanZ(&output.wlr_output.name)});
         output.wlr_output.scheduleFrame();
     };
 }
--- a/river/render.zig
+++ b/river/render.zig
@@ -142,7 +142,6 @@ pub fn renderOutput(output: *Output) void {
         output.wlr_output.setDamage(&damage);
     }
 
-    // TODO: handle failure
     output.wlr_output.commit() catch {
         log.err("output commit failed for {}, retrying", .{output.wlr_output.name});
         output.wlr_output.scheduleFrame();
--- a/river/render.zig
+++ b/river/render.zig
@@ -143,12 +143,10 @@ pub fn renderOutput(output: *Output) void {
     }
 
     // TODO: handle failure
-    if (output.wlr_output.commit()) |_| {
-        return;
-    } else |_| {
-        renderOutput(output);
-        log.err("output commit failed for {}", .{output.wlr_output.name});
-    }
+    output.wlr_output.commit() catch {
+        log.err("output commit failed for {}, retrying", .{output.wlr_output.name});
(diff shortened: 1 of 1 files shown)


river/render.zig
```diff
@@ -143,8 +143,12 @@ pub fn renderOutput(output: *Output) void {
     }
 
     // TODO: handle failure
-    output.wlr_output.commit() catch
+    if (output.wlr_output.commit()) |_| {
+        return;
+    } else |_| {
+        renderOutput(output);
         log.err("output commit failed for {}", .{output.wlr_output.name});
+    }
```
ifreund, line 151:
```suggestion
    output.wlr_output.commit() catch {
        log.err("output commit failed for {}, retrying", .{output.wlr_output.name});
        renderOutput(output);
    };
```

ifreund (commented):
Won't this cause river to enter an infinite loop if committing the output always fails?

Why does this work in the first place, did you figure you figure out the cause of #153?

Q-I-U:
I also thought that setting a maximum amount of retries would be a good idea, but the number of needed retries is pretty random ranging form a few retries to sometimes over 100.
When I was using river like this for a few hours it didn't get stuck in a loop, so I guess that It always
renders the output, sometimes it just takes a few tries.

For me the freezes only occurred when the screen was getting redrawn very fast ie. when playing a game

Leon-Plickat:
So... why would this work? If `renderOutput()` fails, river would try again on the next `output.frame` event anyway.

Q-I-U:
I see, I will investigate this further in that case. I will update the pull request if I find a better approach.
For me to "un-stuck" river (without this patch) I had to switch to a tty and then back to river. 
So I guess that for some reason `output.frame` is only triggered only when switching ttys after `renderOutput()` fails

Q-I-U:
Ok, so it looks like when `wlr_output.commit()` fails it runs `output_state_clear()` which
prevents the frame event from being triggered later on. I changed the patch so it uses `scheduleFrame` 
which AFAIK directly emits the `events.frame` event. This should solve the infinite loop problem the 
previous version had.

I think this is the most direct way to solve the problem.

river/render.zig
```diff
@@ -143,8 +143,10 @@ pub fn renderOutput(output: *Output) void {
     }
 
     // TODO: handle failure
```
ifreund, line 145:
```suggestion
```

river/render.zig
```diff
@@ -143,8 +143,10 @@ pub fn renderOutput(output: *Output) void {
...
-    output.wlr_output.commit() catch
-        log.err("output commit failed for {}", .{output.wlr_output.name});
+    output.wlr_output.commit() catch {
+        log.err("output commit failed for {}, retrying", .{output.wlr_output.name});
```
ifreund, line 147: Lets fix this to stop at the null terminator instead of printing the full array since we're already touching this line.
```suggestion
        log.err("output commit failed for {}, retrying", .{std.mem.spanZ(&output.wlr_output.name)});
```

ifreund (commented):
Scheduling a new frame looks like it does what what we want here, good find.

ifreund (approved):
LGTM, Thanks!

PR #248, opened 2021-03-27
Layer shell fix

### LINKED ISSUE #245: Layershell surfaces can cause the compositor to crash
Surfaces that are too big or outside the display can cause river to crash. I suspect it might be due to the arbitrary geometry limit river imposes on the clients (you can't move a window beyond the screen). 

This can also be tied with this [issue](https://github.com/ifreund/river/issues/226) regarding mako. Basically river's implementation of layershell has issues.

Leon-Plickat:
Should fix #245

**TODO:** What to do when the margins are larger than the bounds? **A:** Just close the surface for now.

### DIFF
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -400,7 +400,25 @@ fn arrangeLayer(
         var new_box: Box = undefined;
 
         // Horizontal alignment
-        if (current_state.desired_width == 0) {
+        const horizontal_margin_size = current_state.margin.left + current_state.margin.right;
+        if (horizontal_margin_size >= bounds.width) {
+            // TODO find a better solution
+            // We currently have not reached a conclusion on how to gracefully
+            // handle this case yet, so we just close the surface. That will
+            // cause the output to be re-arranged eventually, so we can just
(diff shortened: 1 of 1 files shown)


river/Output.zig
```diff
@@ -400,7 +400,24 @@ fn arrangeLayer(
...
         // Horizontal alignment
-        if (current_state.desired_width == 0) {
+        const horizontal_margin_size = current_state.margin.left + current_state.margin.right;
+        if (horizontal_margin_size >= bounds.width) {
+            // TODO find a better solution
+            // We currently have not reached a conclusion on how to gracefully
+            // handle this case yet, so we just close the surface. That will
+            // cause the output to be re-arranged eventually, so we can just
+            // exit here. Technically doing this is incorrect, but this case
+            // should only ever be encountered very rarily, so whatever.
```
ifreund, line 410:
```suggestion
            // should only ever be encountered very rarely and matches the behavior of other compositors.
```

ifreund (commented):
Logic looks good to me.

ifreund (commented):
Thanks!

PR #264, opened 2021-04-14
Added installation through nix

branwright1:
Hello, since river is now part of nixpkgs repo which is official channel for NixOS and non-nixos users, just added little update to readme on how to install prebuild package from nix. River from master branch was merged today and now it's available as binary cache according to [this](https://github.com/NixOS/nixpkgs/pull/119387) pull request. I'm having acces to this derivation so river will be always even with master branch (excluding small edits like fix typos etc.).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,14 @@ zig build -Drelease-safe --prefix /usr install
 ```
 To enable experimental Xwayland support pass the `-Dxwayland` option as well.
 
+## Install from package manager
+Currently river is available in [nixpkgs](https://github.com/NixOS/nixpkgs) for Nix package manager users which follow **nixpkgs-master** channel.
+
+*Note: river in nixpkgs is prebuild with manpages and xwayland support*
+```
+nix-env -iA nixpkgs.river
+```
+
 ## Usage
 
 River can either be run nested in an X11/wayland session or directly


YerinAlexey (approved):
LGTM

ifreund:
Hmm, I don't think adding instructions on how to install river with every package manager/distro to the readme scales well. Instead I propose that we make "installl form a package manager" a wiki page (our wiki is world-editable so you can just do this) and add a link to that wiki page from the readme. Does that make sense to you?

appetrosyan:
Perhaps making the installation instructions available on the NixOS wiki?

Leon-Plickat:
What about a repology badge? That tells every user for what repos river is packaged and we don't even need to maintain it.

novakne:
Yeah repology would be the cleanest solution, wiki is better than in the README but it probably gonna be not up to date. But then I don't know how many distros have a river pkg since it doesn't have any release yet

branwright1:
Hmmm... repology badge would be great idea, forgot about it, I shouldn't add it to NixOS wiki since there is no modules for it yet. I will make module for it after first release, as modules I mean something like `windowManager.wayland.river.enable = true:` where user can add options like `xdg.congFile` and specified packages like layout managers, wofi or mako. I think we can close this pull request and ore of you can add badge.

branwright1:
Closing because of new pull request with repology badge made by someone else.

PR #266, opened 2021-04-15
README: add repology badge

novakne:
Just add a link to repology.org in the README.md, I think it's all we need

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ separate `riverctl` binary implementing it.
 
 ## Building
 
+<a href="https://repology.org/project/river/versions">
+    <img src="https://repology.org/badge/vertical-allrepos/river.svg" alt="Packaging status" align="right">
+</a>
+
 On cloning the repository, you must init and update the submodules as well
 with e.g.
 


README.md
```diff
@@ -1,5 +1,7 @@
 # river
 
+[![Packaging status](https://repology.org/badge/vertical-allrepos/river.svg)](https://repology.org/project/river/versions)
```
YerinAlexey, line 3: Maybe not as the first thing a newcomer sees? A dedicated "packaging status" section will be better IMO

novakne: I feel like I always saw this first in repos using this, but I let @ifreund decide if/where he wants it

novakne: I guess we could move it to the right side, see this [branch](https://github.com/novakne/river/tree/repology-badge-side)

YerinAlexey: > I guess we could move it to the right side, see this [branch](https://github.com/novakne/river/tree/repology-badge-side)

That looks good to me, and doesn't waste as much space

branwright1: Mayby better add it next to building? 
image

ifreund: Having this aligned to the right next to building looks good to me.

novakne: Done!

ifreund (approved):
Thanks!

PR #267, opened 2021-04-20
river: fix to work with latest zig-pixman

novakne:
I think it was just that, at least it build without error so


ifreund (changes requested):
You need to update the `zig-pixman` git submodule in the commit as well, otherwise this won't build.

novakne:
Oh yeah right my bad, done!

ifreund (approved):
Thanks!

PR #268, opened 2021-04-20
completions: Remove completions for removed subcommand "layout"

skuzzymiglet:
Also, what kind of system could be used for auto-generating completions?


YerinAlexey:
> Also, what kind of system could be used for auto-generating completions?

I'm not aware of any at the time, at least for river's setup

ifreund (approved):
Thanks! I don't know of any way to autogenerate these, but would be happy to learn of a lightweight solution.

PR #273, opened 2021-04-22
pixman: update and fix PointerConstraint

novakne:


river/PointerConstraint.zig
```diff
@@ -86,11 +86,11 @@ fn constrainToRegion(self: *Self) void {
 
         if (!self.constraint.region.containsPoint(cx, cy, &box)) {
             var nRects: c_int = undefined;
-            const rects = self.constraint.region.rectangles(&nRects);
+            const rects = self.constraint.region.rectangles();
 
             if (nRects > 0) {
```
ifreund, line 91: `nRects` is now `undefined`, you need to instead check if `rects.len > 0`

ifreund (changes requested):

river/PointerConstraint.zig
novakne: So what does nRects does now, we just need to declare it and everything happens in the background? I mean we don't need to use it, pixman handle it?

Leon-Plickat: It is no longer needed at all, if I am not mistaken. `rects` is now a slice instead of a c multi-object pointer and already contains the amount of objects it points to.

ifreund: `nRects` should be deleted, with the new `zig-pixman` api it is no longer needed. It was an out parameter to get the length o the returned data.

novakne: Ok thanks for the explanation, done!

ifreund (approved):
LGTM, thanks!

PR #277, opened 2021-04-27
river: always render floating views above layout views

ifreund:
This is a bit of an experiment, I'm not sure if I prefer this behavior in all cases yet but it does significantly help UX in some situations IMO (e.g. file dialogs and other such popups).

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -423,8 +423,9 @@ fn handleRequestSetCursor(
 }
 
 /// Find the topmost surface under the output layout coordinates lx/ly
-/// returns the surface if found and sets the sx/sy parametes to the
+/// returns the surface if found and sets the sx/sy parameters to the
 /// surface coordinates.
+/// This function must be kept in sync with the rendering order in render.zig.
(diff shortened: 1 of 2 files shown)


YerinAlexey:
It works but I would prefer to have this as an option (or even better, as a filer) because in my typical workflows I use floating windows as a "scratchpad" to say, put the terminal away when watching videos in mpv

ifreund:
> It works but I would prefer to have this as an option (or even better, as a filter) because in my typical workflows I use floating windows as a "scratchpad" to say, put the terminal away when watching videos in mpv

I'd recommend putting that terminal on a separate tag and toggling focus of that tag.

Leon-Plickat:
I prefer the current behavior, but this works.

Although maybe we can just stick floating pop-ups on top of their parent views.

ifreund:
> Although maybe we can just stick floating pop-ups on top of their parent views.

This would be significantly more complicated to implement, and still wouldn't solve this issue for all popups. Take for example the gtk pinentry popup commonly used for gpg2 which has no parent.

I've been running with this PR for 4 days now apparently and have yet to hit a case where I felt that the UX was worse than the status quo, I think I'll keep testing it for a few more days then merge this if nothing comes up.

PR #283, opened 2021-05-02
example/init: Fix typos in introductory comment and keybindings

cr-ren:
A typo in the init file uses `I` instead of `L` to increase the size of main_factor and main_count. This PR changes the occurances of `I` back to `L`.

Based on git history and the comments in example/init (and the behaviour of dwm), I am assuming `L` is the intended key.


cr-ren:
I knew there was something else! There's also a repeat of `river(1)` in the introductory comment. I am assuming it was supposed to be `rivertile(1)` because this is the only other manpage that a river build currently creates as far as I know.

YerinAlexey (approved):

ifreund (approved):
LGTM, I personally use `I` not `L` here because of colemak :D

PR #286, opened 2021-05-07
Allow changing output focus with pointer

Leon-Plickat:
This patch makes it possible to use the pointer to focus views on an unfocused output.

TODO: Clicking on an empty desktop should also change focus.


YerinAlexey (approved):

ifreund (approved):
LGTM, this is a very nice usability upgrade for multi monitor setups :)

PR #288, opened 2021-05-08
protocol: replace unstable with staging

### LINKED ISSUE #262: Protocols: s/unstable/staging
See https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/55

novakne:
see:
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/55

* move river protocols to staging folder
* remove the prefix z from interface name
* remove unstable from file name
* add a disclaimer, but maybe it could be move somewhere better

Move wlr protocols to unstable as said in wayland README:
'Unstable protocols (see Legacy protocol phases)
can be found in the unstable/ directory, but new ones should never be
placed here.'

Not sure if you want to have folder staging/unstable or not

Closes #262


ifreund:
I'm not sure renaming these before there is a breaking change is worth it. In particular, renaming river-status would break waybar and yambar.

novakne:
> In particular, renaming river-status would break waybar and yambar.

Oh I totally didn't thought of that... I guess you're right then it's probably not a godd idea

novakne:
So should I close it or there is some change you want to keep?

ifreund:
I think we can close this, I think we should avoid breaking `river-status` until necessary.

PR #291, opened 2021-05-15
rivertile: clamp main_amount

### LINKED ISSUE #290: riverctl generates unexpected layout when main_count is greater than the maximum amount of views
Main views will not use all available space.

Also max_amount smaller than 1 are not possible, not sure if that is intended behaviour though.

IMG_20210515_130601_286

Leon-Plickat:
fixes #290


ifreund (approved):
LGTM, thanks!

PR #303, opened 2021-06-04
Add basic input configuration

Leon-Plickat:
Since our original idea of doing this via a Wayland protocol will probably not happen in the near future and since that one hack to configure libinput isn't very pretty I decided to experiment with simple input configuration in river itself. No idea if this will lead anywhere...

Keyboard configuration is deliberately left out for now, this work is focusing on libinput.

#### TODO

* [x] Implement `wlr.InputDevice.getLibinputDeviceHandle()`
* [x] Implement ~~all the common~~ *a lot of the* libinput configuration things
* [x] Add a command to add one configuration to one input device, matching by identifier
* [X] Apply to new devices
* [x] Apply to existing devices


doc/riverctl.1.scd
```diff
@@ -268,6 +268,69 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
 	and is made available through the _XCURSOR_THEME_ and _XCURSOR_SIZE_
 	environment variables.
 
+## INPUT CONFIGURATION
+
+The _input_ command can be used to create a configuration rule for an input
+device identified by its _name_.
+
+Below find a lost of all device properties that can be configured. However note
```
ifreund, line 276:
```suggestion
A list of all device properties that can be configured may be found below. However note
```

doc/riverctl.1.scd
```diff
@@ -268,6 +268,69 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
...
+that not every input device supports every property.
+
+*input* _name_ *events* *enabled*|*disabled*|*disabled-on-external-mouse*
```
ifreund, line 279: What does `disabled-on-external-mouse` mean? Should we perhaps point users to libinput docs on these options?

river/InputManager.zig
```diff
@@ -34,6 +37,43 @@ const default_seat_name = "default";
...
+    identifier: []const u8,
+
+    pub fn init(self: *InputDevice, device: *wlr.InputDevice) !void {
+        // The identifier is formatted exactly as in Sway
+        var identifier = try std.fmt.allocPrint(
+            util.gpa,
+            "{}:{}:{s}",
+            .{ device.vendor, device.product, mem.trim(
+                u8,
+                mem.spanZ(device.name),
```
ifreund, line 55: https://github.com/ziglang/zig/pull/8698
```suggestion
                mem.sliceTo(device.name, 0),
```

river/InputManager.zig
```diff
@@ -34,6 +37,43 @@ const default_seat_name = "default";
...
+    device: *wlr.InputDevice,
+
+    // Careful: The identifier is not unique! A physical input device may have
+    // multiple logical input devices with the exact same vendor id, product id
+    // and name. However identifiers of InputConfigs are unique.
+    identifier: []const u8,
+
+    pub fn init(self: *InputDevice, device: *wlr.InputDevice) !void {
+        // The identifier is formatted exactly as in Sway
+        var identifier = try std.fmt.allocPrint(
```
ifreund, line 50:
```suggestion
        const identifier = try std.fmt.allocPrint(
```

river/InputManager.zig
```diff
@@ -34,6 +37,43 @@ const default_seat_name = "default";
 
 const log = std.log.scoped(.input_manager);
 
+pub const InputDevice = struct {
+    device: *wlr.InputDevice,
+
+    // Careful: The identifier is not unique! A physical input device may have
+    // multiple logical input devices with the exact same vendor id, product id
+    // and name. However identifiers of InputConfigs are unique.
```
ifreund, line 45: should be a `///` doc comment

river/command/input.zig
```diff
@@ -0,0 +1,108 @@
...
+    var conf_it = server.input_manager.input_configs.first;
+    while (conf_it) |conf_node| : (conf_it = conf_node.next) {
+        if (mem.eql(u8, conf_node.data.identifier, args[1])) {
+            input_config = &conf_node.data;
+            break;
+        }
+    }
+
+    // If no matching InputConfig has been found, create one.
+    if (input_config == null) {
```
ifreund, line 51: Instead of making `input_config` optional and checking if it is null here, it would be more idiomatic to use the else clause of the while loop to handle the case in which it does not exist:
```
const input_config = while (...) {
} else blk: {
     // create input config
     break :blk input_config;
};
```
with this approach the `.?` on `input_config` throughout the rest of this function can be avoided.

river/command/input.zig
```diff
@@ -0,0 +1,108 @@
...
+        input_config.?.tap_button_map = try InputConfig.TapButtonMap.parse(args[3]);
+    } else if (mem.eql(u8, "pointer-accel", args[2])) {
+        input_config.?.pointer_accel = try InputConfig.PointerAccel.parse(args[3]);
+    } else if (mem.eql(u8, "scroll-method", args[2])) {
+        input_config.?.scroll_method = try InputConfig.ScrollMethod.parse(args[3]);
+    } else if (mem.eql(u8, "scroll-button", args[2])) {
+        input_config.?.scroll_button = try InputConfig.ScrollButton.parse(args[3]);
+    }
+
+    // Update matching existing input devices, if exist.
```
ifreund, line 99:
```suggestion
    // Update matching existing input devices
```

river/InputConfig.zig
```diff
@@ -0,0 +1,474 @@
...
+// TODO - keyboards
+//      - mapping to output / region
+//      - calibration matrices
+//      - scroll factor
+
+pub const EventState = enum {
+    unset,
+    enabled,
+    disabled,
+    disabled_on_external_mouse,
```
ifreund, line 43: If we used `@"disabled-on-external-mouse"` instead the parse function could be replaced by `std.meta.stringToEnum()`, which would be cleaner IMO. The same goes for all the other enums in this file.

river/InputConfig.zig
```diff
@@ -0,0 +1,474 @@
...
+    pub fn apply(event_state: EventState, device: *c.libinput_device) void {
+        const want = switch (event_state) {
+            .unset => return,
+            .enabled => c.LIBINPUT_CONFIG_SEND_EVENTS_ENABLED,
+            .disabled => c.LIBINPUT_CONFIG_SEND_EVENTS_DISABLED,
+            .disabled_on_external_mouse => c.LIBINPUT_CONFIG_SEND_EVENTS_DISABLED_ON_EXTERNAL_MOUSE,
+        };
+        const current = c.libinput_device_config_send_events_get_mode(device);
+        if (want != current)
+            _ = c.libinput_device_config_send_events_set_mode(device, @intCast(u32, want));
```
ifreund, line 66: Not a hard rule since I don't have any way to enforce it, but the style I've been moving towards for if statements is no braces if they fit on one line but always braces if they don't:
```zig
if (foo) bar(); // good

if (foo) {
    bar(); // good
}

if (foo)
    bar(); // I don't like this
```

ifreund (commented):
Here's some initial code review, I haven't done any actual testing yet.

doc/riverctl.1.scd
novakne: it means 'disable the device while an external mouse is plugged in', not sure if there is a man page instead of https://wayland.freedesktop.org/libinput/doc/latest/configuration.html

Leon-Plickat:
@ifreund  Your first round of reviews managed to knock off about 200 lines!

doc/riverctl.1.scd
```diff
@@ -268,6 +268,69 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
 	and is made available through the _XCURSOR_THEME_ and _XCURSOR_SIZE_
 	environment variables.
 
+## INPUT CONFIGURATION
+
+The _input_ command can be used to create a configuration rule for an input
+device identified by its _name_.
```
ifreund, line 274: How does one find out the device names? I'm assuming there's some command to list them but the average user probably isn't aware of it.

river/command/input.zig
```diff
@@ -0,0 +1,121 @@
...
+            input_config.apply(&device_node.data);
+            // We don't break here because it is common to have multiple input
+            // devices with the same identifier.
+        }
+    }
+}
+
+fn parse(comptime T: type, str: []const u8) !T {
+    if (std.meta.stringToEnum(T, str)) |ret| {
+        if (ret != .unset) return ret;
```
ifreund, line 118: We could avoid this check (and the existence of this wrapper function) by removing the `unset` field from the enums and making the fields of `InputConfig` `?EnumFoo` instead, defaulting to `null`.

doc/riverctl.1.scd
Leon-Plickat: There isn't a good way currently. The best way is either to read the logs for "new input device" messages, or to hop into sway and use `swaymsg -t get_inputs`.

ifreund: looks like libinput ships a command line tool that can provide this information. `libinput list-devices` prints off a list for me, although the names aren't in the format we expect.

Leon-Plickat: I decided to use the same scheme to build the identifiers as Sway does (`<vendor id>:<product id>:<name>`) for the sake of consistency.

The identifier sadly can't be just the advertised device name, as that would lead to even more duplicate identifiers.

`libinput list-devices` does not show the vendor or device id unfortunately.

novakne: Would it be hard to implement a command like `riverctl devices-info` that return the devices identifier, maybe using json? Could be a better user friendly way than looking at the log

river/command/input.zig
```diff
@@ -0,0 +1,154 @@
...
+    } else if (mem.eql(u8, "scroll-method", args[2])) {
+        input_config.scroll_method = std.meta.stringToEnum(InputConfig.ScrollMethod, args[3]) orelse
+            return Error.UnknownOption;
+    } else if (mem.eql(u8, "scroll-button", args[2])) {
+        const event_code_name = try std.cstr.addNullByte(allocator, args[3]);
+        defer allocator.free(event_code_name);
+        const ret = c.libevdev_event_code_from_name(c.EV_KEY, event_code_name);
+        if (ret < 1) return Error.InvalidButton;
+        input_config.scroll_button = InputConfig.ScrollButton{ .button = @intCast(u32, ret) };
+    }
```
ifreund, line 143: We need an `else` branch here to error on an unknown subcommand.

doc/riverctl.1.scd
```diff
@@ -268,6 +268,72 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
...
+	- _none_: No scrolling
+	- _two-finger_: Scroll by swiping with two fingers simultaneously
+	- _edge_: Scroll by swiping along the edge
+	- _button_: Scroll with pointer movement while holding down a button
+
+*input* _name_ *scroll-button* _button_
+	Set the scroll button of an input device. _button_ is the name of a linux
+	input event code.
+
+*list-inputs*
```
ifreund, line 334: can we put this first in the input section to make it a bit more discoverable?

ifreund (commented):
I hit a panic while testing:
```
thread 3392 panic: invalid enum value
???:?:?: 0x294216 in ??? (???)
/home/ifreund/projects/river/river/command/input.zig:40:36: 0x21b291 in command.input.listInputs (river)
        try input_list.appendSlice(@tagName(node.data.device.type));
                                   ^
/home/ifreund/projects/river/river/command.zig:128:16: 0x2538f1 in command.run (river)
    try impl_fn(allocator, seat, args, out);
               ^
/home/ifreund/projects/river/river/Control.zig:107:24: 0x23e55f in Control.handleRequest (river)
            command.run(util.gpa, seat, args.items, &out) catch |err| {
                       ^
/home/ifreund/projects/river/zig-cache/zig-wayland/common.zig:101:76: 0x2546e9 in .wayland.common.Dispatcher(.wayland.river_control_unstable_v1_server.ControlV1,*Control).dispatcher (river)
                    @ptrCast(fn (*Obj, Payload, Data) void, implementation)(
                                                                           ^
```

Also, we should probably add completions for these new riverctl subcommands...

Leon-Plickat:
For that panic to happen, I assume `wlr.InputDevice.type` was set to a value not in `wlr.InputDevice.Type`. Since that struct is `extern` the problem may not be on our end.

Leon-Plickat:
Did you unplug the mouse and plugged it back in again? I think I forgot to handle input device removals.

Leon-Plickat:
Removed devices are now handled. That panic should no longer happen.

river/InputManager.zig
```diff
@@ -43,6 +95,8 @@ relative_pointer_manager: *wlr.RelativePointerManagerV1,
 virtual_pointer_manager: *wlr.VirtualPointerManagerV1,
 virtual_keyboard_manager: *wlr.VirtualKeyboardManagerV1,
 
+input_configs: std.SinglyLinkedList(InputConfig) = .{},
```
ifreund, line 98: We don't seem to store pointers to `InputConfig`s anywhere which would require pointer stability. Therefore, I think we could just use an `ArrayList` (or `ArrayListUnmanaged`) here which should simplify things a bit.

ifreund (commented):
Nice, that seems to have fixed things as I can no longer reproduce the panic. Just one more simplification I noticed then I think this is good to merge!

Leon-Plickat:
I wonder if we can use an ArrayList for `InputManager.input_devices` as well.

Leon-Plickat:
I also decided to add a `list-input-configs` command, which I found useful for testing.

river/command/input.zig
```diff
@@ -66,27 +65,19 @@ pub fn input(
...
-        new_input_config_node = try util.gpa.create(std.SinglyLinkedList(InputConfig).Node);
-        errdefer util.gpa.destroy(new_input_config_node.?);
-        new_input_config_node.?.data = .{
-            .identifier = try util.gpa.dupe(u8, args[1]),
-        };
-        server.input_manager.input_configs.prepend(new_input_config_node.?);
-        break :blk &new_input_config_node.?.data;
+        const ic = try server.input_manager.input_configs.addOne();
+        errdefer _ = server.input_manager.input_configs.pop();
+        ic.* = .{ .identifier = try util.gpa.dupe(u8, args[1]) };
```
ifreund: I'd do the following as the control flow is a bit simpler.
```zig
try server.input_manager.input_configs.ensureUnusedCapacity(1);
server.input_manager.input_configs.appendAssumeCapacity(.{ .identifier = try util.gpa.dupe(u8, args([1]) });
```

river/InputManager.zig
```diff
@@ -95,7 +95,7 @@ relative_pointer_manager: *wlr.RelativePointerManagerV1,
 virtual_pointer_manager: *wlr.VirtualPointerManagerV1,
 virtual_keyboard_manager: *wlr.VirtualKeyboardManagerV1,
 
-input_configs: std.SinglyLinkedList(InputConfig) = .{},
+input_configs: std.ArrayList(InputConfig) = undefined,
```
ifreund, line 98: We should not use a default value here, as omitting `input_configs` when initializing `InputManager` would be a bug.

river/command/input.zig
```diff
@@ -0,0 +1,238 @@
...
+    var input_list = std.ArrayList(u8).init(allocator);
+
+    for (server.input_manager.input_configs.items) |*input_config| {
+        try input_list.appendSlice(input_config.identifier);
+        try input_list.appendSlice("\n");
+
+        if (input_config.event_state) |event_state| {
+            try input_list.appendSlice("\tevents: ");
+            try input_list.appendSlice(@tagName(event_state));
+            try input_list.appendSlice("\n");
```
ifreund, line 72: We can use the writer interface and std.fmt to make this code less verbose. 
```suggestion
            try writer.print("\tevents: {}\n", .{event_state});
```

river/command/input.zig
```diff
@@ -0,0 +1,238 @@
...
+    out.* = input_list.toOwnedSlice();
+}
+
+pub fn listInputConfigs(
+    allocator: *mem.Allocator,
+    seat: *Seat,
+    args: []const []const u8,
+    out: *?[]const u8,
+) Error!void {
+    var input_list = std.ArrayList(u8).init(allocator);
```
ifreund, line 63:
```suggestion
    var input_list = std.ArrayList(u8).init(allocator);
    const writer = input_list.writer();
```

ifreund (commented):
> I wonder if we can use an ArrayList for InputManager.input_devices as well.

We can't as we need the input devices to be pinned in memory due to the destroy listener.

> I also decided to add a list-input-configs command, which I found useful for testing.

Nice, this is certianly useful for testing and likely for users as well :)

ifreund:
Oh, and the new commands are missing from the completion scripts currently. It would be nice to add them but I won't block merging this on that. (It would block 0.1.0 though)

river/command/input.zig
```diff
@@ -0,0 +1,220 @@
...
+    const writer = input_list.writer();
+    var prev = false;
+
+    for (server.input_manager.input_configs.items) |*input_config| {
+        if (prev) try input_list.appendSlice("\n");
+        prev = true;
```
ifreund, line 73: Ok, I promise this is my last bikeshed before merging this :D
```suggestion

    for (server.input_manager.input_configs.items) |*input_config, i| {
        if (i > 0) try input_list.appendSlice("\n");
```

ifreund (approved):
Great work on this, LGTM!

PR #304, opened 2021-06-08
config: make attach-mode global

### LINKED ISSUE #269: Make `attach-mode` global instead of per-output
This adds the ability to set `attach-mode` globally.
1 more reason is all the other configurations are consistent everywhere, and only `attach-mode` has to rely on the focused output.

novakne:
I don't have multi output to try it

Closes #269


ifreund (approved):
Thanks!

PR #311, opened 2021-06-14
Add directional output focusing

Leon-Plickat:
Directional focusing for outputs makes sense for complex output configurations, where the list based focusing can be tedious. Humans are very good at spacial navigation, so let's make use of that.


Leon-Plickat:
Also made `send-to-output` accept spacial directions and moved both commands into one file.

ifreund:
What would you think of removing the list based output focusing in favor of this?

river/command/output.zig
```diff
@@ -0,0 +1,105 @@
...
+        destination_output.arrangeViews();
+        server.root.startTransaction();
+    }
+}
+
+/// Find an output adjacent to the currently focused based on either logical or
+/// spacial direction
+fn getOutput(seat: *Seat, str: []const u8) !?*Output {
+    if (std.meta.stringToEnum(Direction, str)) |direction| { // Logical direction
+        // Focus the next/prev output in the list if there is one, else wrap
```
ifreund, line 81:
```suggestion
        // Return the next/prev output in the list if there is one, else wrap
```

river/command/output.zig
```diff
@@ -0,0 +1,105 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+const std = @import("std");
+
+const wl = @import("wlroots");
```
ifreund, line 20: calling this `wl` is confusing, `wl` is usually `@import("wayland").server.wl`.
```suggestion
const wlr = @import("wlroots");
```

river/command/output.zig
```diff
@@ -0,0 +1,105 @@
...
+            @intToFloat(f64, focus_box.x + @divFloor(focus_box.width, 2)),
+            @intToFloat(f64, focus_box.y + @divFloor(focus_box.height, 2)),
+        ) orelse return null;
+        var it = server.root.outputs.first;
+        while (it) |node| : (it = node.next) {
+            if (node.data.wlr_output == wlr_output) {
+                return &node.data;
+            }
+        }
+        return null;
```
ifreund, line 101: No need to iterate here, just `@ptrToInt(*Output, wlr_output.data);`

ifreund (commented):
> I use both. On my laptop I use the list based approach, because since it wraps it works well for two outputs (I just have Super+Tab to switch to the other output, which works because the list commands wrap). On my desktop I have recently installed a third screen and the spacial focusing works better for this.

Makes sense to me, supporting both isn't a huge cost.

river/command/output.zig
Leon-Plickat: typo

river/command/output.zig
Leon-Plickat: I always forget the wlroots objects have user data.

Leon-Plickat: But you do mean `@intToPtr()`, right?

ifreund: > But you do mean @intToPtr(), right?

Yeah definitely

ifreund (approved):
LGTM, thanks!

PR #313, opened 2021-06-14
completions/zsh: add input commands

novakne:
Cleanup a bit too

Feel like the script is a little hacky, but as far as I understand `_arguments` only works with args like `-c |--commands` so I need to use `_alternative`, a bit redundant


ifreund (approved):
Thanks, this seems to work well and I don't need to mess around with zsh scripting :)

PR #314, opened 2021-06-14
add river-toplevel-v1

Leon-Plickat:


protocol/river-toplevel-v1.xml
```diff
@@ -0,0 +1,110 @@
...
+        Seont once when binding the interrface and again whenever the tags
+        of the toplevel change.
+      </description>
+      <arg name="floating" type="bool"/>
+    </event>
+
+    <request name="set_tags">
+      <description summary="set the tags of the toplevel">
+        The client can use this request to change the pending tags of the view.
+        This state is double buffered and only applied when comitting.
```
novakne, line 89:
```suggestion
        This state is double buffered and only applied when committing.
```

protocol/river-toplevel-v1.xml
```diff
@@ -0,0 +1,110 @@
...
+      </description>
+      <arg name="tags" type="uint" summary="32-bit bitfiled"/>
+    </request>
+
+    <request name="set_floating">
+      <description summary="set the floating state of the toplevel">
+        The client can use this request to change the pending floating state of
+        the view. This state is double buffered and only applied when comitting.
```
novakne, line 97:
```suggestion
        the view. This state is double buffered and only applied when committing.
```

protocol/river-toplevel-v1.xml
```diff
@@ -0,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="river_layout_v2">
```
novakne, line 2:
```suggestion
<protocol name="river_toplevel_v1">
```

novakne (commented):
Just some typo reading through it

Leon-Plickat:
@novakne thanks!

Leon-Plickat:
The current idea is to let river-toplevel be a protocol purely for fetching river specific view stats. Controlling views will be done with river-control-v2 (because we need a way to operate on an unspecified view, which river will interpret as an operation on the focused view).

**What information do we want to advertise?**

* tags
* floating

Do we need a `done` event? We can probably re-use the one of the zwlr-foreign-toplevel-handle, so likely no.

novakne:
Would it be possible to get the positions/dimensions of a view?

Leon-Plickat:
Certainly, but I can't think of a use case for that.

novakne:
Right I guess for now it would be just more info, I don't have any use case too. Maybe it could be useful if river add a -tags flag to open view in specific tags or at a specific dimensions/positions

snakedye:
> Certainly, but I can't think of a use case for that.

Tag preview

Leon-Plickat:
> Tag preview

That won't work.

Not focused tags are not layed out, therefore the window positions you'd use for a preview have a very high chance of being garbage.

Leon-Plickat:
Note to self: add urgency

novakne:
So I was reading `sway-ipc(7)` man page to get some inspiration, and I see it returns the `current_border_width`, not sure if this is the right protocol to have this, but it would be really useful to add smart border

marcosrdac:
Is there any way that we could also loop through view windows and know which one is focused? It would be really useful.

I wanted to be able to call my scratchpad windows and also **focus one of them**, then toggle them all and get focus to my previous window.

Leon-Plickat:
@marcosrdac this protocol is intended to extent the foreign-toplevel protocol, which already lets you know which views are active ("focused").

However "looping through" is not how any of this will work.

PR #324, opened 2021-06-15
completions: add fish completion for input command

FollieHiyuki:
I can't find a way to use switch case, so I resolved to use 3 "same" functions.


novakne:
You can remove all the completion from the river-option protocol, get-option, set-option, unset-option, mod-option

FollieHiyuki:
> You can remove all the completion from the river-option protocol, get-option, set-option, unset-option, mod-option

I completely missed them. Removed.

ifreund (approved):
Thanks, adding completion for input devices as well was a nice touch!

PR #328, opened 2021-06-16
Direct scanout for fullscreen views

Leon-Plickat:


river/render.zig
```diff
@@ -47,13 +47,65 @@ const SurfaceRenderData = struct {
...
+    if (output.getLayer(.overlay).first != null) return false;
+    if (server.root.drag_icons.first != null) return false;
+
+    if (view.saved_buffers.items.len != 0) return false;
+    if (view.surface == null or view.surface.?.buffer == null) return false;
+
+    // If the surface has subsurfaces we need to composite.
+    var surfaces: u32 = 0;
+    view.surface.?.forEachSurface(*u32, countSurfaceIterator, &surfaces);
+    if (surfaces > 1) return false;
```
ifreund, line 72: cant we just check if `surface.subsurfaces_{above,below}` are empty instead of iterating?

Leon-Plickat: Yes we can.

Issue #330, opened 2021-06-17
Some configuration is irrreversible


Leon-Plickat:
I think it would be nice if all configuration options exposed via `riverctl` / the river-control protocol would be reversible. That way, any possible configuration state could be reached from any other possible configuration state without restarting river. Currently that is not possible and to reach certain configuration states river needs to be restarted.

#### Irreversible options:
* `float-filter-add`
* `csd-filter-add`
* `declare-mode`
* configuring inputs

ifreund:
I agree that having all configuration be reversible is desirable as needing to restart the compositor is annoying. One idea I had with regards to this was a command that returns river to its initial state with regards to config. This would include restoring things like border color, key/pointer mappings, etc. but would not change the currently focused tags for example. Putting this command first in one's init script would essentially make re-running the init script effectively the same as sway's in-place reload command.

> Also the filter commands do not affect already existing windows, which is inconsistent with the rest of rivers commands.

I think this behavior is correct as-is. These filters do not force matched views to remain floating or whatever, they merely change the default on mapping the view.

sheerun:
@ifreund What is current recommended way to reload river after making changes to ~/.config/river/init? Exit river with super+shift+e and type river again?

Leon-Plickat:
Just re-run your init script.

Leon-Plickat:
On Mon Oct 23, 2023 at 4:39 PM CEST, Max wrote:
> For example removing a keybind set previously requires you to restart river as described

Keybinds are reversible: you can use `rivercl unmap`

MaxVerevkin:
What about `riverctl unmap -all`? Or something more generic, like `riverctl reset`?

PR #331, opened 2021-06-19
Add commands to remove entries from the filter lists & update all applicable view on CSD filter list change

Leon-Plickat:
This belongs to #330 

Maybe there should also be commands to clear filters, meaning removing all entries?


YerinAlexey (approved):

ifreund (commented):
What do you think of the "reset all config" command I proposed here https://github.com/ifreund/river/issues/330#issuecomment-863744617? If we were to implement that I'm not sure we need individual remove commands for everything...

river/command/filter.zig
```diff
@@ -23,14 +23,25 @@ const util = @import("../util.zig");
...
+    for (list.items) |*filter, i| {
+        if (std.mem.eql(u8, filter.*, args[1])) {
+            if (operation == .remove) {
+                allocator.free(list.orderedRemove(i));
+            }
+            return;
+        }
+    }
+    if (operation == .add) {
+        try list.append(try std.mem.dupe(allocator, u8, args[1]));
```
ifreund, line 43: This will leak memory currently if the append() fails bu the dupe doesn't. To fix this we can first call `ensureUnususedCapacity(1)` and then use `appendAssumeCapacity()`.

doc/riverctl.1.scd
```diff
@@ -23,14 +23,24 @@ over the Wayland protocol.
 *csd-filter-add* _app-id_
 	Add _app-id_ to the CSD filter list. Views with this _app-id_ are
 	told to use client side decoration instead of the default server
-	side decoration.
+	side decoration. Note that this affects only new views, not already
+	existing ones.
```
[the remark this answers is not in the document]
Leon-Plickat: I think only applying to new views makes sense if we had a command to toggle the CSD mode of a view. That way it would be identical to the float filter.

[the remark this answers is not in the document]
Leon-Plickat: I'd prefer if we kept the behaviour of the different filters consistent.

river/command/filter.zig
Leon-Plickat: Actually, according to `git blame` it **is** my fault :D

Leon-Plickat:
I like the idea of a reset command.

The filter remove commands where just an idea for consistency, since `map` and `map-pointer` also have inverse commands.

river/command/filter.zig
```diff
@@ -52,9 +49,82 @@ pub fn csdFilterAdd(
...
+    if (operation == .add) {
+        try list.ensureUnusedCapacity(1);
+        list.appendAssumeCapacity(try std.mem.dupe(allocator, u8, args[1]));
+    }
+}
+
+fn csdFilterUpdateViews(app_id: []const u8, operation: enum { add, remove }) void {
+    // There is no link between Decoration and View, so we need to iterate over
+    // both separately. Note that we do not need to arrange the outputs here; If
+    // the clients decoration mode changes, it will receive a configure event.
```
ifreund, line 91: There is a link from Decoration to View, so there's no need to iterate over the views.
```zig
    const view = @ptrToInt(*View, decoration.xdg_toplevel_decoration.surface.data);
````

File: river/command/filter.zig [hunk 4]
```diff
@@ -52,9 +49,82 @@ pub fn csdFilterAdd(
...
+    var decoration_it = server.decoration_manager.decorations.first;
+    while (decoration_it) |decoration_node| : (decoration_it = decoration_node.next) {
+        const xdg_toplevel_decoration = decoration_node.data.xdg_toplevel_decoration;
+        if (std.mem.eql(
```
ifreund, line 95: I'd stick a `const mem = std.mem;` at the top of this file, you use it a fair bit.

river/command/filter.zig
```diff
@@ -52,9 +49,82 @@ pub fn csdFilterAdd(
...
+    try modifyFilter(allocator, &server.config.csd_filter, args, .remove);
+    csdFilterUpdateViews(args[1], .remove);
+}
+
+fn modifyFilter(
+    allocator: *std.mem.Allocator,
+    list: *std.ArrayList([]const u8),
+    args: []const []const u8,
+    operation: enum { add, remove },
+) Error!void {
```
ifreund, line 71: I think handling add and remove in the same function doesn't make much sense. The only reason I see for this choice is avoiding duplication of the length checking for `args`, which I don't think is worth it. That length checking is very simple and already duplicated a million places.

ifreund (changes requested):

ifreund (approved):
LGTM now, I just finished this up as I needed it to test my gtk xdg-decoration patch.

PR #341, opened 2021-06-23
Reset cursor state on output or tag focus change

Leon-Plickat:
~~Not sure if this is the right approach.~~ It works, so it'll be fine.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -20,6 +20,7 @@ const Self = @This();
 const build_options = @import("build_options");
 const std = @import("std");
 const assert = std.debug.assert;
+const os = std.os;
 const math = std.math;
 const wlr = @import("wlroots");
 const wayland = @import("wayland");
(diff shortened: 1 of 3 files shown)


river/Cursor.zig
```diff
@@ -133,6 +133,19 @@ pub fn deinit(self: *Self) void {
...
+/// intended to be used when the output or tag focus changes.
+pub fn resetState(self: *Self) void {
+    log.debug("reset state, current mode is {s}", .{@tagName(self.mode)});
+
+    // Go back to .passthrough mode. Note that if the previous mode was .down,
+    // we do not send button release events. This is on purpose.
+    self.mode = .passthrough;
+
+    const now = util.getNowMsec() catch return;
+    self.passthrough(now);
```
Leon-Plickat, line 146: I wonder if we can instead get the time from the event that caused the tag/output focus change. Might be annoying to pass around though.

river/Seat.zig
```diff
@@ -64,6 +64,9 @@ prev_mode_id: usize = 0,
 /// is currently available for focus.
 focused_output: *Output,
 
+/// Has the output focus changed since the last transaction?
+output_focus_changed: bool = false,
```
Leon-Plickat, line 68: Should `Seat` perhaps also have a `State` struct with a `pending` and `current` instance?

river/Cursor.zig
ifreund: I don't think that would really be worth it.

ifreund (commented):
I think it would make things a lot cleaner if we could call `Cursor.resetState()` at the end of *every* transaction. To make this possible, 2 things would be needed:
1. store the position of the view at the time the mode is entered in `Cursor.Mode` for down/move/resize and rework `processMotion()` to be relative to the stored position.
2. change `resetState()` to only revert to passthrough mode if the target view is not shown on the currently focused output.

I think with these proposed changes, calling `resetState()` for each seat at the end of every transaction would achieve the correct results in all cases.

What do you think?

Leon-Plickat:
>  store the position of the view at the time the mode is entered in Cursor.Mode for down/move/resize and rework processMotion() to be relative to the stored position.

Started working on this again and this caught my eye. I don't see why we'd need that.

ifreund:
> Started working on this again and this caught my eye. I don't see why we'd need that. 

I believe this would be needed to properly handle the case where a view is moved during a transaction but the cursor is still over it.

Leon-Plickat:
> I believe this would be needed to properly handle the case where a view is moved during a transaction but the cursor is still over it.

I don't think it is. In my testing, starting a transaction while a view is currently moved with the pointer works as expected.

The problem is resizing: If you start resizing the view, switch to a different tag, switch to the previous tag again and release the mouse button over the view you resized, then moving the view with the pointer acts out.

ifreund:
Hmm, I think I haven't been very clear on what I think the intended behavior here looks like. If a transaction starts while in cursor mode `down`, `move`, or `resize` and this transaction causes the target view to no longer be visible on the output, I think we should return to `passthrough` mode and send the pointer leave event if we were in `down` mode.

I think that should avoid the problem you mentioned with resizing as we stop the resize when the view is moved to a different tag and made no longer visible.

We should also abort an ongoing move or resize if the target view becomes fullscreen or is returned to the layout I believe.

Handling `down` mode properly is a little different. This is because unlike `move` and `resize` which cause the view to immediately become floating, the target view of the `down` mode may be part of the layout and moved as the result of a transaction. I think the ideal behavior here is that we only exit the mode before the pointer is released if the transaction causes the view to no longer be visible. If the view remains visible we should continue to send the pointer events to the view, relative to its new position. Furthermore, I don't think we need to exit down mode if the view's floating/fullscreen state is modified as long as the view is still visible.

So I think you are correct and there's no need to store the position of the view when then cursor mode is entered in order to implement these semantics, sorry for the confusion.

Do you agree with these proposed semantics? I haven't had a chance to give your current implementation a read yet but should have time sunday or early next week to dig into the code.

ifreund (approved):
LGTM, thanks for fixing this! Just made a few minor tweaks to my taste and improved the commit message as I started other work based off of this locally.

PR #346, opened 2021-07-02
completions: typo

steef435:
Hello,

I noticed a typo in the bash completions while configuring river.

Thank you for the compositor!


ifreund (approved):
Thanks, glad that river is useful to you!

PR #354, opened 2021-07-11
Feat/last focused tags

nagy135:
Adds feature ~~*last-focused-tags*~~ *focus-previous-tags*

* selects previously selected tags (hooks added to functions that change them)
* updates docs/completions

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -139,6 +139,10 @@ are ignored by river.
 	will be assigned the tags 000010001. If no tags would remain after
 	filtering, the _tagmask_ is ignored.
 
+*focus-previous-tags*
+	Sets tags to their previous value on the currently focused output,
+	allowing jumping back and forth between 2 tag setups.
+
 ## MAPPINGS
 
 Mappings are modal in river. Each mapping is associated with a mode and is
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -72,6 +72,9 @@ views: ViewStack(View) = .{},
 current: State = State{ .tags = 1 << 0 },
 pending: State = State{ .tags = 1 << 0 },
 
+/// Remembered version of tags (from last run)
+previous_tags: u32 = 1 << 0,
+
 /// The currently active LayoutDemand
 layout_demand: ?LayoutDemand = null,
 
--- a/river/command/tags.zig
+++ b/river/command/tags.zig
@@ -31,6 +31,7 @@ pub fn setFocusedTags(
 ) Error!void {
     const tags = try parseTags(allocator, args, out);
     if (seat.focused_output.pending.tags != tags) {
+        seat.focused_output.previous_tags = seat.focused_output.pending.tags;
         seat.focused_output.pending.tags = tags;
         seat.focused_output.arrangeViews();
         seat.focus(null);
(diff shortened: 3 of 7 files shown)


river/command/tags.zig
```diff
@@ -102,6 +104,28 @@ pub fn toggleViewTags(
...
+    allocator: *std.mem.Allocator,
+    seat: *Seat,
+    args: []const []const u8,
+    out: *?[]const u8,
+) Error!void {
+    const last_tags = seat.focused_output.pending.last_tags;
+    if (last_tags == 0) {
+        out.* = try std.fmt.allocPrint(allocator, "no remembered tags found on this output", .{});
+        return Error.Other;
+    }
```
ifreund, line 119: Instead of having this special case, why not just initialize `last_tags` to 1 and assert that it is `!= 0` here?

river/Output.zig
```diff
@@ -42,6 +42,9 @@ const State = struct {
     /// A bit field of focused tags
     tags: u32,
 
+    /// Remembered version of tags (from last run)
+    last_tags: u32,
+
```
ifreund, line 47: This doesn't need to go in the state struct, as it doesn't need double-buffering. We can just put it in `Output` directly. Notice how `Output.current.last_tags` is never accessed in your code.

doc/riverctl.1.scd
```diff
@@ -117,6 +117,7 @@ are ignored by river.
 	Toggle visibility of views with tags corresponding to the set bits
 	of _tags_ on the currently focused output.
 
+
```
ifreund, line 120:
```suggestion
```

ifreund (changes requested):
Here's some code review, I still need to play with this and see how it feels with tags as opposed to workspaces though.

It would also be nice to add bash completions, which should be quite straightforward.

novakne:
Tested with no problem on my side, I tried with a key bind, could be interesting

river/command/tags.zig
nagy135: You are absolutelly right, I got stuck in bspwm world where was not always the same. Will commit update, it should simplify it a lot

river/Output.zig
nagy135: I tried to do what you suggested. I must add I never coded in zig, its just so readable small edit like this is possible even for a novice. Works on my end still

ifreund (changes requested):
Had a chance to try this out today and it feels quite nice. Great idea!

I think we should rename the command to `focus-last-tags` or `focus-previous-tags` though so that it has a verb. `riverctl focus-previous-tags` reads a lot better than the current `riverctl last-focused-tags` which sounds like it will output the tags which were last focused.

If we rename the command to `focus-previous-tags` it would make sense to change the variable name from `last_tags` to `previous_tags` as well.

After that, this should be good to merge!

nagy135:
> Had a chance to try this out today and it feels quite nice. Great idea!

Glad you like it! This was for sure the most important feature I missed! It frees your attention to not have to remember "where you came from".

> I think we should rename the command to `focus-last-tags` or `focus-previous-tags` though so that it has a verb. `riverctl focus-previous-tags` reads a lot better than the current `riverctl last-focused-tags` which sounds like it will output the tags which were last focused.
> 
> If we rename the command to `focus-previous-tags` it would make sense to change the variable name from `last_tags` to `previous_tags` as well.
> 
> After that, this should be good to merge!

Since we have `toggle-focused-tags`, `set-focused-tags` I would suggest command that operates on basically the same thing should follow this naming where I see first word as **action** and rest of it is context on which action is performed. But it should be verb, but i cant think of anything fitting.

I will change it to `focus-previous-tags` as you suggested and rename every "last" to "previous" since "last" kinda has this meaning of "there will be no more"

ifreund (approved):
LGTM, Thanks!

PR #359, opened 2021-07-15
river-layout: update to v3

ifreund:
- Remove advertise_view and advertise_done events. Using the information
provided by these for any purpose would make the layout far less
predictable. Futhermore, in the months this has been available for use,
to my knowledge nobody has actually used it for anything useful.

- Replace the set/mod layout value events with a single user_command
event. This simplifies the protocol and is more flexible for clients.

- Rename main factor to main ratio in rivertile. The "factor" name was
just legacy from dwm, "ratio" is much more accurate.


ifreund:
Reminder to self: need to update completions before merge.

novakne:
You forgot here https://github.com/ifreund/river/blob/004b185a6fc949a23a7c3c104b20fe95b7fd6791/rivertile/main.zig#L358

`fatal("wayland compositor does not support river-layout-v3.\n", .{});`

novakne:
Is it not possible to merge `"set-main-count", "mod-main-count", "set-main-ratio", "mod-main-ratio"`  in just two command?

ifreund:
> Is it not possible to merge `"set-main-count", "mod-main-count", "set-main-ratio", "mod-main-ratio"` in just two command?

Hmm, yeah on second thought I don't see much point in `set-main-count` and `set-main-ratio`. Maybe we should just get rid of them.

novakne:
Yes, IMO it would be more consistent with the new layout-v3 and I don't really see the point to have two command for basicaly the same thing. `mod` would makes sense because the user  always gonna modify the value, not really setting default, that should be in the layout client

Leon-Plickat:
You could also make it set the value if it is just a plain value, increase if it has the `+` prefix and descrease if it has the `-` suffix. That's probably how I'll do it in my layout.

protocol/river-layout-v3.xml
```diff
@@ -0,0 +1,178 @@
...
+      <description summary="commit a layout">
+        This request indicates that the client is done pushing dimensions
+        and the compositor may apply the layout. This completes the layout
+        demand with matching serial, any other requests sent with the serial
+        are a protocol error.
+
+        The compositor is free to use this proposed layout however it chooses,
+        including ignoring it.
+      </description>
+      <arg name="serial" type="uint" summary="serial of layout demand"/>
```
Leon-Plickat, line 162: I wonder if we should bring back layout names.

As a reminder, back in the earliest drafts we had a parameter here that was basically an arbitrary string passed back to river, with the intention that river would send it to status tracking clients. The idea was to allow status bars to display layout information, for example simple layout names like `[]=` in dwm, or more complex layout stats, like for example how many views are in a monocle layout.

That is certainly something I see people wanting to have.

ifreund: Hmm, this could also be added later in a non-breaking way as an optional request sent before the commit. Having it as part of the commit request would be nicer though.

Do you remember why we didn't end up including layout names in the protocol? I don't and #112 has over 100 comments.

Leon-Plickat: I removed it because river-options could be used for this instead.

ifreund: Ah yeah that makes sense. I think it's a very reasonable feature to have so I'll add it back in this PR and have a look at actually exposing it over river-status

ifreund:
> You could also make it set the value if it is just a plain value, increase if it has the `+` prefix and descrease if it has the `-` suffix. That's probably how I'll do it in my layout.

Nice idea, I've implemented this.

novakne:
Nice,  looks way better imo, a good cleanup for rivertile!

protocol/river-layout-v3.xml
ifreund: done :)

ifreund:
Thanks to both of you for the feedback, I think this is ready to merge now but I'll wait till after the weekend in case we think of anything else or someone shows up with a really awesome use-case for the `advertise_view` events which changes my mind.

nogerine:
Would it be possible to allow responding to `user_command`s? This can be used for error reporting and for requesting the current values.
For example, doing `riverctl send-layout-cmd rivertile main-ratio` would print the current main ratio. And if the command is malformed it would print a message saying so.

Leon-Plickat:
@nogerine what's the use case? The layout generator is the only program I can think off that actually has a use for those values.

Also a layout generator can already have error messages. Just send them to `stderr`. If your layout generator has error messages worth reading, you very likely only need to look for/at those when setting up river and the layout generator, not when running river day-to-day. For that case you can just run the layout generator in a terminal. If something is off in your current session, you can also kill the layout generator running in the background and restart it in a terminal to get the error messages.

Leon-Plickat:
However what may be an interesting idea is to have some sort of `ack_command` request to signal river whether the layout daemon has done anything with the command, so river knows whether a new layout demand is necessary.

Thoughts @ifreund ?

ifreund:
> Would it be possible to allow responding to `user_command`s? This can be used for error reporting and for requesting the current values.

I agree with @Leon-Plickat here that simply logging to stderr is the right way for layout generators to handle error reporting. I don't think a standardized way to query values should be necessary either because the values should be clearly visible in the resulting layout.

> However what may be an interesting idea is to have some sort of ack_command request to signal river whether the layout daemon has done anything with the command, so river knows whether a new layout demand is necessary.

This would mean an extra roundtrip when, for example, changing the main ratio with rivertile. I don't think this would be beneficial.

nogerine:
Error reporting was more of an afterthought, anyway. It would still be pretty neat if there was an ability to get the current values, that would simplify some of my scripts and make them less error-prone. But I guess it is more about getting the current state of river itself. I'll probably open a separate issue about that.

ifreund:
>  It would still be pretty neat if there was an ability to get the current values, that would simplify some of my scripts and make them less error-prone.

Do you have some specific use-cases you could share? I'm not opposed to such a feature in principle, I just don't want to add unnecessary features.

nogerine:
A very simple example with rivertile is the "next" main_location. Basically, a single button switches from left -> top -> left and so on. My exact use-case is a bit more complex, since it also handles some other layouts, but this is the main idea.

It is currently implemented using several files at a know location, but there is one problem with that: if somewhere else I forget to change these files when I change the main_location or the layout itself, it wouldn't work as expected. While it is easy to fix, it would be great if these kinds of issues were impossible in the first place.

Hmm... looking through my scripts, it seems this is the only one that uses the layout values specifically, but there are others that use river state.

Leon-Plickat:
> A very simple example with rivertile is the "next" main_location. Basically, a single button switches from left -> top -> left and so on. My exact use-case is a bit more complex, since it also handles some other layouts, but this is the main idea.

That is hacky. A better way too achieve this would be to patch rivertile (or write a new layout generator) to accept a `next_main_position` user command. The layout generator already knows the current main position, so this would be trivial.

ifreund:
Querying and then setting a some layout-specific value such as rivertile's main_location with two separate commands is also inefficient as it requires an extra roundtrip. I think storing the state yourself as you currently do or implementing a command in the layout generator to cycle through main locations would be a better way to do this.

PR #360, opened 2021-07-15
river: remove opacity command

ifreund:
This code is complex and increases maintenance burden but doesn't
add any functionality, only eye-candy.

Futhermore, neither I nor any of the core contributors use it.

There may be a place in river for such eye-candy down the line, in which
case this code could be revived. Currently river is early enough in its
development that our focus should be on core functionality instead.


ifreund:
Note: the maintainer of this eye-candy based river fork said that they plan to keep the feature around: <https://github.com/HeavyRain266/river>

HeavyRain266:
I totaly avree about removing this feature from river, keeping fancy compositing ecfects at this stage of development makes river harder to maintain and somebow it's in the way of simplicity and design goals of compositor, many people asked me about rounded corners and other stuff related to advancdd compositing which would require shaders. My fork was created because of all those questions about comositing effects from Pucom, "river is bad because doesn'r implemenr

PR #372, opened 2021-07-24
Add -version flag to river, riverctl and rivertile

novakne:


HeavyRain266:
What about changing it from only `-version` to `-v` and `--version` like with `-h` and `--help`?

novakne:
No, it's the other way around, `-h` and `--help` will become only `-help`

riverctl/main.zig
```diff
@@ -64,7 +64,15 @@ fn _main() !void {
     // Skip our name, send all other args
     // This next line is needed cause of https://github.com/ziglang/zig/issues/2622
     const args = os.argv;
-    for (args[1..]) |arg| control.addArgument(arg);
+    for (args[1..]) |arg| {
+        if (mem.eql(u8, mem.span(arg), "-version")) {
+            const stdout = std.io.getStdOut().writer();
+            try stdout.print("{s}\n", .{@import("build_options").version});
+            std.os.exit(0);
```
ifreund, line 71: I think we should put this check before the for loop and require `-version` to be the first argument. `riverctl send-layout-cmd -version` seems confusing.

river/main.zig
```diff
@@ -42,6 +42,7 @@ const usage: []const u8 =
     \\  -h            Print this help message and exit.
     \\  -c <command>  Run `sh -c <command>` on startup.
     \\  -l <level>    Set the log level to a value from 0 to 7.
+    \\  -version      Print river version.
```
ifreund, line 45:
```suggestion
    \\  -version      Print version number and exit.
```

rivertile/main.zig
```diff
@@ -52,6 +52,7 @@ const usage =
     \\Usage: rivertile [options]
     \\
     \\  -h, --help      Print this help message and exit.
+    \\  -version        Print river version.
```
ifreund, line 55:
```suggestion
    \\  -version        Print version number and exit.
```

build.zig
```diff
@@ -48,6 +48,21 @@ pub fn build(b: *zbs.Builder) !void {
 
     const examples = b.option(bool, "examples", "Set to true to build examples") orelse false;
 
+    const version = v: {
+        const version_string = "0.1.0";
+
+        var ret: u8 = undefined;
```
ifreund, line 54:
```suggestion
```

build.zig
```diff
@@ -48,6 +48,21 @@ pub fn build(b: *zbs.Builder) !void {
...
+        const git_dir = fs.path.join(b.allocator, &[_][]const u8{ b.build_root, "/.git" }) catch unreachable;
```
ifreund, line 55:
```suggestion
        const git_dir = fs.path.join(b.allocator, &[_][]const u8{ b.build_root, ".git" }) catch unreachable;
```

build.zig
```diff
@@ -48,6 +48,21 @@ pub fn build(b: *zbs.Builder) !void {
...
+        const git_commit_hash = b.execAllowFail(
+            &[_][]const u8{ "git", "--git-dir", git_dir, "--work-tree", b.build_root, "rev-parse", "--short", "HEAD" },
+            &ret,
+            .Ignore,
+        ) catch break :v version_string;
+
+        const dev_version = b.fmt("{s}-dev-{s}", .{ version_string, git_commit_hash });
```
ifreund, line 62: let's use `std.fmt.allocPrintZ()` instead here and get rid of the `dupeZ()` calls below.

build.zig
```diff
@@ -48,6 +48,21 @@ pub fn build(b: *zbs.Builder) !void {
...
+    const version = v: {
+        const version_string = "0.1.0";
+
+        var ret: u8 = undefined;
+        const git_dir = fs.path.join(b.allocator, &[_][]const u8{ b.build_root, "/.git" }) catch unreachable;
+        const git_commit_hash = b.execAllowFail(
+            &[_][]const u8{ "git", "--git-dir", git_dir, "--work-tree", b.build_root, "rev-parse", "--short", "HEAD" },
+            &ret,
+            .Ignore,
+        ) catch break :v version_string;
```
ifreund, line 60: I don't think we should return a bare version string without the `-dev` if git does not exist. Instead we should return `0.1.0-dev` currently as river `0.1.0` has not yet been released. Maybe we should do something like this:
1. Have a global `const version = "0.1.0-dev";` at the top of the build.zig
2. In the `build()` function, check if the `-dev` suffix is present. If it is, then try and obtain the git commit hash as you are currently doing. If not, simply use the version string as defined.
3. When river 0.1.0 is released, we set the global version const to `0.1.0` for the commit that gets tagged.
4. In the next commit after the 0.1.0 tag, we set the global version const to `0.2.0-dev`.

What do you think of that approach?

ifreund (changes requested):
Thanks for taking this on!

riverctl/main.zig
novakne: Yeah makes sense didn't though about this case

build.zig
novakne: hmm why? I need it in `execAllowFail()`, or how should I make it?

build.zig
novakne: Sounds good to me

> Have a global const version = "0.1.0-dev"; at the top of the build.zig

Should I use `river_version` so I can use `const version = :v {}` after?

ifreund: I'd use `version` for the global and do `const full_version = blk: {}` after.

build.zig
ifreund: Ah, I missed that this was passed to `execAllowFail()` and thought it was just unused.

novakne:
I tried with version set to `0.1.0` and `0.1.0-dev` and it works as intended

build.zig
```diff
@@ -48,6 +50,22 @@ pub fn build(b: *zbs.Builder) !void {
 
     const examples = b.option(bool, "examples", "Set to true to build examples") orelse false;
 
+    const full_version = blk: {
+        if (mem.endsWith(u8, version, "-dev")) {
+            var ret: u8 = undefined;
+            const git_dir = fs.path.join(b.allocator, &[_][]const u8{ b.build_root, ".git" }) catch unreachable;
```
ifreund, line 56:
```suggestion
            const git_dir = try fs.path.join(b.allocator, &[_][]const u8{ b.build_root, ".git" });
```

build.zig
```diff
@@ -48,6 +50,22 @@ pub fn build(b: *zbs.Builder) !void {
...
+            const git_commit_hash = b.execAllowFail(
+                &[_][]const u8{ "git", "--git-dir", git_dir, "--work-tree", b.build_root, "rev-parse", "--short", "HEAD" },
+                &ret,
+                .Ignore,
+            ) catch break :blk version;
+
+            const dev_version = std.fmt.allocPrintZ(b.allocator, "{s}-{s}", .{ version, git_commit_hash }) catch return;
+            break :blk dev_version;
```
ifreund, line 64:
```suggestion
            break :blk try std.fmt.allocPrintZ(b.allocator, "{s}-{s}", .{ version, git_commit_hash });
```

build.zig
```diff
@@ -48,6 +50,22 @@ pub fn build(b: *zbs.Builder) !void {
...
     const examples = b.option(bool, "examples", "Set to true to build examples") orelse false;
 
+    const full_version = blk: {
+        if (mem.endsWith(u8, version, "-dev")) {
+            var ret: u8 = undefined;
+            const git_dir = fs.path.join(b.allocator, &[_][]const u8{ b.build_root, ".git" }) catch unreachable;
+            const git_commit_hash = b.execAllowFail(
+                &[_][]const u8{ "git", "--git-dir", git_dir, "--work-tree", b.build_root, "rev-parse", "--short", "HEAD" },
+                &ret,
+                .Ignore,
```
ifreund, line 60: I think this is more correct so that anything `git` logs to stderr will still be printed.
```suggestion
                .Inherit,
```

riverctl/main.zig
```diff
@@ -64,6 +64,13 @@ fn _main() !void {
     // Skip our name, send all other args
     // This next line is needed cause of https://github.com/ziglang/zig/issues/2622
```
ifreund, line 65: These comments need to be moved/updated due to your change.

build.zig
```diff
@@ -4,6 +4,8 @@ const fs = std.fs;
 const mem = std.mem;
 const zbs = std.build;
 
+const version = "0.1.0-dev";
```
ifreund, line 7: This could use a doc comment explaining when/how to change this when doing a release.

ifreund (approved):
Thanks!

PR #376, opened 2021-07-26
Fix typo in rivertile doc/manual

paemuri:
Basically, renames the duplicated `main-count` to the correct `main-ratio`, according to the command's description.


ifreund (approved):
Thanks!

PR #378, opened 2021-07-28
build: Use pkg-config to handle river protocols

novakne:
- Install river-layout-v3 in `$prefix/share/river-protocols/`
- Create `river-protocols.pc` in `zig-cache/` and install it in `$prefix/share/pkgconfig/`

Now for example in a 3rd party program you could use:
```zig
const std = @import("std");
const Builder = @import("std").build.Builder;

const ScanProtocolsStep = @import("deps/zig-wayland/build.zig").ScanProtocolsStep;

pub fn build(b: *Builder) !void {
    [...]
    
    const protocol_path = std.mem.trim(u8, b.exec(
        &[_][]const u8{ "pkg-config", "--variable=pkgdatadir", "river-protocols" },
    ) catch unreachable, &std.ascii.spaces);
    const layout_protocol = try std.fs.path.join(b.allocator, &[_][]const u8{protocol_path, "river-layout-v3.xml"});
    scanner.addProtocolPath(layout_protocol);
    
    [...]
}
```

I'm not sure about the version @ifreund, so tell me if you want something else

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -124,7 +124,26 @@ pub fn build(b: *zbs.Builder) !void {
         rivertile.install();
     }
 
-    b.installFile("protocol/river-layout-v3.xml", "share/river/river-layout-v3.xml");
+    {
+        const file = try fs.path.join(b.allocator, &[_][]const u8{ b.cache_root, "river-protocols.pc" });
+        const pkgconfig_file = try std.fs.cwd().createFile(file, .{});
+
+        const writer = pkgconfig_file.writer();
+        try writer.print(
+            \\prefix={s}
+            \\datadir=${{prefix}}/share
+            \\pkgdatadir=${{datadir}}/river-protocols
+            \\
+            \\Name: river-protocols
+            \\URL: https://github.com/ifreund/river
+            \\Description: protocol files for the river wayland compositor
+            \\Version: {s}
+        , .{ b.install_prefix, full_version });
+        defer pkgconfig_file.close();
+
+        b.installFile("protocol/river-layout-v3.xml", "share/river-protocols/river-layout-v3.xml");
+        b.installFile(file, "share/pkgconfig/river-protocols.pc");
+    }
 
     if (man_pages) {
         const scdoc_step = ScdocStep.create(b);


novakne:
I just thought about this, maybe you want to use stable/unstable/staging folder?

build.zig
```diff
@@ -156,6 +166,19 @@ pub fn build(b: *zbs.Builder) !void {
...
+fn writePkgconfigFile(writer: anytype, prefix: []const u8) !void {
+    try writer.print(
+        \\prefix={s}
+        \\datarootdir=${{prefix}}/share
+        \\pkgdatadir=${{datarootdir}}/river-protocols
+        \\
+        \\Name: River Protocols
+        \\URL: https://github.com/ifreund/river
+        \\Description: River protocol files
+        \\Version: 0.1.0
```
ifreund, line 178: why don't we use the `full_version` here?

build.zig
```diff
@@ -156,6 +166,19 @@ pub fn build(b: *zbs.Builder) !void {
     }
 }
 
+fn writePkgconfigFile(writer: anytype, prefix: []const u8) !void {
+    try writer.print(
+        \\prefix={s}
+        \\datarootdir=${{prefix}}/share
+        \\pkgdatadir=${{datarootdir}}/river-protocols
```
ifreund, line 173: the wayland protocols pkg config file does
```
pkgdatadir=${pc_sysrootdir}${datarootdir}/wayland-protocols
```
What's the difference here?

build.zig
```diff
@@ -124,7 +124,17 @@ pub fn build(b: *zbs.Builder) !void {
...
+    {
+        const file = try fs.path.join(b.allocator, &[_][]const u8{ b.cache_root, "river-protocols.pc" });
+        const pkgconfig_file = try std.fs.cwd().createFile(file, .{});
+
+        const writer = pkgconfig_file.writer();
+        try writePkgconfigFile(writer, b.install_prefix);
+        defer pkgconfig_file.close();
+
+        b.installFile("protocol/river-layout-v3.xml", "share/river-protocols/river-layout-v3.xml");
+        b.installFile(file, "lib/pkgconfig/river-protocols.pc");
```
ifreund, line 136: wayland-protocols puts this in `$PREFIX/share/pkgconfig/`, maybe we should do the same?

ifreund:
> I just thought about this, maybe you want to use stable/unstable/staging folder?

Sure, let's follow `wayland-protocols` and put `river-layout-v3` in staging for now. Then protocol stability is de-coupled from river's releases.

build.zig
novakne: Oh right of course, I hadn't thought about it

build.zig
novakne: Wayland protocols use some variables from meson see https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/meson.build#L88 and also see this commit message https://gitlab.freedesktop.org/wayland/wayland-protocols/-/commit/1ad48a50a45b4a4dbc84d9e85ea49311bb864629

So to be honest  I'm not sure if we need an hack like this or not, pkg-config documentation use variable like I did

build.zig
novakne: pkg-config man page say they are both correct, I personally have `~/.local/lib/pkgconfig` in my $PKG_CONFIG_PATH` so that's why and looks like most packages I have use `lib`

build.zig
ifreund: I was more talking about the `pc_sysrootdir` present in the wayland protocols .pc file but not here. If I understand correctly, that's required for `PKG_CONFIG_SYSROOT_DIR` to work. It seems to me that wayland-protocols does this wrong though and it should be used to define the `prefix` variable with
```
prefix=${{pc_sysrootdir}}/{s}
```

We don't need the `datarootdir` hack from wayland-protocols as we don't and never will use autotools, let's switch to the more standard `datadir`.

build.zig
ifreund: I suspect most packages use `lib` because they are using their `.pc` file to give information about files they have installed in `lib` while `wayland-protocols` and we don't install anything to `lib`, only to `share`.

novakne: From the man page "the default is `libdir/pkgconfig:datadir/pkgconfig`" so maybe they just put it in the first one, anyway if you prefer share I'll use share, no problem

build.zig
novakne: Hmm yeah you're right I misunderstood that, so do we use
```
        \\prefix=${{pc_sysrootdir}}/{s}
        \\datadir=${{prefix}}/share
        \\pkgdatadir=${{datadir}}/river-protocols
```
or  just
```
        \\prefix=${{pc_sysrootdir}}/{s}
        \\datadir=${{prefix}}/share/river-protocols
```

ifreund: You know what, grepping around my system almost nothing seems to use `pc_sysrootdir`. Let's just leave it out unless someone complains. I think this should be good:
```
        \\prefix={s}
        \\datadir=${{prefix}}/share
        \\pkgdatadir=${{datadir}}/river-protocols

```

build.zig
ifreund: cool, let's go with share then.

build.zig
novakne: Same, only python use it IIRC

novakne:
Should be all good

novakne:
@ifreund btw do the example I wrote in the first post looks good to you or is there a better way? That's what I tried with rivercarro. It works but maybe it could be better

build.zig
```diff
@@ -124,7 +124,17 @@ pub fn build(b: *zbs.Builder) !void {
...
-    b.installFile("protocol/river-layout-v3.xml", "share/river/river-layout-v3.xml");
+    {
+        const file = try fs.path.join(b.allocator, &[_][]const u8{ b.cache_root, "river-protocols.pc" });
+        const pkgconfig_file = try std.fs.cwd().createFile(file, .{});
+
+        const writer = pkgconfig_file.writer();
+        try writePkgconfigFile(writer, b.install_prefix, full_version);
+        defer pkgconfig_file.close();
+
+        b.installFile("protocol/river-layout-v3.xml", "share/river-protocols/staging/river-layout-v3.xml");
```
ifreund, line 135: After thinking about this a bit more, I don't think I want to make the distinction between stable/staging protocols. I never intend to say that river will avoid all breaking changes until the end of time, and the same goes for river's protocols. Things will get highly stable eventually, but set in stone with no chance to fix mistakes.

build.zig
```diff
@@ -156,6 +166,19 @@ pub fn build(b: *zbs.Builder) !void {
     }
 }
 
+fn writePkgconfigFile(writer: anytype, prefix: []const u8, v: []const u8) !void {
+    try writer.print(
+        \\prefix={s}
+        \\datadir=${{prefix}}/share
+        \\pkgdatadir=${{datadir}}/river-protocols
+        \\
+        \\Name: River Protocols
```
ifreund, line 175:
```diff
-        \\Name: River Protocols
+        \\Name: river-protocols
```

build.zig
```diff
@@ -156,6 +166,19 @@ pub fn build(b: *zbs.Builder) !void {
...
+        \\URL: https://github.com/ifreund/river
+        \\Description: River protocol files
```
ifreund, line 177:
```diff
-         \\Description: River protocol files
+         \\Description: protocol files for the river wayland compositor
```

build.zig
```diff
@@ -156,6 +166,19 @@ pub fn build(b: *zbs.Builder) !void {
     }
 }
 
+fn writePkgconfigFile(writer: anytype, prefix: []const u8, v: []const u8) !void {
```
ifreund, line 169: I don't think splitting this out into a function really makes sense, it's only used in one place.

ifreund (commented):
Cool, just a couple of minor tweaks then this is good to go!

build.zig
novakne: So you want a `stable` folder or just no folder, everything in river-protocols?

ifreund: Just no folder, everything in `share/river-protocols`

build.zig
novakne: Yes you're right, it was because my first version when I was testing was different and need a function but now it doesn't makes sense

ifreund:
> @ifreund btw do the example I wrote in the first post looks good to you or is there a better way? That's what I tried with rivercarro. It works but maybe it could be better

That seems fine, though you could use `try b.exec()` instead of `catch unreachable`.

novakne:
Allright clean and ready IMO

> That seems fine, though you could use `try b.exec()` instead of `catch unreachable`.
> 
> I think the build system will need some kind of shell like dsl to make stuff like this easier at some point.

Thank you! It would be great yeah

ifreund (approved):
LGTM, Thanks!

PR #379, opened 2021-07-28
ci: add musl runner

HeavyRain266:
I'm opening this as Draft because GitHub has problems with checkout on containers with musl libc as for now (same command works locally but fails here) also aready opened issue about this in [checkout](https://github.com/actions/checkout/issues/560) repo. All what I did was copying base runner and changing container name to `voidlinux/voidlinux-msul` and adding `musl` to name of every job.

This PR partially closes this [340](https://github.com/ifreund/river/issues/340), there is no easy way to run FreeBSD in GitHub Actions, I'd tried using docker container and vm as `macos-latest` (only macOS has VMs as CI). I think easiest way would be copy of river on SourceHut because they are having FreeBSD as native system for CIs, thats the same way which Zig itself uses for tests on FreeBSD.

I'm FreeBSD user and can confirm that after switch to Zig 0.8 didn't had to add any patches to even compile it, also ports tree auto build and deploy latest commit and there is info if fails or not.

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -22,12 +22,6 @@ jobs:
         run: |
           # wlroots dependencies as well as we build from source
           sh ~/river/.github/workflows/install_deps.sh
-
-          git clone https://github.com/swaywm/wlroots.git
-          cd wlroots
-          git checkout 0.14.0
-          meson build --auto-features=enabled -Dexamples=false -Dwerror=false -Db_ndebug=false
-          ninja -C build install
       - name: checkout
         uses: actions/checkout@v2
         with:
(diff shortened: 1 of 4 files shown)


ifreund:
> All what I did was copying base runner and changing container name to `voidlinux/voidlinux-msul` and adding `musl` to ma,e of every job.

It would be nice if we could factor the common code into an `install_deps.sh` script or similar.

> I think easiest way would be copy of river on SourceHut because they are having FreeBSD as native system for CIs, thats the same way which Zig itself uses for tests on FreeBSD.

Yes, that would be fine. I believe @Leon-Plickat mentioned that he'd be willing to let river use his sr.ht account for such a CI runner. I'm not sure what setting that up involves, but if you or someone else were to PR whatever config/script would be needed then I'm sure we could figure it out from there.

HeavyRain266:
> It would be nice if we could factor the common code into an `install_deps.sh` script or similar.

Done, tested script in local docker container and works fine.

ifreund (commented):
Can't the shell script compile wlroots as well? Shouldn't we use the script for the xwayland builds as well?

HeavyRain266:
> Can't the shell script compile wlroots as well? Shouldn't we use the script for the xwayland builds as well?

I had it added to xwayland, but probably copied older version before push, sorry I will add wlroots compilation to script as well...

Leon-Plickat:
> I believe @Leon-Plickat mentioned that he'd be willing to let river use his sr.ht account for such a CI runner.

Yes, since I have access to that service, we may as well use it.

If anyone wants to figure out how to hook up river to that, I believe [this](https://man.sr.ht/dispatch.sr.ht/github.md) will be helpful.

HeavyRain266:
Hmmm... checks with script fails because cannot execute it for some reason, tested it in docker before and after push in same container as here and was working. Mayby should I pass chmod +x before executing it?

ifreund:
> Hmmm... checks with script fails because cannot execute it for some reason, tested it in docker before and after push in same container as here and was working. Mayby should I pass chmod +x before executing it?

At first glance it looks like the path is wrong. Maybe get rid of the `~/river` and just do `./.github/install_deps.sh`? If you committed the file with the correct unix permissions git should preserve them.

HeavyRain266:
> At first glance it looks like the path is wrong. Maybe get rid of the `~/river` and just do `./.github/install_deps.sh`? If you committed the file with the correct unix permissions git should preserve them.

I will simply run `ls` command on another branch to get file structure inside actions container.

HeavyRain266:
Looks like they're not going to do something about it, it would be better idea to use sourcehut dispatch for musl too.

PR #384, opened 2021-08-05
Add 'main-cols' to rivertile

minego:
I am hoping to daily drive river soon, but unfortunately I use a super wide monitor (32:9 aspect ratio) and for that to be usable I really want the ability to have multiple main columns.

This PR adds a 'main-cols' count. If you leave this as 1 then the existing behavior is unchanged but raising it allows much more control over the layout.

### DIFF
--- a/rivertile/main.zig
+++ b/rivertile/main.zig
@@ -251,7 +251,7 @@ const Output = struct {
                 // view slightly larger if the height is not evenly divisible
                 var main_width: u32 = undefined;
                 var main_height: u32 = undefined;
-                var main_count_per_col: u32 = undefined;
+                var main_rows_per_col: u32 = undefined;
 
                 var secondary_width: u32 = undefined;
                 var secondary_height: u32 = undefined;
@@ -279,7 +279,7 @@ const Output = struct {
                 }
 
                 if (main_count > 0) {
-                    main_count_per_col = main_count / main_cols;
+                    main_rows_per_col = main_count / main_cols;
                 }
 
                 var i: u32 = 0;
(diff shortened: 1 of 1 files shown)


rivertile/main.zig
```diff
@@ -60,6 +60,9 @@ const usage =
     \\                  layout. (Default left)
     \\  -main-count     Set the initial number of views in the main area of the
     \\                  layout. (Default 1)
+    \\  -main-cols      Set the initial number of columns to spread the main
+    \\                  area of the layout over. (Default 1)
+    \\                  layout. (Default 1)
```
novakne, line 65:
```suggestion
```

novakne (commented):
I don't know if ifreund want this, but you should update the rivertile man page too.

Also it would be better if when you open a PR you don't use your master branch, it makes it difficult to fetch your PR to test it locally

ifreund:
@minego Hey, thanks for your interest in river and for the patch. However, I consider `rivertile` to be feature complete at this point and don't intend to accept any more features for it, including the proposed `main-cols`, as the main point of having the river-layout protocol is so that everyone can have exactly the layout client they want without me needing to maintain all the code in this repository.

I'd encourage you to simply fork rivertile with these changes, for example as @novakne has already done with [rivercarro](https://git.sr.ht/~novakane/rivercarro). There are a few other layout clients listed on [this wiki page](https://github.com/ifreund/river/wiki/Community-Layouts) as well.

> Also it would be better if when you open a PR you don't use your master branch, it makes it difficult to fetch your PR to test it locally

@novakne you can do e.g. `git fetch upstream pull/384/head:local-branch-name`.

minego:
No problem, I can understand that. Thanks.

novakne:
@ifreund oh I always use this command but always though the branch name after `head` should be the one from the PR, well TIL, thanks

@minego As ifreund said I have a fork of rivertile and it looks like a fair request to me, so if you don't want to make your own fork or use an other layout generator from the wiki, feel free to send a patch for rivercarro and I'll look into it.

minego:
Thanks, I may do that. I'm also considering just writing my own I'm C, just because I'm much more comfortable in C and long term I know I will want to continue to tinker. I'm a dwm user currently, and continual tinkering is a big thing :-)

Mostly this has been an attempt for me to learn and prepare for my switch to Wayland. One big issue I still have is that I'm used to having a special column in my layout that keeps my chat/mail apps. It is a bit of a weird hacky case, and I don't see a way to do anything similar with river. The layout generator just doesn't have enough information about the clients to do that kind of thing.

Once I find a good way to keep my slack window in the same spot most of the time (and to do screen sharing with zoom, ugh) then I should be ready to actually switch.

Issue #375, opened 2021-07-26
Border not updated on tag toggle (qt)


elshize:
## Disclaimer

I'm running nvidia with open source drivers, so that might be a factor.

# Steps to reproduce

Open a Qt5 app (I tried qutebrowser, Discover, and Dolphin) on tag 1, and another app (doesn't matter what) on tag 2.

1. Start with tag 1 activated.
2. Toggle tag 2 so that both 1 and 2 are activated.
3. Toggle back tag 1 so that only 2 is activated (doesn't matter if you toggle 1 or set focused to 2, works the same either way).
4. Set focused tag to 1.

# Outcome

![image](https://user-images.githubusercontent.com/3026120/127056378-ff228cfa-bb30-48fb-987d-1fff8ef1dcb8.png)

elshize:
I can actually confirm that the same thing is happening on my lemur laptop with intel gpu, so must be something else. Let me know if you have problems reproducing, I'll try to dig in to see what could be the common denominator between the two. Both are fedora 34 installations, for one, but not sure if that could play a role.

Leon-Plickat:
I wonder what backend plugin Qt is running here. Did you manually set `QT_QPA_PLATFORM` to anything?

HeavyRain266:
Qutebrowser and other Qt apps works natively if you pass env variables to enable Wayland support, bugged border isn't river's problem but Qt apps running under X11/xwayland are broken outside KDE and have general problems with decorations provided by compositors and window managers like borders in this case.

There is env variables to remove CSD from all Qt programs and should help with this problem as well as setting wayland as backend.

ifreund:
> There is env variables to remove CSD from all Qt programs and should help with this problem as well as setting wayland as backend.

This doesn't seem to be the case, I can reproduce with `QT_QPA_PLATFORM=wayland qutebrowser`

elshize:
@ifreund tried setting the variable as well, and also still having the issue.

elshize:
I also tried `QT_WAYLAND_DISABLE_WINDOWDECORATION=1` but same problem.

ifreund:
So after some investigation it looks like this is a QT bug. Here's the relevant `WAYLAND_DEBUG=1` snippet from `qutebrowser`:
```
[1860693.742] xdg_toplevel@26.configure(1019, 970, array)
[1860693.750] xdg_surface@25.configure(82)
[1860693.754] wl_callback@20.done(230034835)
[...]

PR #387, opened 2021-08-06
render: draw borders before rendering views

### LINKED ISSUE #382: Borders are rendered after XDG Popups
As a results, large popup menus on small views look weird.

20210805_14h35m15s_grim

keithhub:
Menus, tooltips, etc. can extend beyond a view's borders.  Render views after their borders so floating content appears on top.

Unfocused floating content can still be obscured by views higher in the stack and the focused view.

Fixes #382.

### DIFF
--- a/river/render.zig
+++ b/river/render.zig
@@ -93,32 +93,32 @@ pub fn renderOutput(output: *Output) void {
         it = ViewStack(View).iter(output.views.last, .reverse, output.current.tags, renderFilter);
         while (it.next()) |view| {
             if (view.current.focus != 0 or view.current.float) continue;
-            renderView(output, view, &now);
             if (view.draw_borders) renderBorders(output, view, &now);
+            renderView(output, view, &now);
         }
 
         // focused, non-floating views
         it = ViewStack(View).iter(output.views.last, .reverse, output.current.tags, renderFilter);
         while (it.next()) |view| {
             if (view.current.focus == 0 or view.current.float) continue;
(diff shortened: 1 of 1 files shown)


novakne:
It works as intended for me, although I don't know if there is a downside elsewhere to changing the orders like this.

ifreund (commented):
I think the ideal rendering order would be xdg toplevels, then borders drawn by river, then xdg popups. With the current patch, if the contents of the client's toplevel surfaces extend past the geometry reported by the client with `xdg_surface.set_window_geometry` these surface contents will be drawn over the borders.

On the other hand though, I don't think that situation should ever happen with a well behaving client when using server-side decorations and setting the xdg_toplevel state to be tiled on all sides. Not all clients are well-behaved though sadly: #375.

I'll give some thought as to whether or not this should be merged as is or if we should implement the more complex rendering order I described above.

novakne:
> On the other hand though, I don't think that situation should ever happen with a well behaving client when using server-side decorations and setting the xdg_toplevel state to be tiled on all sides. Not all clients are well-behaved though sadly: #375.

Does it really makes sense to adapt something because of bug in external clients? I guess if it really introduce a lot of problems for river users you could think about it but it would be better to do it the right way in river an expect fixes upstream

ifreund (approved):
Let's switch to this order and see if anyone complains about e.g. CSD shadows being drawn above river's borders in practice.

Thanks for the PR!

PR #388, opened 2021-08-06
Decrease timeout

Leon-Plickat:


river/LayoutDemand.zig
```diff
@@ -38,7 +38,7 @@ const log = std.log.scoped(.layout);
 
 const Error = error{ViewDimensionMismatch};
 
-const timeout_ms = 1000;
+const timeout_ms = 200;
```
ifreund, line 41: IIRC it was found in #394 that 200 was more than enough, and we could probably drop this as low as 50ms without causing issues. However, I'm not sure dropping it so low really gains us anything. The only bad effect of a layout generator timing is that river will appear to freeze rendering until the timeout expires. I think if someone writes a really slow layout generator we shouldn't totally prevent them from using it, that's their choice.

Let's just set this to 100ms and call it a day.

Leon-Plickat: done

ifreund (approved):
LGTM, Thanks!

PR #392, opened 2021-08-09
Add 'reset-config' command.

Leon-Plickat:
As discussed some time ago.

The command creates a new empty config and also resets all configuration of seats and outputs.

Not changed is input configuration, which I will add that in a future patch.

Also included is a small bug-fix that prevents the `output-layout` from leaking the previous allocated namespace.


river/command/reset_config.zig
```diff
@@ -0,0 +1,66 @@
...
+    var new_config = try Config.init();
+    server.config.deinit();
+    server.config = new_config;
+
+    // Return all seats to normal mode.
+    {
+        var it = server.input_manager.seats.first;
+        while (it) |node| : (it = node.next) {
+            node.data.mode_id = 0;
+            node.data.prev_mode_id = 0;
```
ifreund, line 46: What if we are currently in locked mode due to an active input inhibitor?

Leon-Plickat: Ah, I forgot that.

ifreund (changes requested):
I think we need to clear any currently repeating mappings as well now that #389 has been merged.

ifreund:
Closing as stale, feel free to reopen or create a new PR if you like.

I pushed a fix for the mentioned leak in command/layout.zig to master.

PR #394, opened 2021-08-10
contrib: add python layout

Leon-Plickat:


contrib/layout.py
```diff
@@ -0,0 +1,115 @@
...
+    layout.commit(str(view_count) + " windows layout out by python", serial)
+
+class Output(object):
+    def __init__(self):
+        self.output = None
+        self.layout = None
+        self.id = None
+
+    def destroy(self):
+        if self.layout != None:
```
elshize, line 59: Although it might work most of the time, it's a bad practice to use `==` checking if something's `None`. You should use `self.layout is None` (or `is not None` in some cases below). The reason for that is `==` can be technically overridden for some types and not behave the way you expect, while `is None` always works as expected.

Note there are several of these comparisons below.

ifreund (commented):
This looks pretty neat, could I get a TLDR on how to get pywayland to scan the river-layout protocol so I can test this? Probably would be worth adding that to the comment at the top as well, similar to the build instructions for `contrib/layout.c`.

Leon-Plickat:
TBH I have no idea how pywaylands scanner is supposed to work.

I got it to scan the protocol and generate bindings by copying it to `/usr/share/wayland-protocols/river-layout-v3/river-layout-v3.xml` and running this command once:

```sh
sudo python3 -m pywayland.scanner --with-protocols 
```

Leon-Plickat:
@ifreund I have found out how to make the scanning work.

```sh
mkdir protocols
python -m pywayland.scanner -o protocols -i /usr/share/wayland/wayland.xml ../protocols/river-layout-v3.xml
```

The scanner finds `wayland.xml` automatically by default, but when using `-i` you do not add *additional* protocol files to scan, you bypass the automatic path finding. So `wayland.xml` apparently needs to be specified explicitly. Pretty unintuitive, but it works.

Now I only need to figure out how to make python import the local bindings.

Leon-Plickat:
Ok, I give up. I don't think pywayland supports using local bindings. It seems like all bindings have to be part of the actual package, which means the only two options I see are either having a local copy of pywayland with the bindings, or adding the bindings to the global pywayland package.

The command to add the bindings to the global package:
```sh
sudo python3 -m pywayland.scanner --with-protocols -i /usr/share/wayland/wayland.xml river-layout-v3.xml
```

contrib/layout.py
```diff
@@ -0,0 +1,139 @@
+#!/bin/env python3
```
jbeich, line 1: `/usr/bin/env` is more likely to be available even on distros that don't symlink `/usr` into `/`.
```suggestion
#!/usr/bin/env python3
```

Leon-Plickat:
Are there still any blockers, other than pywayland being unwieldy?

ifreund:
> Are there still any blockers, other than pywayland being unwieldy?

Nope, I just need to find time to deal with setting up pywayland to test.

ifreund (approved):
LGTM, Thanks!

ifreund:
Note: I found that one can avoid running the pywayland scanner as root by using a virtual environment, which is a lot cleaner. I added a commit recommending this instead.

For example:
```
cd contrib
python -m venv fooenv
source ./fooenv/bin/activate
pip install pywayland
python -m pywayland.scanner -i /usr/share/wayland/wayland.xml ../protocol/river-layout-v3.xml
python layout.py
```

PR #395, opened 2021-08-12
command: Use std.ComptimeStringMap to get impl. function from command string

Leon-Plickat:
The `// TODO` told me to do it.

Looks cleaner, but for the low amount of strings the performance improvement is probably not measurable.

### DIFF
--- a/river/command.zig
+++ b/river/command.zig
@@ -36,59 +36,58 @@ pub const Orientation = enum {
     vertical,
 };
 
-// TODO: this could be replaced with a comptime hashmap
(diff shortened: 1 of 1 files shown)


ifreund (approved):
LGTM, just needs conflicts resolved.

ifreund:
And yeah, this is more about cleanliness than performance.

Leon-Plickat:
Rebased against master, should be good now.

ifreund (approved):
LGTM, Thanks!

PR #397, opened 2021-08-12
Implement xdg_activation_v1

novakne:
Currently doesn't do anything with urgent views since we need to discuss what to do

I updated zig-wlroots to the latest commit, maybe @ifreund you prefer just to the commit that add xdg_activation_v1 binds?

### DIFF
--- a/protocol/river-status-unstable-v1.xml
+++ b/protocol/river-status-unstable-v1.xml
@@ -16,7 +16,7 @@
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
   </copyright>
 
-  <interface name="zriver_status_manager_v1" version="1">
+  <interface name="zriver_status_manager_v1" version="2">
     <description summary="manage river status objects">
       A global factory for objects that receive status information specific
       to river. It could be used to implement, for example, a status bar.
@@ -47,7 +47,7 @@
     </request>
   </interface>
 
-  <interface name="zriver_output_status_v1" version="1">
+  <interface name="zriver_output_status_v1" version="2">
     <description summary="track output tags and focus">
       This interface allows clients to receive information about the current
       windowing state of an output.
@@ -74,6 +74,14 @@
         of the output changes.
       </description>
       <arg name="tags" type="array" summary="array of 32-bit bitfields"/>
+    </event>
+
+    <event name="urgent_tags" since="2">
+      <description summary="tags of the output with an urgent view">
+        Sent once on binding the interface and again whenever the set of
+        tags with at least one urgent view changes.
+      </description>
+      <arg name="tags" type="uint" summary="32-bit bitfield"/>
     </event>
   </interface>
 
--- a/river/OutputStatus.zig
+++ b/river/OutputStatus.zig
@@ -38,9 +38,16 @@ pub fn init(self: *Self, output: *Output, output_status: *zriver.OutputStatusV1)
 
     output_status.setHandler(*Self, handleRequest, handleDestroy, self);
 
-    // Send view/focused tags once on bind.
+    // Send view/focused/urgent tags once on bind.
     self.sendViewTags();
     self.sendFocusedTags(output.current.tags);
+
+    var urgent_tags: u32 = 0;
+    var view_it = self.output.views.first;
+    while (view_it) |node| : (view_it = node.next) {
+        if (node.view.current.urgent) urgent_tags |= node.view.current.tags;
+    }
+    self.sendUrgentTags(urgent_tags);
 }
 
 pub fn destroy(self: *Self) void {
@@ -81,4 +88,10 @@ pub fn sendViewTags(self: Self) void {
 
 pub fn sendFocusedTags(self: Self, tags: u32) void {
     self.output_status.sendFocusedTags(tags);
+}
+
+pub fn sendUrgentTags(self: Self, tags: u32) void {
+    if (self.output_status.getVersion() >= 2) {
+        self.output_status.sendUrgentTags(tags);
+    }
 }
--- a/river/Root.zig
+++ b/river/Root.zig
@@ -379,6 +379,7 @@ fn commitTransaction(self: *Self) void {
         output.current = output.pending;
 
         var view_tags_changed = false;
+        var urgent_tags_dirty = false;
 
         var view_it = output.views.first;
         while (view_it) |view_node| {
@@ -395,12 +396,15 @@ fn commitTransaction(self: *Self) void {
             // Apply pending state of the view
             view.pending_serial = null;
             if (view.pending.tags != view.current.tags) view_tags_changed = true;
+            if (view.pending.urgent != view.current.urgent) urgent_tags_dirty = true;
+            if (view.pending.urgent and view_tags_changed) urgent_tags_dirty = true;
             view.current = view.pending;
 
             view.dropSavedBuffers();
         }
 
         if (view_tags_changed) output.sendViewTags();
+        if (urgent_tags_dirty) output.sendUrgentTags();
 
         output.damage.addWhole();
     }
--- a/river/Seat.zig
+++ b/river/Seat.zig
@@ -218,6 +218,7 @@ pub fn setFocusRaw(self: *Self, new_focus: FocusTarget) void {
                 std.debug.assert(self.focused_output == target_view.output);
                 if (target_view.pending.focus == 0) target_view.setActivated(true);
                 target_view.pending.focus += 1;
+                target_view.pending.urgent = false;
             },
             .layer => |target_layer| std.debug.assert(self.focused_output == target_layer.output),
             .none => {},
--- a/river/View.zig
+++ b/river/View.zig
@@ -71,6 +71,7 @@ const State = struct {
 
     float: bool = false,
     fullscreen: bool = false,
+    urgent: bool = false,
 };
 
 const SavedBuffer = struct {
@@ -130,6 +131,9 @@ foreign_fullscreen: wl.Listener(*wlr.ForeignToplevelHandleV1.event.Fullscreen) =
 foreign_close: wl.Listener(*wlr.ForeignToplevelHandleV1) =
     wl.Listener(*wlr.ForeignToplevelHandleV1).init(handleForeignClose),
 
+request_activate: wl.Listener(*wlr.XdgActivationV1.event.RequestActivate) =
+    wl.Listener(*wlr.XdgActivationV1.event.RequestActivate).init(handleRequestActivate),
+
 pub fn init(self: *Self, output: *Output, tags: u32, surface: anytype) void {
     self.* = .{
         .output = output,
@@ -163,6 +167,8 @@ pub fn destroy(self: *Self) void {
         .xdg_toplevel => |*xdg_toplevel| xdg_toplevel.deinit(),
         .xwayland_view => |*xwayland_view| xwayland_view.deinit(),
     }
+
+    self.request_activate.link.remove();
 
     const node = @fieldParentPtr(ViewStack(Self).Node, "view", self);
     self.output.views.remove(node);
@@ -276,6 +282,11 @@ pub fn sendToOutput(self: *Self, destination_output: *Output) void {
 
     self.output.sendViewTags();
     destination_output.sendViewTags();
+
+    if (self.pending.urgent) {
+        self.output.sendUrgentTags();
+        destination_output.sendUrgentTags();
+    }
 
     if (self.surface) |surface| {
         surface.sendLeave(self.output.wlr_output);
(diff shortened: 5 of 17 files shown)


novakne:
Added a noop `requestActicate()` where we can implement what we want.

I tried to make a `border-color-urgent` that change the color of urgent view, works fine. Added a urgent bool in view state, and then change the color in `renderBorder()` based on this.

So what do we do?
- Change the border color of urgent view
- Focus the view, probably with an config option to turn it off
- Marked the tags as urgent, not sure how to do this and  if it will be useful on river

I am personally mostly interested in the color borders change

ifreund:
> I updated zig-wlroots to the latest commit, maybe @ifreund you prefer just to the commit that add xdg_activation_v1 binds?

The latest commit is fine



> So what do we do?

I'll just paste my answer on IRC here for the github audience: 
```
10:15 <novakane> ifreund: any idea on what you want for urgent views?
10:25 <ifreund> novakane: a border color and river-status event so they can show up in status bars would be great
10:26 <ifreund> e.g. add a urgent_tags event to zriver_output_status_1 and bump the version number of all interfaces
```

novakne:
Add `border-color-urgent`, need to think a bit about how to do the river-status event. I mean the change in the protocol is done, now I need to see how to implement it in river

novakne:
Forgot to say, for anyone who want to test this with foot:

In your foot.ini add
```
[bell]
urgent=yes
```

Open two foot instances, in one exec `sleep 2 && echo -e '\a'`, unfocus the view where you executed the command and wait that it end

novakne:
Allright I think everything is implemented, change the border color, and send an event urgent_tags. Probably some things to improve, but should be good for a first review. I'll `rebase -i` the commits once we're good.

Do I need to change the version of the `river-status-unstable-v1.xml` filename too?

protocol/river-status-unstable-v1.xml
```diff
@@ -16,7 +16,7 @@
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
   </copyright>
 
-  <interface name="zriver_status_manager_v1" version="1">
+  <interface name="zriver_status_manager_v2" version="2">
```
Leon-Plickat, line 19: You don't need to update the `_v1` in the name of the interface. It's still `_v1`, just version 2 of it. Weird, I know.

protocol/river-status-unstable-v1.xml
```diff
@@ -68,6 +68,14 @@
       <arg name="tags" type="uint" summary="32-bit bitfield"/>
     </event>
 
+    <event name="urgent_tags">
+      <description summary="urgent tags of the output">
+        Sent once binding the interface and again whenever the tag is set
+        as urgent.
```
Leon-Plickat, line 74: This description sounds like the event is send for tags individually.

What should happen when one tag is urgent and another tags also gets urgent? I believe the second `urgent_tags` event should contain both tags, but the description does not reflect this.

Also, is this supposed to be send again when a tag looses urgency (yes, because otherwise it isn't really useful)? Should also be in the description.

river/View.zig
```diff
@@ -506,6 +514,13 @@ pub fn notifyAppId(self: Self) void {
     }
 }
 
+/// Called when the view requests to be focused
+pub fn requestActicate(self: *Self) void {
+    self.current.urgent = true;
+    self.output.sendUrgentViews(self.current.tags);
```
Leon-Plickat, line 520: This is not particularly useful, I think.

In a scenario where you have an urgent view on tag 1 and then an urgent view on tag 2 appears, the second urgent tags event will make the client believe tag 1 is no longer urgent, which is wrong.

Instead there should be an `urgent_tags` field in the output state.

Leon-Plickat (changes requested):

protocol/river-status-unstable-v1.xml
novakne: Thanks, I'm so confused by protocols versionning

river/View.zig
novakne: My first version used a `urgent_tags` in the output state, I removed it because I thought it was not worth the complication, multiple urgent views didn't really cross my mind :/

protocol/river-status-unstable-v1.xml
novakne: So I should use an array like in `view_tags` instead right?

Leon-Plickat: Not what I had in mind initially, but I think it is a pretty good idea.

river/View.zig
novakne: I tried something... not sure if it works as intended tbh, I don't know how to test this, I can't make it works as I want for now,

Leon-Plickat:
Two more things:

* Before setting a view to urgent, check if it is focused. Marking a focused view as urgent makes no sense.
* Set urgent to false when a view is focused.

novakne:
Sorry I rebase everything and clean up the commits because it was too messy, sorry if it messed up the reviews.

To resume currently,  the implementation and the borders colors change should be good, so I need:
- Send the urgent_tags events to client when a view is urgent and when it loses its urgency
- Remove urgent state when view is focused

novakne:
Welp I have no idea how to know when the view loses is urgency, and can't find a good way to add urgent_tags in output state and send it to client. So I'm kinda stuck right now :/

river/View.zig
novakne: I mark this as resolved since I changed a lot of things and this code is no more here but the implementation of this is not done

novakne:
Not sure if there is other place than in Seat.zig (4a03733 ) for removing the urgency of  a view when focused.

What would be the best way to implement an `urgent_tags` in `output.state`, `std.ArrayList(u32)`? 

What would be the step?
1. In View. zig: In `requestActivate()` I'll call `self.output.sendUrgentTag()`.
2. In Output.zig: I make a `pub fn sendUrgentTags()` that makes a `std.ArrayList(u32)`. This is where I'm stuck, and not sure how I would remove a tag from this when the view loses urgency.
3. In OutputStatus.zig: A `pub fn sendUrgentTags()` that send a `wl.Array` of the `self.output.current.urgent_tags`.

So I'll take any advice on what's the best way to implement this.

novakne:
So I come with somthing I think is a good solution, at least the best I can think of and make. Though currently I have a segfault, my pointers are messed up but I don't know how to fix it.

@ifreund @Leon-Plickat if someone have the time to look at it and help me, I would appreciate that, thanks, no rush.
[gdb.log](https://0x0.st/-yKx.log)

river/Seat.zig
```diff
@@ -164,6 +164,10 @@ pub fn focus(self: *Self, _target: ?*View) void {
             unreachable;
         }
         self.setFocusRaw(.{ .view = view });
+        if (view.current.urgent) {
+            view.current.urgent = false;
+            view.output.sendUrgentTags();
+        }
```
ifreund, line 170: In order to ensure frame perfection, we can't modify the current (i.e. displayed) state here. Instead we should set the pending state and call sendUrgentTags() in `Root.commitTransaction()` or `XdgToplevel.handleCommit()` if the pending state being applied differs from the current state.

Also this check should be moved to `setFocusRaw()` where `view.setActivated(false)` is called.

river/View.zig
```diff
@@ -506,6 +514,12 @@ pub fn notifyAppId(self: Self) void {
     }
 }
 
+/// Called when the view requests to be focused
+pub fn requestActicate(self: *Self) void {
+    self.current.urgent = true;
```
ifreund, line 519: Same thing here, we must modify the pending state not the current state (and in this case start a transaction to apply it).

Also this function is only called in handleRequestActivate(), why don't we just inline this code there and get rid of this function?

river/View.zig
```diff
@@ -535,3 +549,14 @@ fn handleForeignClose(
...
     self.close();
 }
+
+fn handleRequestActivate(
+    listener: *wl.Listener(*wlr.XdgActivationV1.event.RequestActivate),
+    event: *wlr.XdgActivationV1.event.RequestActivate,
+) void {
+    const self = @fieldParentPtr(Self, "request_activate", listener);
+    const xdg_surface = fromWlrSurface(event.surface);
+    if (xdg_surface) |view| {
```
ifreund, line 559:
```suggestion
    if (fromWlrSurface(event.surface)) |view| {
```

ifreund (changes requested):
The goals here look perfect, just a few issues to work out with the implementation.

river/OutputStatus.zig
```diff
@@ -38,9 +38,10 @@ pub fn init(self: *Self, output: *Output, output_status: *zriver.OutputStatusV1)
 
     output_status.setHandler(*Self, handleRequest, handleDestroy, self);
 
-    // Send view/focused tags once on bind.
+    // Send view/focused/urgent tags once on bind.
     self.sendViewTags();
     self.sendFocusedTags(output.current.tags);
+    self.sendUrgentTags(0);
```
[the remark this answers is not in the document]
novakne: Can I call `self.output.sendUrgentTags()` instead here or do I need to make another loop to iterate?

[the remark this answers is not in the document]
ifreund: Calling `Output.sendUrgentTags()` would send the tags to all existing clients even though nothing has changed. Let's avoid that and make another loop.

river/Seat.zig
novakne: Same question as above, do I need a new loop in `Root.commitTransaction()` and `XdgToplevel.handleCommit()` or can I call `output.sendUrgentTags();` and `view.output.sendUrgentTags();`?

ifreund: In `Root.commitTransaction()` and `XdgToplevel.handleCommit()` we need to send new information to all river status clients if the tags containing urgent views have changed. It looks like your `Output.sendUrgentTags()` function sends the current set of urgent tags on the output to all clients, so it would make sense to use that.

I believe there are three cases in which the set of urgent tags for a given output can change:

1. the urgent state of some view on the output changes
2. the tags of an urgent view change
3. an urgent view is sent to a different output (need o send new set of urgent tags for both outputs)

In each of these cases, we must send the new urgent tag set to all river-status clients.

novakne:
I think I fixed everything, also when it's all good let me know if you want me to rebase and clean up all the commits

protocol/river-status-unstable-v1.xml
```diff
@@ -75,9 +75,18 @@
...
+        Sent once binding the interface and again whenever a tag contain a
+        view marked as urgent or when losing its urgency. Tags stay till
+        the view loses its urgency.
+      </description>
+      <arg name="tags" type="uint" summary="32-bit bitfield"/>
+    </event>
   </interface>
 
-  <interface name="zriver_seat_status_v1" version="1">
+  <interface name="zriver_seat_status_v1" version="2">
```
ifreund, line 89: We actually don't need to bump this interface's version as neither it nor any of its child interfaces have been modified. See: https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-Versioning
```suggestion
  <interface name="zriver_seat_status_v1" version="1">
```

river/Seat.zig
```diff
@@ -195,6 +195,10 @@ pub fn setFocusRaw(self: *Self, new_focus: FocusTarget) void {
             .view => |view| {
                 view.pending.focus -= 1;
                 if (view.pending.focus == 0) view.setActivated(false);
+                if (view.pending.urgent) {
+                    view.pending.urgent = false;
+                    view.output.sendUrgentTags();
```
ifreund, line 200: This shouldn't happen right away, only when the transaction has been completed.
```suggestion
```

river/Root.zig
```diff
@@ -400,7 +400,10 @@ fn commitTransaction(self: *Self) void {
             view.dropSavedBuffers();
         }
 
-        if (view_tags_changed) output.sendViewTags();
+        if (view_tags_changed) {
+            output.sendViewTags();
+            output.sendUrgentTags();
+        }
```
ifreund, line 406: This logic isn't right, we need to be checking for the 3 cases I listed in another comment:

> 1. the urgent state of some view on the output changes
> 2. the tags of an urgent view change
> 3. an urgent view is sent to a different output (need to send new set of urgent tags for both outputs)

Only 1 and 2 are relevant here though and I see you've already handled 3. I'd suggest introducing an `urgent_tags_dirty` bool similar in usage to `view_tags_changed`.

river/XdgToplevel.zig
```diff
@@ -276,7 +276,10 @@ fn handleCommit(listener: *wl.Listener(*wlr.Surface), surface: *wlr.Surface) voi
             } else {
                 const self_tags_changed = view.pending.tags != view.current.tags;
                 view.current = view.pending;
-                if (self_tags_changed) view.output.sendViewTags();
+                if (self_tags_changed) {
+                    view.output.sendViewTags();
+                    view.output.sendUrgentTags();
```
ifreund, line 281: This logic also needs to be fixed in the same way as the logic in `Root.commitTransaction()` described above.

ifreund (changes requested):

river/Root.zig
novakne: Should be better now, I'm not sure if 2 is implemented right though

ifreund (approved):
LGTM, just did a few touchups while testing. Thanks!

PR #399, opened 2021-08-14
cursor: Support resizing in all four directions

Leon-Plickat:
Currently river will always resize from the bottom right corner of a view when resizing via pointer. This patch makes it possible to "grab" all four corners of a view and resize it in that direction.

Still a bit WIP, but works fine

Still to do:
* [x] cursor position
* [x] min size
* [x] max size
* [x] XDG toplevel resize requests
* [x] single-sided resize modes

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -40,6 +40,111 @@ const View = @import("View.zig");
 const ViewStack = @import("view_stack.zig").ViewStack;
 const XwaylandUnmanaged = @import("XwaylandUnmanaged.zig");
 
+pub const ResizeDirection = enum {
+    top,
+    right,
+    bottom,
+    left,
+    topleft,
+    topright,
+    botleft,
+    botright,
+
+    /// Returns the resize direction that makes the most sense based on the
+    /// position of the cursor on the current surface.
+    pub fn fromCursorPosition(cursor: *Self, view: *View) ResizeDirection {
+        const view_box = &view.current.box;
+        const output_box = server.root.output_layout.getBox(view.output.wlr_output).?;
+        const pointer_surface_x = (@floatToInt(i32, cursor.wlr_cursor.x) - output_box.x) - view_box.x;
+        const pointer_surface_y = (@floatToInt(i32, cursor.wlr_cursor.y) - output_box.y) - view_box.y;
+
+        const Position = enum {
+            start, // top or left
+            center,
+            end, // bottom or right
+
+            /// Arbitrary minimum size of resize "handle" area.
+            const min_size = 20;
+
+            comptime {
+                // Otherwise an integer underflow may occur.
+                assert(min_size < View.min_size / 2);
+            }
+
+            fn fromCoords(pos: i32, size: u32) @This() {
+                const handle: u32 = math.max(min_size, size / 5);
+                if (pos < handle) return .start;
+                if (pos > size - handle) return .end;
+                return .center;
+            }
+        };
+
+        const horizontal = Position.fromCoords(pointer_surface_x, view_box.width);
+        const vertical = Position.fromCoords(pointer_surface_y, view_box.height);
+
+        // The compiler complains when returning on outer switch...
+        switch (horizontal) {
+            .start => return switch (vertical) {
+                .start => .topleft,
+                .center => .left,
+                .end => .botleft,
+            },
+            .center => return switch (vertical) {
+                .start => .top,
+                .center => .botright, // the center should not be a dead-zone
+                .end => .bottom,
+            },
+            .end => return switch (vertical) {
+                .start => .topright,
+                .center => .right,
+                .end => .botright,
+            },
+        }
+    }
+
+    /// Tries to convert wlr.Edges to ResizeDirection.
+    pub fn fromWlrEdges(edges: wlr.Edges) !ResizeDirection {
+        // Only two edges may be active at max and they must not be opposing.
+        if (edges.top and edges.bottom) return error.InvalidEdges;
+        if (edges.right and edges.left) return error.InvalidEdges;
+
+        if (edges.top) {
+            if (edges.right) return .topright;
+            if (edges.left) return .topleft;
+            return .top;
+        }
+
+        if (edges.bottom) {
+            if (edges.right) return .botright;
+            if (edges.left) return .botleft;
+            return .bottom;
+        }
+
+        if (edges.right) return .right;
+
+        if (edges.left) return .left;
+
+        return .botright;
+    }
+
+    /// Return the canonical name of the cursor belonging the the resize
+    /// operation corresponding to the resize direction. The return value is a
+    /// NULL terminated string, since the return of this function will only ever
+    /// be used when calling functions from C libraries.
+    pub fn cursorNameZ(self: ResizeDirection) [*:0]const u8 {
+        return switch (self) {
+            .top => "n-resize",
+            .right => "e-resize",
+            .bottom => "s-resize",
+            .left => "w-resize",
+            .topleft => "nw-resize",
+            .topright => "ne-resize",
+            .botleft => "sw-resize",
+            .botright => "se-resize",
+        };
+    }
+};
+
 const Mode = union(enum) {
     passthrough: void,
     down: *View,
@@ -58,7 +163,9 @@ const Mode = union(enum) {
         view: *View,
         delta_x: f64 = 0,
         delta_y: f64 = 0,
-        /// Offset from the lower right corner of the view
+        direction: ResizeDirection,
+
+        /// The distance between cursor and the nearest corner or edge.
         offset_x: i32,
         offset_y: i32,
     },
@@ -233,7 +340,7 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
     if (event.state == .pressed) {
         self.pressed_count += 1;
     } else {
-        std.debug.assert(self.pressed_count > 0);
+        assert(self.pressed_count > 0);
         self.pressed_count -= 1;
         if (self.pressed_count == 0 and self.mode != .passthrough) {
             self.leaveMode(event);
@@ -251,7 +358,7 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
                     // handled we are done here
                     if (self.handlePointerMapping(event, view)) return;
                     // Otherwise enter cursor down mode, giving keyboard focus
-                    self.enterMode(.down, view);
+                    self.enterMode(.down, view, null);
                 }
             },
             .layer_surface => |layer_surface| {
@@ -355,8 +462,12 @@ fn handlePointerMapping(self: *Self, event: *wlr.Pointer.event.Button, view: *Vi
     return for (server.config.modes.items[self.seat.mode_id].pointer_mappings.items) |mapping| {
         if (event.button == mapping.event_code and std.meta.eql(modifiers, mapping.modifiers)) {
             switch (mapping.action) {
-                .move => if (!fullscreen) self.enterMode(.move, view),
-                .resize => if (!fullscreen) self.enterMode(.resize, view),
+                .move => if (!fullscreen) self.enterMode(.move, view, null),
+                .resize => if (!fullscreen) self.enterMode(
+                    .resize,
+                    view,
+                    ResizeDirection.fromCursorPosition(self, view),
+                ),
             }
             break true;
         }
@@ -647,7 +758,7 @@ fn surfaceAtFilter(view: *View, filter_tags: u32) bool {
     return view.surface != null and view.current.tags & filter_tags != 0;
 }
 
-pub fn enterMode(self: *Self, mode: std.meta.Tag((Mode)), view: *View) void {
+pub fn enterMode(self: *Self, mode: std.meta.Tag((Mode)), view: *View, resize_direction: ?ResizeDirection) void {
     log.debug("enter {s} cursor mode", .{@tagName(mode)});
 
     self.seat.focus(view);
@@ -655,21 +766,48 @@ pub fn enterMode(self: *Self, mode: std.meta.Tag((Mode)), view: *View) void {
     switch (mode) {
         .passthrough => unreachable,
         .down => {
+            assert(resize_direction == null);
             self.mode = .{ .down = view };
             server.root.startTransaction();
         },
         .move, .resize => {
             switch (mode) {
                 .passthrough, .down => unreachable,
-                .move => self.mode = .{ .move = .{ .view = view } },
+                .move => {
+                    assert(resize_direction == null);
+                    self.mode = .{ .move = .{ .view = view } };
+                },
                 .resize => {
-                    const cur_box = &view.current.box;
+                    view.setResizing(true);
+
                     self.mode = .{ .resize = .{
                         .view = view,
-                        .offset_x = cur_box.x + @intCast(i32, cur_box.width) - @floatToInt(i32, self.wlr_cursor.x),
-                        .offset_y = cur_box.y + @intCast(i32, cur_box.height) - @floatToInt(i32, self.wlr_cursor.y),
+                        .direction = resize_direction.?,
+                        .offset_x = undefined,
+                        .offset_y = undefined,
                     } };
-                    view.setResizing(true);
+
+                    const cur_box = &view.current.box;
+                    const pointer_surface_x = @floatToInt(i32, self.wlr_cursor.x) - cur_box.x;
+                    const pointer_surface_y = @floatToInt(i32, self.wlr_cursor.y) - cur_box.y;
+                    switch (resize_direction.?) {
+                        .top, .left, .topleft => {
+                            self.mode.resize.offset_x = pointer_surface_x;
+                            self.mode.resize.offset_y = pointer_surface_y;
+                        },
+                        .topright => {
+                            self.mode.resize.offset_x = util.safeIntDownCast(i32, cur_box.width) - pointer_surface_x;
+                            self.mode.resize.offset_y = pointer_surface_y;
+                        },
+                        .bottom, .right, .botright => {
+                            self.mode.resize.offset_x = util.safeIntDownCast(i32, cur_box.width) - pointer_surface_x;
+                            self.mode.resize.offset_y = util.safeIntDownCast(i32, cur_box.height) - pointer_surface_y;
+                        },
+                        .botleft => {
+                            self.mode.resize.offset_x = pointer_surface_x;
+                            self.mode.resize.offset_y = util.safeIntDownCast(i32, cur_box.height) - pointer_surface_y;
+                        },
+                    }
                 },
             }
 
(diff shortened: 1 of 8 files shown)


novakne:
Works fine for me too, didn't encounter any problems. I don't often use floating view but it is clearly an improvement.

river/Cursor.zig
```diff
@@ -747,26 +786,99 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
+            // readable for humans. The compiler should optimize the conversion out.
+            var botright_x = box.x + @intCast(i32, box.width);
+            var botright_y = box.y + @intCast(i32, box.height);
+
+            switch (data.corner) {
+                .topleft => {
+                    // If the max_{width,height} is larger than the max of u16,
+                    // it is larger than any existing screen anyway and usually
+                    // means that the view has not maximum size, so we can
+                    // truncate without losing relevant data.
```
ifreund, line 804: We could truncate to a u31 instead, and use `@as()` instead of `@intCast()` after truncation.

river/Cursor.zig
```diff
@@ -747,26 +786,99 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
+                    const max_x = box.x + @intCast(i32, @truncate(u16, constraints.max_width));
+                    std.debug.assert(min_x <= max_x);
+                    botright_x = math.clamp(botright_x + @floatToInt(i32, delta_x), min_x, max_x);
+
+                    const min_y = box.y + min_height;
+                    const max_y = box.y + @intCast(i32, @truncate(u16, constraints.max_height));
+                    std.debug.assert(min_y <= max_y);
+                    botright_y = math.clamp(botright_y + @floatToInt(i32, delta_y), min_y, max_y);
+                },
+            }
```
ifreund, line 848: I think it should possible to avoid this level of code duplication. Perhaps something like the approach in tinywl would work:
https://github.com/swaywm/zig-wlroots/blob/c0703a0b59c2930b782e4811a2aedf4b6ab7fbae/tinywl/tinywl.zig#L275-L313

ifreund (commented):
This is a very nice step up! Are you planning to support single-edge resizes in this PR as well? Clients may request them and it would be cool if we could honor those requests.

Leon-Plickat:
> Are you planning to support single-edge resizes in this PR as well?

Yes, I want to support all resize-modes that clients can request. However for serverside resize bindings I think only the corner ones should be accessible (which resize in two directions at once), because otherwise the detection code will get a bit too messy (currently we only have to check if the cursor is in one of four rectangles, which is trivial. Also supporting the four single sided ones for this would require checking for eight areas, four of which not being square, which I really would like to avoid).

On the other hand, I'll have to redo the detection code anyway, since the way I implemented it currently won't work for XDG toplevel resize requests.

Leon-Plickat:
There are three possible ways to handle server side pointer resize binds:
gh1
gh2

**A:** This is how I currently have implemented it. It only exposes the corner resize modes. It is the simplest to implement and my preferred solution.

**B:** This exposes all eight possible resize modes. We could get rid of the slopes, but then there would be a dead-zone in the center which is IMO bad UX. This is the most complex option to implement.

**C:** This exposes only the edge resize options. Definitely simpler then **B** but still complex enough that I'd like to avoid it.  Also the UX of this option is probably the worst, since resizing a window  could take two separate operations with this.

ifreund:
What if instead of a dead zone we added a secondary bottom right resize zone?
resize

We could also then choose to make the edge/corner resize zones smaller, resulting in something that mirrors the current behavior of always doing a bottom right resize if the user doesn't make an effort to grab a specific edge/corner.
resize2

Leon-Plickat:
Interesting idea, I will try that.

river/Cursor.zig
Leon-Plickat: Interestingly, `@truncate()`-ing to u31 crashes.

Leon-Plickat:
This should be ready for a second round of review now.

Resizing with pointer binds works as expected.

I have yet to test with XDG toplevel resize request, because I don't know any program that makes use of these and did not feel like modifying an existing one tonight (Anything GTK won't work, because here clicking on the CSD shadow sends the resize requests. River focuses the view when clicking on the shadow but apparently does not send the click event to the client?)

river/Cursor.zig
```diff
@@ -747,26 +886,78 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
+
+            // Using two points instead of a box makes this logic a bit more
+            // readable for humans. The compiler should optimize the conversion out.
+            var botright_x = box.x + @intCast(i32, box.width);
+            var botright_y = box.y + @intCast(i32, box.height);
+
+            if (data.resize_direction == .top or data.resize_direction == .topleft or data.resize_direction == .topright) {
+                const min_y = math.max(@intCast(i32, border_width), botright_y - @intCast(i32, @truncate(u16, constraints.max_height)));
+                const max_y = botright_y - min_height;
+                std.debug.assert(min_y <= max_y);
```
ifreund, line 902: This assert is redundant with the one inside `std.math.clamp()`:
```zig
pub fn clamp(val: anytype, lower: anytype, upper: anytype) @TypeOf(val, lower, upper) {
    assert(lower <= upper);
    return max(lower, min(val, upper));
}
```

river/Cursor.zig
```diff
@@ -39,15 +39,125 @@ const Seat = @import("Seat.zig");
...
     resize: struct {
         view: *View,
-        /// Offset from the lower right corner of the view
-        offset_x: i32,
-        offset_y: i32,
+        resize_direction: ResizeDirection,
+
+        /// The distance between cursor and the nearest corner or edge.
+        offset_x: u64,
+        offset_y: u64,
```
ifreund, line 160: Why do these need to be made 64 bit?

river/Cursor.zig
```diff
@@ -98,17 +98,15 @@ pub const ResizeDirection = enum {
...
-        if (edges.top and edges.right and edges.left) return .botright;
-        if (edges.top and edges.bottom and edges.left) return .botright;
-        if (edges.right and edges.bottom and edges.left) return .botright;
+        if (edges.top and edges.top == edges.bottom) return error.InvalidEdges;
+        if (edges.right and edges.right == edges.left) return error.InvalidEdges;
+        if (edges.top and edges.right and edges.bottom and edges.left) return error.InvalidEdges;
+        if (edges.top and edges.right and edges.bottom) return error.InvalidEdges;
+        if (edges.top and edges.right and edges.left) return error.InvalidEdges;
+        if (edges.top and edges.bottom and edges.left) return error.InvalidEdges;
+        if (edges.right and edges.bottom and edges.left) return error.InvalidEdges;
```
ifreund, line 109: I'm pretty sure this can be simplified to just
```suggestion
        if (edges.top and edges.bottom) return error.InvalidEdges;
        if (edges.right and edges.left) return error.InvalidEdges;
```

river/Cursor.zig
Leon-Plickat: They do not need to be. I made them 64 bits to match the types other coordinate vars.

ifreund: We use `i32` for coordinates of views in most places. Layout coordinates are `f64`, but don't be deceived by the 64 in that type, an `f64` cannot represent all values of a `u64`/`i64` precisely.

Leon-Plickat:
Oh, you're right. It's a typo.

river/Cursor.zig
```diff
@@ -747,26 +878,74 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
         },
         .resize => |data| {
             const border_width = if (data.view.draw_borders) server.config.border_width else 0;
-
-            // Set width/height of view, clamp to view size constraints and output dimensions
             const box = &data.view.pending.box;
-            box.width = @intCast(u32, math.max(0, @intCast(i32, box.width) + @floatToInt(i32, delta_x)));
-            box.height = @intCast(u32, math.max(0, @intCast(i32, box.height) + @floatToInt(i32, delta_y)));
+            const constraints = data.view.getConstraints();
+            const min_width = @intCast(i32, math.max(View.min_size, constraints.min_width));
```
ifreund, line 883: This can potentially overflow if the min_width is the maximum value of a u32.

river/Cursor.zig
```diff
@@ -747,26 +878,74 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
+            const min_height = @intCast(i32, math.max(View.min_size, constraints.min_height));
+
+            // Using two points instead of a box makes this logic a bit more
+            // readable for humans. The compiler should optimize the conversion out.
+            var botright_x = box.x + @intCast(i32, box.width);
+            var botright_y = box.y + @intCast(i32, box.height);
+
+            if (data.resize_direction == .top or data.resize_direction == .topleft or data.resize_direction == .topright) {
+                const min_y = math.max(@intCast(i32, border_width), botright_y - @intCast(i32, @truncate(u16, constraints.max_height)));
```
ifreund, line 892:
```suggestion
                const min_y = math.max(@intCast(i32, border_width), botright_y - @as(i32, @truncate(u16, constraints.max_height)));
```

Also, we could move the `@intCast()` for border_width up to where it is defined, or perhaps even better just make it a u16 in `Config`.

It might make sense to move the casting for max_height up by the casting for min_height.

river/Cursor.zig
```diff
@@ -637,13 +743,38 @@ pub fn enterMode(self: *Self, mode: std.meta.Tag((Mode)), view: *View) void {
...
+                    const cur_box = &view.current.box;
+                    const pointer_surface_x = @floatToInt(i32, self.wlr_cursor.x) - cur_box.x;
+                    const pointer_surface_y = @floatToInt(i32, self.wlr_cursor.y) - cur_box.y;
+                    switch (resize_direction.?) {
+                        .top, .left, .topleft => {
+                            self.mode.resize.offset_x = pointer_surface_x;
+                            self.mode.resize.offset_y = pointer_surface_y;
+                        },
+                        .topright => {
+                            self.mode.resize.offset_x = @intCast(i32, cur_box.width) - pointer_surface_x;
```
ifreund, line 766: we should probably either truncate here or not bother with the truncation below and just intCast()

ifreund (commented):
This is pretty nice, though I feel like the areas for corner/edge resize are a little small. I wonder if making them scale with the size of the view would be nicer, for example use 10% of the width/height.

This also seems to be buggy on monitors that are not at coordinates 0,0. Remember that View.current.box is in coordinates relative to the output.

Leon-Plickat:
> I wonder if making them scale with the size of the view would be nicer, for example use 10% of the width/height.

done.

Leon-Plickat:
> This also seems to be buggy on monitors that are not at coordinates 0,0.

Fixed.

Leon-Plickat:
This should be ready for the next round of review.

river/XdgToplevel.zig
```diff
@@ -349,15 +350,18 @@ fn handleRequestMove(
...
 }
 
 /// Called when the client asks to be resized via the cursor.
 fn handleRequestResize(listener: *wl.Listener(*wlr.XdgToplevel.event.Resize), event: *wlr.XdgToplevel.event.Resize) void {
     const self = @fieldParentPtr(Self, "request_resize", listener);
     const seat = @intToPtr(*Seat, event.seat.seat.data);
     if ((self.view.pending.float or self.view.output.pending.layout == null) and !self.view.pending.fullscreen)
-        seat.cursor.enterMode(.resize, self.view);
+        seat.cursor.enterMode(.resize, self.view, Cursor.ResizeDirection.fromWlrEdges(event.edges) catch {
+            log.err("Client requested resize with invalid edge values. The request will be ignored.", .{});
```
ifreund, line 362: I feel like this should be a protocol error but I don't see any error enum value that would match this. I asked in `#wayland`.

river/Cursor.zig
```diff
@@ -40,6 +40,108 @@ const View = @import("View.zig");
...
+            if (edges.right) return .botright;
+            if (edges.left) return .botleft;
+            return .bottom;
+        }
+
+        if (edges.right) return .right;
+
+        if (edges.left) return .left;
+
+        return error.InvalidEdges;
```
ifreund, line 124: I'm pretty sure this needs to handle the case in which no edges are set as well. `none` is a valid value of the `resize_edge` enum in the protocol. In this case let's default to bottom right.
```suggestion
        return .botright;
```

river/Cursor.zig
```diff
@@ -58,9 +160,11 @@ const Mode = union(enum) {
         view: *View,
         delta_x: f64 = 0,
         delta_y: f64 = 0,
-        /// Offset from the lower right corner of the view
-        offset_x: i32,
-        offset_y: i32,
+        resize_direction: ResizeDirection,
```
ifreund, line 163: Since this is usually accessed as `resize.direction` or similar, what do you thing about dropping the usually redundant prefix?
```suggestion
        direction: ResizeDirection,
```

river/Cursor.zig
```diff
@@ -802,26 +935,92 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
-            box.height = @intCast(u32, math.max(0, @intCast(i32, box.height) + @floatToInt(i32, dy)));
+            const constraints = data.view.getConstraints();
+
+            // Avoid potential integer overflow. No reasonable client will have
+            // a minimum size even remotely as large, so this should not be
+            // interfering with well behaved toplevels, but it prevents bad view
+            // constraints from crashing us. A max size larger than this is
+            // pretty common however as most views do not have a max size, which
+            // means they have max_width = max_height = maxInt(u32).
+            const max = math.maxInt(i32) / 2;
```
ifreund, line 947: Perhaps this logic belongs in `XdgToplevel.getConstraints()` and `XwaylandView.getConstraints()`?

river/Cursor.zig
```diff
@@ -802,26 +935,92 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
+            const min_width = @intCast(i32, math.max(View.min_size, math.min(max, constraints.min_width)));
+            const min_height = @intCast(i32, math.max(View.min_size, math.min(max, constraints.min_height)));
+            const max_width = @intCast(i32, math.min(max, constraints.max_width));
+            const max_height = @intCast(i32, math.min(max, constraints.max_height));
```
ifreund, line 951: Could use `math.clamp()` for the first two, and I think the `@intCast()`s are unnecessary.

river/Cursor.zig
```diff
@@ -802,26 +935,92 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
+                const max_y = botright_y - min_height;
+                box.y = math.clamp(box.y + @floatToInt(i32, dy), min_y, max_y);
+            } else if (data.resize_direction == .bottom or
+                data.resize_direction == .botleft or
+                data.resize_direction == .botright)
+            {
+                const min_y = box.y + min_height;
+                const max_y = box.y + max_height;
+                botright_y = math.clamp(botright_y + @floatToInt(i32, dy), min_y, max_y);
+            }
```
ifreund, line 972: I think this would be cleaner as a switch instead of an if chain:
```zig
switch (data.resize_direction) {
   .top, .topleft, .topright => { ... },
   .bottom, .botleft, .botright => { ... },
   .left, .right =>  {}, 
};
```
same below for the x axis

river/Cursor.zig
```diff
@@ -802,26 +935,92 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
...
+            // pretty common however as most views do not have a max size, which
+            // means they have max_width = max_height = maxInt(u32).
+            const max = math.maxInt(i32) / 2;
+            const min_width = @intCast(i32, math.max(View.min_size, math.min(max, constraints.min_width)));
+            const min_height = @intCast(i32, math.max(View.min_size, math.min(max, constraints.min_height)));
+            const max_width = @intCast(i32, math.min(max, constraints.max_width));
+            const max_height = @intCast(i32, math.min(max, constraints.max_height));
+
+            // Using two points instead of a box makes this logic a bit more
+            // readable for humans. The compiler should optimize the conversion out.
```
ifreund, line 954: We really don't care if it does, this will be plenty fast either way.
```suggestion
            // Using two points instead of a box makes this logic a bit more
            // readable for humans.
```

river/Cursor.zig
```diff
@@ -58,9 +160,11 @@ const Mode = union(enum) {
...
         delta_x: f64 = 0,
         delta_y: f64 = 0,
-        /// Offset from the lower right corner of the view
-        offset_x: i32,
-        offset_y: i32,
+        resize_direction: ResizeDirection,
+
+        /// The distance between cursor and the nearest corner or edge.
+        offset_x: f64,
+        offset_y: f64,
```
ifreund, line 167: I don't see a reason these need to be floating point, can we just make them `i32`s?

ifreund (changes requested):
Looking good, here's another round of picky code review!

river/XdgToplevel.zig
ifreund: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/123

river/Cursor.zig
Leon-Plickat: Not sure, since it depends on the type of width and height in the constraints sruct.

Perhaps `getConstraints()` should have a comptime type argument so that either i32 or u32 can be requested. Then that logic can be added to the i32 case.

river/XdgToplevel.zig
Leon-Plickat: I'll add a todo comment there.

Leon-Plickat:
Relevant messages from IRC:

```irc
<novakane> working great for me, just a bit weird with firefox https://0x0.st/-nV3.mp4 
...
<leon-p> so some other gtk programs are also a bit glithy, but nowhere near to the level of firefox. 
<leon-p> I suspect it has something to do with those programs updating to new sized rather slowly 
...
<ifreund> leon-p: I think the issue you're seeing with resizing firefox and other slow clients on your PR is a likely a bug in the transaction which has been hidden till now by the fact that the x/y of views are frozen during resize
<ifreund> and I don't think touching the transaction code right before a release is the best idea...
<leon-p> ifreund: agreed 
```

So turns out this apparently uncovered a flaw in the transaction system causing glitches when the client is a bit slow to respond to resizes. This only seems to happen when the resized window is using one of the new resize directions, meaning it not being anchored at the top left corner, which explains why it has not been discovered earlier.

ifreund:
Let's shelf this PR for now and come back to it after migrating to the wlroots scene graph and reworking the transaction system. I still want this user experience as I think it's better than the status quo but I don't think it's worth rewriting the transaction system right now to make it happen.

Leon-Plickat:
Agreed.

PR #402, opened 2021-08-18
docs: default keyboard mapping mode is "normal"

keithhub:


ifreund (approved):
Thanks, `git blame` says this was my typo :D

Issue #289, opened 2021-05-09
Better window matching for filters


Leon-Plickat:
Currently both the float and CSD filter use only the app-id to match windows, which is insufficient. Consider for example the case of the dev-tools window (which I want floating) of a browser (which I want tiled), which both per spec have the same app-id, and therefore will be treated identical by the filters.

I propose that filters should also be capable to match windows based on the view title.

ThreeFx:
How far do we want to take this? In the end we are probably talking about a filter DSL (who knows what crazy setups are required for some special applications).

Adding a matching option for window titles would help with day-to-day work (mainly getting the Firefox media sharing window to float nicely). I'm imagining something like this:

```
# matches the app-id 'asdf' for backwards compatibility
riverctl float-filter-add asdf

# matches the app-id 'asdf'
riverctl float-filter-add app-id:asdf

# matches the title 'asdf' exactly
riverctl float-filter-add title:asdf
```

Alternatively the title could be an optional argument

```
# matches app-id 'asdf' AND title 'qwer'
riverctl float-filter-add asdf qwer
```

What do you think?

Leon-Plickat:
> How far do we want to take this?

Not far at all.

The plan is to eventually outsource filters to external clients.

ThreeFx:
Sounds good, I was wanting to start hacking on some wayland stuff anyway :)

PR #415, opened 2021-08-31
Allow float-filter-add to inspect window titles

ThreeFx:
This extends the float-filter-add command using a 'prefix:string_to_match' syntax. The following prefixes are supported:

  * `title`, which matches window titles
  * `app-id`, which matches app ids

If no prefix is given, then `app-id` is assumed for backwards compatibility.

As an example following configuration floats all windows with the title 'asdf with spaces'.

    riverctl float-filter-add 'title:asdf with spaces'

The syntax was briefly discussed in #289. As I understand Leon, this is intended as a small quality-of-life change for existing setups. This implementation only checks view titles on creation, not upon title changes. I wouldn't know of a usecase for the latter, so I'd say we don't implement it.

Feedback wanted :) I'm new to both wayland and zig, so don't hold back!

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -41,14 +41,14 @@ over the Wayland protocol.
 *exit*
 	Exit the compositor, terminating the Wayland session.
 
-*float-filter-add* _app-id_
-	Add _app-id_ to the float filter list. Views with this _app-id_
-	will start floating. Note that this affects only new views, not already
-	existing ones.
+*float-filter-add* *app-id*|*title* _pattern_
+	Add a pattern to the float filter list. Note that this affects only new
+	views, not already existing ones. Title updates are also not taken into
+	account.
 
-*float-filter-remove* _app-id_
-	Remove an _app-id_ from the float filter list. Note that this affects only
-	new views, not already existing ones.
+*float-filter-remove* *app-id*|*title* _pattern_
+	Remove an app-id or title from the float filter list. Note that this
+	affects only new views, not already existing ones.
 
 *focus-output* *next*|*previous*|*up*|*right*|*down*|*left*
 	Focus the next or previous output or the closest output in any direction.
--- a/river/XdgToplevel.zig
+++ b/river/XdgToplevel.zig
@@ -212,15 +212,10 @@ fn handleMap(listener: *wl.Listener(*wlr.XdgSurface), xdg_surface: *wlr.XdgSurfa
         view.current.float = true;
         view.pending.float = true;
         view.pending.box = view.float_box;
-    } else {
-        // Make views with app_ids listed in the float filter float
-        if (toplevel.app_id) |app_id| {
-            if (server.config.float_filter.contains(mem.span(app_id))) {
-                view.current.float = true;
-                view.pending.float = true;
-                view.pending.box = view.float_box;
-            }
-        }
+    } else if (server.config.shouldFloat(view)) {
+        view.current.float = true;
+        view.pending.float = true;
+        view.pending.box = view.float_box;
     }
 
     // If the toplevel has an app_id which is not configured to use client side
--- a/river/command/filter.zig
+++ b/river/command/filter.zig
@@ -27,19 +27,31 @@ const ViewStack = @import("view_stack.zig").ViewStack;
 const Error = @import("../command.zig").Error;
 const Seat = @import("../Seat.zig");
 
+const FilterKind = enum {
+    @"app-id",
+    title,
+};
+
 pub fn floatFilterAdd(
     allocator: *mem.Allocator,
     seat: *Seat,
     args: []const [:0]const u8,
     out: *?[]const u8,
 ) Error!void {
-    if (args.len < 2) return Error.NotEnoughArguments;
-    if (args.len > 2) return Error.TooManyArguments;
+    if (args.len < 3) return Error.NotEnoughArguments;
+    if (args.len > 3) return Error.TooManyArguments;
 
-    const gop = try server.config.float_filter.getOrPut(util.gpa, args[1]);
+    const kind = std.meta.stringToEnum(FilterKind, args[1]) orelse return Error.UnknownOption;
+    const map = switch (kind) {
+        .@"app-id" => &server.config.float_filter_app_ids,
+        .title => &server.config.float_filter_titles,
+    };
+
+    const key = args[2];
+    const gop = try map.getOrPut(util.gpa, key);
     if (gop.found_existing) return;
-    errdefer assert(server.config.float_filter.remove(args[1]));
-    gop.key_ptr.* = try std.mem.dupe(util.gpa, u8, args[1]);
+    errdefer assert(map.remove(args[1]));
+    gop.key_ptr.* = try std.mem.dupe(util.gpa, u8, key);
 }
 
 pub fn floatFilterRemove(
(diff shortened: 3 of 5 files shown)


ThreeFx:
It isn't really proof, but here's a picture of Firefox' sharing icon, which nicely floats when configured as `title:Firefox — Sharing Indicator`.

image

river/command/filter.zig
```diff
@@ -36,10 +36,21 @@ pub fn floatFilterAdd(
...
+    if (std.mem.startsWith(u8, args[1], "title:")) {
+        key = key[6..];
+        map = &server.config.float_filter_titles;
+    } else {
+        if (std.mem.startsWith(u8, args[1], "app-id:")) {
+            key = key[7..];
+        }
+    }
+
+    const gop = try map.*.getOrPut(util.gpa, key);
```
ifreund, line 50: This explicit dereference to call a function in the struct's namespace isn't necessary in zig.
```suggestion
    const gop = try map.getOrPut(util.gpa, key);
```

river/command/filter.zig
```diff
@@ -51,7 +62,18 @@ pub fn floatFilterRemove(
...
 
-    if (server.config.float_filter.fetchRemove(args[1])) |kv| util.gpa.free(kv.key);
+    var map = &server.config.float_filter_app_ids;
+    var key = args[1];
+    if (std.mem.startsWith(u8, args[1], "title:")) {
+        key = key[6..];
+        map = &server.config.float_filter_titles;
+    } else {
+        if (std.mem.startsWith(u8, args[1], "app-id:")) {
+            key = key[7..];
```
ifreund, line 72: Could use the same `"foo".len` pattern here instead of the magic numbers.

ifreund (changes requested):
Thanks for the PR, this will be a nice feature to have :)

Seeing as river isn't even `0.1.0` yet, I'm not sure we should really care all that much about backwards compatibility. I think it would be a bit cleaner to use a separate argument instead of a string prefix for this feature. What would you think of the following?
```
riverctl float-filter-add app-id foobar
riverctl float-filter-add title barfoo
```

river/XdgToplevel.zig
```diff
@@ -215,7 +215,15 @@ fn handleMap(listener: *wl.Listener(*wlr.XdgSurface), xdg_surface: *wlr.XdgSurfa
     } else {
         // Make views with app_ids listed in the float filter float
         if (toplevel.app_id) |app_id| {
-            if (server.config.float_filter.contains(mem.span(app_id))) {
+            if (server.config.float_filter_app_ids.contains(mem.span(app_id))) {
+                view.current.float = true;
+                view.pending.float = true;
+                view.pending.box = view.float_box;
+            }
```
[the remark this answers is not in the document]
ThreeFx: Done.

river/command/filter.zig
```diff
@@ -36,10 +36,21 @@ pub fn floatFilterAdd(
     if (args.len < 2) return Error.NotEnoughArguments;
     if (args.len > 2) return Error.TooManyArguments;
 
-    const gop = try server.config.float_filter.getOrPut(util.gpa, args[1]);
+    var map = &server.config.float_filter_app_ids;
+    var key = args[1];
+    if (std.mem.startsWith(u8, args[1], "title:")) {
+        key = key[6..];
```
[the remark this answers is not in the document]
ThreeFx: Fortunately this isn't necessary anymore if we split the kind and the pattern into different arguments :)

river/command/filter.zig
```diff
@@ -33,13 +33,24 @@ pub fn floatFilterAdd(
...
     out: *?[]const u8,
 ) Error!void {
-    if (args.len < 2) return Error.NotEnoughArguments;
-    if (args.len > 2) return Error.TooManyArguments;
+    if (args.len < 3) return Error.NotEnoughArguments;
+    if (args.len > 3) return Error.TooManyArguments;
 
-    const gop = try server.config.float_filter.getOrPut(util.gpa, args[1]);
+    var kind = args[1];
+    var map = &server.config.float_filter_app_ids;
```
ThreeFx, line 40: I don't really like this, but I couldn't get it to work in any other way. Ideally, I'd like something like this:

```zig
var map = switch (kind) {
    "app-id" => ...,
    "title" => ...,
    else => return Error.UnknownOption,
};
```

but it seems like switching on strings is not supported (yet?).

Having the map be uninitialized would be fine as well, but that doesn't compile (I tried `var map;`). Maybe there is some Zig-foo I am missing to make this work @ifreund ?

ThreeFx:
On a side note (this is mostly interesting for upgrading users): The error messages generated when using the old syntax are very generic:

```
error: not enough arguments
error: not enough arguments
```

Maybe we should be a bit more verbose here?

river/command/filter.zig
ifreund: something like this would work:

```zig

File: river/command/filter.zig [hunk 6]
```diff
@@ -33,13 +33,24 @@ pub fn floatFilterAdd(
...
+    if (mem.eql(u8, kind, "app-id")) {
+        map = &server.config.float_filter_app_ids;
+    } else if (mem.eql(u8, kind, "title")) {
+        map = &server.config.float_filter_titles;
+    } else {
+        return Error.UnknownOption;
+    }
+
+    var key = args[2];
```
ifreund, line 49:
```suggestion
    const key = args[2];
```

river/command/filter.zig
```diff
@@ -48,10 +59,21 @@ pub fn floatFilterRemove(
...
+    if (mem.eql(u8, kind, "app-id")) {
+        map = &server.config.float_filter_app_ids;
+    } else if (mem.eql(u8, kind, "title")) {
+        map = &server.config.float_filter_titles;
+    } else {
+        return Error.UnknownOption;
+    }
 
-    if (server.config.float_filter.fetchRemove(args[1])) |kv| util.gpa.free(kv.key);
+    var key = args[2];
```
ifreund, line 75:
```suggestion
    const key = args[2];
```

doc/riverctl.1.scd
```diff
@@ -41,14 +41,14 @@ over the Wayland protocol.
...
-	will start floating. Note that this affects only new views, not already
-	existing ones.
-
-*float-filter-remove* _app-id_
-	Remove an _app-id_ from the float filter list. Note that this affects only
-	new views, not already existing ones.
+*float-filter-add* *app-id*|*title* _pattern_
+	Add a pattern to the float filter list. Note that this affects only new
+	views, not already existing ones. Titles are updates are also not
+	taken into account.
```
ifreund, line 47:
```suggestion
	Add a pattern to the float filter list. Note that this affects only new
	views, not already existing ones. Title updates are also not
	taken into account.
```

ifreund (changes requested):

river/command/filter.zig
ThreeFx: Oh that is much better. Thanks!

river/command/filter.zig
```diff
@@ -27,19 +27,31 @@ const ViewStack = @import("view_stack.zig").ViewStack;
...
 ) Error!void {
-    if (args.len < 2) return Error.NotEnoughArguments;
-    if (args.len > 2) return Error.TooManyArguments;
+    if (args.len < 3) return Error.NotEnoughArguments;
+    if (args.len > 3) return Error.TooManyArguments;
+
+    const kind = std.meta.stringToEnum(FilterKind, args[1]) orelse return Error.UnknownOption;
+    const map = switch (kind) {
+        FilterKind.@"app-id" => &server.config.float_filter_app_ids,
+        FilterKind.title => &server.config.float_filter_titles,
```
ifreund, line 47: We can use enum literals here to keep things a bit more concise:
```suggestion
        .@"app-id" => &server.config.float_filter_app_ids,
        .title => &server.config.float_filter_titles,
```

river/command/filter.zig
```diff
@@ -48,10 +60,17 @@ pub fn floatFilterRemove(
...
```
ifreund, line 69:
```suggestion
        .@"app-id" => &server.config.float_filter_app_ids,
        .title => &server.config.float_filter_titles,
```

ifreund (commented):
> Maybe we should be a bit more verbose here?

Honestly, I think it's fine as it is. I'm not too concerned about people running my software off of the master branch needing to deal with breaking changes. The upgrade UX from a 0.1.0 release to a 0.2.0 release would be more important to make as smooth as possible.

ifreund (approved):
LGTM, Thanks!

I wonder if we should do the same thing for the csd-filter-add/remove commands as well for consistency...

ThreeFx:
@ifreund I wondered as well, since `FilterKind` sounds like it should be able to be applied to any filtering. I'll open a new PR with those changes.

Issue #419, opened 2021-09-03
adaptive sync/VRR support


JopStro:
wlroots has the ability to enable variable refresh rate support on a monitor (under the name adaptive sync), and it would be nice to have the ability to use that feature with this compositor as I have a monitor which supports it.

I know that dwl just has it enabled by default on all monitors so I imagine that it should be safe to the same, but having it as a configuration option would likely work as well.

JopStro:
sorry if I was unclear, by "configuration option" I meant as in a toggle via riverctl to enable or disable it in case there are any unpredictable issues. Wlroots offers just offers a function that can be used to allow adaptive sync/VRR to be used on an output. I used dwl as an example because it just has is enabled by default on all outputs (by running the function whenever an output is created). It should be as possible with as little as 1 line of code (before you ask, I did try and do it myself to little avail. granted I gave up quite quickly and might give it another crack later).

Leon-Plickat:
I know exactly what you meant. But I would prefer to either implement this the right way, or not at all. And since we do not want riverctl to get output configuration options, the right way is to extend the output management protocol.

JopStro:
ok, so having it be configurable by the end user isn't an option (maybe global configuration is an option?). I've managed to get it working, on a fork. I'll set up a pull request if you want to test it/take a look to see if it's okay, as it is just one line.

ifreund:
I've added support for adaptive sync to the wlr-output-management protocol upstream: https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/commit/4264185db3b7e961e7f157e1cc4fd0ab75137568

When wlroots 0.16.0 is released river upgrades this issue can be closed.

PR #420, opened 2021-09-04
output: enable adaptive sync on new outputs

JopStro:
A small change that should enable adaptive sync on any monitor that supports it. I don't have any monitors that don't support adaptive sync so it may be worth testing it, but I imagine it won't present an issue as this is precisely how dwl implements adaptive sync and they don't seem to have any issue (though I do admit it feels pretty hacky, regardless). This is one of the options I suggested in an issue I raised earlier today - #419

### DIFF
--- a/.github/workflows/ci_build.yml
+++ b/.github/workflows/ci_build.yml
@@ -20,10 +20,10 @@ jobs:
     steps:
       - name: install deps
         run: |
-          xbps-install -S
-          xbps-install -uy xbps
+          xbps-install -Sy xbps
+          xbps-install -uy
           # wlroots dependencies as well as we build from source
-          xbps-install -uy MesaLib-devel libseat-devel eudev-libudev-devel libdrm-devel \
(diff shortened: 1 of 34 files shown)


ifreund (changes requested):
Welcome and thanks for the patch!

I don't think this is something we should enable unconditionally however. Adaptive sync is know to cause flickering on some hardware and with some clients. See for example https://github.com/swaywm/sway/issues/5076.

JopStro:
I added the ability to enable or disable adaptive sync on all outputs via riverctl. Which should work as a workaround in the meantime before a more robust fix. A neat side effect of this is that adaptive sync can be enabled and disabled in real time, possibly allowing for someone to make a script that temporarily enables it while running a program such as a game (sort of how it's disscussed in the issue you mentioned)

ifreund:
I added support for adaptive sync to the `wlr-output-management` protocol upstream. When river upgrades to wlroots version 0.16.0 it will support the new version of that protocol.

PR #424, opened 2021-09-06
river: Allow applying CSD based on window titles

ThreeFx:
This extends the `csd-filter-add` command to allow matching on window
titles as well, using a `csd-filter-add kind pattern` syntax. The
following kinds are supported:

  * `title`, which matches window titles
  * `app-id`, which matches app ids

Only exact matches are considered.

As an example following configuration applies client-side decorations to
all windows with the title 'asdf with spaces'.

    riverctl csd-filter-add title 'asdf with spaces'

----

This commit makes the `*-filter-add` consistent over floats and csds.

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -28,15 +28,15 @@ over the Wayland protocol.
 *close*
 	Close the focused view.
 
-*csd-filter-add* _app-id_
-	Add _app-id_ to the CSD filter list. Views with this _app-id_ are
-	told to use client side decoration instead of the default server
-	side decoration. Note that this affects both new views, as well as already
-	existing ones.
+*csd-filter-add* *app-id*|*title* _pattern_
+	Add _pattern_ to the CSD filter list. Views with this _pattern_ are told to
+	use client side decoration instead of the default server side decoration.
+	Note that this affects new views as well as already existing ones. Title
+	updates are not taken into account.
 
-*csd-filter-remove* _app-id_
-	Remove an _app-id_ from the CSD filter list. Note that this affects both new
-	views, as well as already existing ones.
+*csd-filter-remove* *app-id*|*title* _pattern_
+	Remove _pattern_ from the CSD filter list. Note that this affects new views
+	as well as already existing ones.
 
 *exit*
 	Exit the compositor, terminating the Wayland session.
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -62,8 +62,9 @@ modes: std.ArrayList(Mode),
 float_filter_app_ids: std.StringHashMapUnmanaged(void) = .{},
 float_filter_titles: std.StringHashMapUnmanaged(void) = .{},
 
-/// Set of app_ids which are allowed to use client side decorations
-csd_filter: std.StringHashMapUnmanaged(void) = .{},
+/// Sets of app_ids and titles which are allowed to use client side decorations
+csd_filter_app_ids: std.StringHashMapUnmanaged(void) = .{},
+csd_filter_titles: std.StringHashMapUnmanaged(void) = .{},
 
 /// The selected focus_follows_cursor mode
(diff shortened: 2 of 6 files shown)


river/Config.zig
```diff
@@ -156,3 +165,19 @@ pub fn shouldFloat(self: Self, view: *View) bool {
 
     return false;
 }
+
+pub fn csdAllowed(self: Self, toplevel: *wlr.XdgToplevel) bool {
```
ThreeFx, line 169: The implementation differs from the one for `filter`s, however, because I did not find a way to get a `View` from a `Decoration`: Instead, we pass `*wlr.XdgToplevel` around. I'm not sure whether this is neccessary or there is a better approach.

ThreeFx: I've pushed a commit which uses `*View`s, but thereby introducing additional complexity at the call site in `Decoration.zig`. I use the same mechanism as is used in `csdFilterUpdateViews`, let me know if this is better/worse than the above alternative.

ThreeFx:
I'm not clear on what semantics we want to have for `csd-filter-add`. When should a view have client-side decorations? My current understanding of the current implementation is:

* When the app-id filter matches the view's toplevel

What semantics do we want when also including the title? Currently, I think I've implemented it the following way:

* When the app-id or title matches the view itself

I don't have enough experience with Wayland to judge whether this is ok or not.

------

Separate point:

Adding/removing filters at runtime works for both app-id filters as well as title filters, but as is the case with floating, this currently does _not_ handle title updates. Since csds are applied for existing and new windows, should we handle title updates in this case?

doc/riverctl.1.scd
```diff
@@ -28,14 +28,13 @@ over the Wayland protocol.
...
-	told to use client side decoration instead of the default server
-	side decoration. Note that this affects both new views, as well as already
-	existing ones.
-
-*csd-filter-remove* _app-id_
-	Remove an _app-id_ from the CSD filter list. Note that this affects both new
+*csd-filter-add* *app-id*|*title* _pattern_
+	Add _pattern_ to the CSD filter list. Views with this _pattern_ are told to
+	use client side decoration instead of the default server side decoration.
+	Note that this affects both new views, as well as already existing ones.
```
ifreund, line 34: Not your typo, but this comment should be removed and "both" could be dropped as well.
```suggestion
	Note that this affects new views as well as already existing ones.
```

Leon-Plickat:
Only use CSD if the filter matches is the right thing here. This filter exists mostly for cosmetics.

I don't think we should care about updating titles. Eventually all this will be done by an external program, which can deal with that sort of thing.

ifreund (commented):
For any given view, the view's app-id and title are exactly the app-id and title of the view's backing xdg-toplevel. River's `View` struct exists purely to abstract over xwayland surfaces and normal xdg-toplevels. So, the way you've implemented that part looks good to me.

With regards to title updates, I'm not sure if it makes sense to handle them or not. Therefore let's go with the simpler option of not handling them for now and see if anyone comes up with a motivating use-case for doing so.

ThreeFx:
Alright, then this is ready for review. I've added a short note to `riverctl(1)` stating that we do not account for title updates.

I haven't tested this though, because I haven't got the slightest clue which apps use csd and which don't. Let me know which ones are good test candidates.

ifreund:
> I haven't tested this though, because I haven't got the slightest clue which apps use csd and which don't. Let me know which ones are good test candidates.

All apps theoretically support CSD thanks to GNOME's influence on early wayland developement.

I usually just test this stuff with foot, you should see a title bar appear if you run `riverctl csd-filter-add app-id foot`.

ThreeFx:
Ah indeed. I've tested it using my terminal (Alacritty) with both app-id and title - works like a charm.

ifreund (approved):
LGTM thanks!

Issue #368, opened 2021-07-22
Geometry resize/move


nagy135:
On x11/bspwm I was making use of [xdo](https://github.com/baskerville/xdo) in a script that positioned window to exact size/location (width,height, x, y).

Use case: I was watching video via mpv and wanted to do other stuff as well so I had bind that put it to floating mode and positioned it at one of bottom corners as a miniature above everything (and much more).

As I understand, external program cant resize other window like this on wayland (correct me if I m wrong) so it would have to be implemented in river. I wanted to try implementing it but first I would like to ask if adding feature like this to river would be desired at all.

novakne:
Did you look into `snap`,`move` and `resize` commands?

nagy135:
Yes I did, what I m looking for is **exact positioning with geometry** these commands work with **delta**. Imagine you want to move currently focused window to size of 200x200 in bottom right corner. How would you script that? If I knew current size and location I can imagine writing some script I guess, not sure how to do that tho ...

Leon-Plickat:
> i m not sure how to achieve exact size without knowing current one. Is there a way to get that information?

No, that is currently not available.

I think a `set-dimensions [width]x[height]` and a `set-position x y` command is not unreasonable (with the normal window bounds applied, of course). In the future those could potentially even be moved to the river-toplevel protocol extension.

As a workaround for now you can use the `snap` command and try to set the size via an mpv command flag.

nagy135:
I also think there should be *instance agnostic way* to achieve this because there are programs without command flag like that.

This exact usecase could be achieved with currently possible **delta resize/move** if i knew current size so i was thinking if cli command that **dumps current state** wouldnt be more reasonable. It might be usefull in other usecases as well (bspwm could dump json with that information). Any opinions about that? It might be more general solution to my feature request.

Leon-Plickat:
> It might be usefull in other usecases as well (bspwm could dump json with that information). Any opinions about that?

Generally features are only implemented in river if they have a good use case, so until some concrete example that can not be solved better another way is found, I am against this. Your use case can be solved far easier with commands to directly set the position and dimensions of a window on an output (note that the output dimensions are already available).

I feel like this is information clients should not have, users would only it for weird hacks. We should keep doing what other Wayland compositors do and only expose information that is strictly needed.

[...]

PR #430, opened 2021-09-11
view: Implement set-position

ThreeFx:
The command `set-position x y` sets the current view's coordinates to
(x, y) relative to the current output. The view will always remain
within the bounds of the current output. The view will be set to
floating.

This is the first half of #368.


ThreeFx:
Stuff to think about

- [ ] Using `u32` instead of `i32`. I'm not sure what the runtime guarantees for overflows are, any help is appreciated

ThreeFx:
In the IRC channel we found out that there is some discussion to be had here. The current implementation of `riverctl move` always retains the window size when moving a window. We are not sure what the best way to handle moving out-of-output for the `set-position` and `set-geometry` commands are. Two possible approaches are:

1. Mirror the current `move` and `resize` commands: Snap everything inbounds, preserving the window size as good as possible.
2. Have the borders "eat" the views if they move inside of them. An illustration of "eating":

```
    +----------+
    |  window  |
    |      +---+------+
    +------+---+      |
           |  output  |
           |          |
           +----------+
```

becomes

```
           +---+------+
           +---+      |
           |  output  |
           |          |
           +----------+
```

Option 1 is better for interactive commands (`move`, `resize`), while IMO option 2 is better for precision commands (`set-position`, `set-geometry`). Any opinions @Leon-Plickat @ifreund ?

Leon-Plickat:
Relevant IRC logs:

```
<bfiedler> Should a theoretical set-position x y be relative to the current output or absolute?
<leon-p> bfiedler: relative to the current output.
<leon-p> and also it must follow current convention by never moving a window outside an output
<bfiedler> leon-p: makes sense :)
<bfiedler> riverctl resize <direction> moves the output by effective_delta / 2 to make it a bit more natual
<bfiedler> Should a riverctl set-geometry command do that as well?
<leon-p> bfiedler: Good question. I think it should not.
<leon-p> The two commands have different use cases
<bfiedler> Me neither
<bfiedler> It makes it _much_ easier to use in conjunction with set-position
<leon-p> the resize by delta command is meant for interactive use. Moving the view so it appears like its center remains stationary is just good UX. However the set-geometry and set-position commands are meant for precise control and this feature would just be annoying here.
<bfiedler> is it guaranteed for a view that x_pos < effective_output_width - border_width >= 0?
<bfiedler> s/>= 0//g
<leon-p> bfiedler: no. Layout generators currently have no restrictions on where to place windows and therefore can position them outside the bounds of the output.
<bfiedler> what should happen when setting the geometry of an off-screen view?
<bfiedler> should we pull the view into focus
<leon-p> bfiedler: since it is set to floating, it should be pulled into the output bounds
<leon-p> going even further, I think these restrictions should apply to layout generators as well
<bfiedler> leon-p: that or let it be a protocol error
<leon-p> bfiedler: sounds reasonable. I'll open an issue about that
<bfiedler> The case that currently bugs me is the following: calling 'set-geometry w h' on a window that is placed at something like 2px from the lower-right corner. what should happen here?
<bfiedler> I want to move the position of the window as little as possible, however I have to move it since otherwise the box would be too small
<leon-p> I see two possible things to do: 1) keep the position and resize the view to min(possible_size, desired_size), or 2) move the view
<leon-p> and perhaps 3) print an error and abort the operation
<leon-p> which of these makes more sense depends on the use case I think
<bfiedler> I prefer 1), but this leads to cases where we might place parts of views outside of the output bounds
<bfiedler> 3) sounds good in this case
<leon-p> well, any part of the view being outside the output bounds is a complete no-go.
<leon-p> tbh I think there should be some unified View.moveResize() function that takes care of all the constraints, so we don't have to worry about it anymore in the future
<bfiedler> leon-p: +1 for moveResize, that's what I'm basically working on anyway ATM
<bfiedler> but then we have to decide what get's preference: the position or the geometry
<novakane> does view could be place between two outputs? since I don't have multiple monitors I don't know what's the current behavior
<bfiedler> I think it shouldn't be possible at the moment
<bfiedler> Is 2 * border_height < min(output.width, output.height) checked?
<bfiedler> that looks like it makes sense (maybe only with <= but nonetheless)
<leon-p> bfiedler: a view is constrained to an output even if there is another output next to it.
<leon-p> (not handling this case is one of the technological reasons for this restriction, btw.)
<leon-p> bfiedler: can't remember what's checked where, sorry
<bfiedler> resizing a window is actually resize-move
<leon-p> bfiedler: depends on the direction, but yes
<bfiedler> hmm
<bfiedler> I'm thinking of ways to unify the resizing handling, but it is not clear how to do it
<leon-p> yeah, it's a bit annoying. Be happy this is Wayland. On X it's not even possible to do that in a reliably frame-perfect way.
<bfiedler> riverctl resize would like a resize which can temporarily go outside of the view borders since it fixes itself by "centering" the move afterwards
<bfiedler> however set-geometry would like a resize which automatically clamps at the output bounds
<leon-p> bfiedler: it also allows to have a size bigger than the output dimensions and fixes itself afterwards AFAIK
<bfiedler> yeah if size > output_width then it gets clamped by the available width
<leon-p> what if you approch the problem from the POV of positioning two points on the output, the upper-left and lower-right corner, keeping both inside the bounds? Then both geometry or position overrule the other based on position. Perhaps that is abit more intuitive
<leon-p> bfiedler: ^
<leon-p> I used that method in my pointer resize PR and it works great there.
<bfiedler> sure, but the question is still: what if moveResize cannot handle it's arguments? should x/y positioning be retained and width/height truncated or should width/height be retained and x/y changed?
<leon-p> bfiedler: that's the point. My idea is to covert x,y,w,h into two sets of x,y and treat them individually.
<leon-p> whether dimensions or position is corrected then depends on which of the points moves outside of the bounds
<leon-p> that will ensure that the size of the view is exactly the size of the visible part of the view if it /could/ be moved outside the bounds, with the expection of the minimum view size
<bfiedler> what if minimum size comes into play?
<bfiedler> i.e. resizeMove (999, 999, 100, 100) on a 1000x1000 display
<leon-p> I think For min size the best way is to first calculate the new size and position, then if necessary correct the size and then if necessary correct position. Min size will usually never be encountered, so it's fine if it introduces a slight inconsistency IMO
<leon-p> same for the xdg toplevel min size
<leon-p> Or to explain it in simple words: "if you push the window into a border, the border eats that part of the window, until the window can get no smaller, then it bounces back"
<bfiedler> agree
<bfiedler> but then, this is unintuitive for move right
<leon-p> how so?
<bfiedler> It is nicer to have the window stop at the right border if I want to snap it right
<bfiedler> ah but then I could just have used snap
<leon-p> ah, that is what you mean. My proposal actually only applies to move-resize ops., not to pure move ops.
<bfiedler> ah, my bad
<leon-p> fore pure movement I don't think windows should be resized at all
<leon-p> should have pointed that out prbably
<bfiedler> makes sense
<leon-p> also pointer resize/move still needs to be special-cased, as they are right now.
<leon-p> I think at least.
<bfiedler> what about set-position vs move
<leon-p> good question. And TBH I don't have an answer.
<bfiedler> move should not change the window size, but I feel like set-position should have the border-eats-up-extra-window behaviour
<leon-p> I can get onboard with that
<bfiedler> I feel like that is the best tradeoff between precision and intuition
<leon-p> you know, we could always wait until monday and defer the decission making to ifreund :D
<leon-p> anyway, I think this requires more discussion
<bfiedler> +1
<bfiedler> I'll let it rest until Mo then :)
<leon-p> especially with regards to usecase. Something that is used interactively should behave differently than something that is used through scripts / a wayland client
<bfiedler> Alright, I've added a comment to my (now WIP) PR: https://github.com/ifreund/river/pull/430 
```

PR #432, opened 2021-09-11
river: add send-to-previous-tags command

pvsr:
Closes https://github.com/ifreund/river/issues/416. This isn't just for symmetry's sake, as I find that `send-to-previous-tags` really does complement `focus-previous-tags` well. `focus-previous-tags` is great because it allows you to temporarily replace the mental overhead of alternating between tags like "$mod+1, $mod+7, $mod+1..." with one repeated keystroke, but if you want to send a view to the alternate tagset you have to break out of that shortcut and recall which tags you were using. My typical use case is starting a couple of related, slow to start applications, e.g. an IDE and a SQL client, or Steam and a game, in my working tagset and sending them to the alternate tagset at they pop up.

Would it be worth adding default bindings for this and `focus-previous-tags` in the example init file?

### DIFF
--- a/completions/bash/riverctl
+++ b/completions/bash/riverctl
@@ -25,6 +25,7 @@ function __riverctl_completion ()
 			send-layout-cmd \
 			set-focused-tags \
 			focus-previous-tags \
+			send-to-previous-tags \
 			set-view-tags \
 			toggle-focused-tags \
 			toggle-view-tags \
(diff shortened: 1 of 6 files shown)


ifreund (approved):
LGTM, Thanks!

As for adding bindings to the example init file, feel free to do that in a follow up PR. We can discuss what bindings make the most sense there if necessary.

pvsr:
Thank you! I was under the impression that i3 or sway had default bindings for their `back_and_forth` commands that we could just copy, but it looks like I misremembered and they actually don't bind them by default, so I'm fine with doing the same thing and leaving them unbound.

tgy:
@pvsr this was never added to the example init file. could you detail your config to make the back-and-forth pattern work?

basically, i want Super+$i to switch to workspace $i if it's not the current workspace, but switch back to workspace $j ($j != $i) if $i is the current workspace

this is the common pattern i'm used it i3/sway/hyprland and that i want to reproduce in river

tgy:
@pvsr this was never added to the example init file. could you detail your config to make the back-and-forth pattern work?

basically, i want Super+$i to switch to workspace $i if it's not the current workspace, but switch back to workspace $j ($j != $i) if $i is the current workspace

this is the common pattern i'm used it i3/sway/hyprland and that i want to reproduce in river

pvsr:
My config is just
```shell
riverctl map normal $mod Z focus-previous-tags
riverctl map normal $mod+Shift Z send-to-previous-tags
```

I also like the pattern you describe but I don't think it's possible using the init file at the moment since command mappings are unconditional. You could probably do it by remapping commands every time you switch workspaces but I don't know of a good way to achieve that.

tgy:
this has now been implemented as a tool in by @Leon-Plickat [here on sr.ht](https://git.sr.ht/~leon_plickat/river-bnf)

PR #447, opened 2021-10-07
view management protocol design

Leon-Plickat:
**Everyone who is interested in this PR, please note that this is just a collection of late-night braindumps and that the design I came up with here turned out to be insufficient and will not be the basis for the final thing. That, by the way, is still some time away, so don't hold your breath.**

### DIFF
--- a/protocol/river-keybind-v1.xml
+++ b/protocol/river-keybind-v1.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="river_control_v2">
+  <copyright>
+    Copyright 2021 The River Developers
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose with or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+    MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+  </copyright>
+
+  <description summary="let clients register global keybinds">
+    This protocol specifies a way for clients to register global keybinds.
+
+    Warning! The protocol described in this file is currently in the
+    testing phase. Backward compatible changes may be added together with
+    the corresponding interface version bump. Backward incompatible changes
+    can only be done by creating a new major version of the extension.
+  </description>
+
+  <interface name="river_keybind_manager_v1" version="1">
+    <description summary="manage keybind objects">
+      A global factory for river.
+    <description>
+
+    <request name="destroy" type="destructor">
+      <description summary="destroy the river_keybind_manager_v1 object">
+        This request indicates that the client will not use the
+        river_keybind_manager object any more. Objects that have been created
+        through this instance are not affected.
+      </description>
+    </request>
+
+    <request name="register_keybind">
+      <description summary="create a river_keybind_v1 object">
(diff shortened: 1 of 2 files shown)


Leon-Plickat:
## Some things are a bit opinionated

* The protocol does not have a concept of hidden views. Whether a view is displayed is solely based on the `river_view_handle.set_output` request. If the requested output is `NULL`, the view is not displayed on any output and therefore hidden.
* VM can set the border colours of views (maybe it should also be able to set the width?). However any view decoration more fancy than that is not possible. For a `v1` I think this is fine, but at least the option to decide between SSD and CSD should be there IMO.

## Some open questions

* What happens when the VM disconnects? I think river should keep all views on their outputs and any view not displayed (by being on the `NULL` output) should be moved to the first output. This way if the VM crashes, the user can restart it without complete session loss. The protocol does allow a VM to recognize already existing views, so I see no problem here.
* What happens when an output with views on it disappears? I think river should move all these views to the `NULL` output and let the VM decide the next steps.
* How to handle moving views via pointer? I think some things should stay as they are: For example by not allowing views to leave the output bounds we avoid the massively annoying problem of having a view on multiple outputs simultaneously, which the protocol currently straight up ignores. This requires more thought.
* What is the best way to send a colour through Wayland? A string certainly is not the right solution here.

Leon-Plickat:
Perhaps we should stay even more opinionated and simple than this protocol proposal already is.

Maybe all we should do is expand the river-layout protocol a little bit so that:

* layout generators can register keybinds
* layout generators have control over the view focus
* layout generators have control over the view border colour
* *perhaps* layout generators can hide views to implement tags / workspaces client side

This means:

* floating vs tiling separation is still server side
* no need to come up with something for pointer move / resize
* no way to implement a floating WM for river, but I think that is fine as river already has decent floating window management when no layout generator is active.

---

I think we should still fundamentally cater towards tiling window management. So it might make sense to stay with layout generators over view managers and simply grant them just a little more agency. I would really like to avoid doing the pointer related stuff over a protocol.

If we expand layout generators with just two abilities, control over view stack and focus, we already gain most of what users of a possible view management protocol would want anyway.

Leon-Plickat:
Another midnight brain dump. 

Things to do if tags / workspaces end up being implemented by the view manager, not the server:

* We should keep the status string we have in the current layout generator protocol, but change its intention from human readable to machine readable. River would then forward this status to clients. This is needed to allow status panels and other desktop widgets to display information like the currently focused workspace / tag. Otherwise, the view manager would need to include their own status widgets, which might be nice for full desktop shells, but not something I want to enforce. Unfortunately this likely leads to widgets needing to explicitly support layout generators, but it is better than the alternative IMO. And maybe there could be some sort of informal standard.
* View managers should be able to attach an arbitrary string to views and query it. This would allow view manager configuration to survive restarting the view manager. As an example, it could set the string to `stacktile-v1-workspace:4`. A user could then kill it and when the view manager is restarted it reads the string and puts the view on that workspace again. This would allow things like updating your view manager without losing the session and without having to re-organize everything.

Not sure if those two ideas make sense, I have only considered them superficially so far, but the features they make possible are things I'd like to have.

Magnitudexx:
If or whenever this need to be tested I am up for it!

and3rson:
Me too. This feature is the last blocker for me before I can fully switch to river from sway!

Leon-Plickat:
It should be noted that this won't be done and implemented anytime soon (the attempt I made here already turned out to be insufficient). And even then it will not be immediately stabilized. The layout protocol showed us that multiple iterations are necessary before a protocol turns out the way we want it too and then still may contain errors and design bugs.

jokeyrhyme:
One thing I really like about `river` is that config/init and `rivertile` are separate executables, and able to be swapped with alternative implementations

I'd like to suggest is that we preserve this modularity and allow/support/encourage the "view manager" to be just another executable that is registered with a `riverctl` command during config/init

Leon-Plickat:
@jokeyrhyme We don't have a design yet, but on IRC we more or less came to the decision to get rid of `riverctl` and  the init file. The view manager however will of course still be a separate executable that you can easily swap out for a custom one (that is the entire point of all this).

The idea is that river at startup loads a view manager which is then responsible for window management, configuration and user interaction. Our goal is to make river super flexible and adaptable and this design makes that possible. On the one hand you can have view managers that execute an init file at startup and have a CLI tool for configuration, meaning a river experience pretty much equal to what we have right now. On the other hand, it allows people to write full on highly integrated desktop-shells (which I might do eventually), that would simply not work well with the current model.

Basically we want to make even less assumptions about peoples setups.

Another reason is that a lot of what riverctl can do depends on functionality/state that river would loose anyway when we have view managers. It makes a lot more sense to make view manager developers write their own riverctl-like tools for their view managers, which also allows them to make it work well with whatever window management paradigm and quirks they end up implementing.

Leon-Plickat:
Closing, because people seem mistake this for a real PR instead of just an idea-dump.

PR #449, opened 2021-10-09
view: Implement set-position and set-geometry

ThreeFx:
The command `set-position x y` sets the current view's coordinates to be
(x, y) relative to the current output. The view will always remain
within the bounds of the current output. Any parts of the view that end
up outside of the current output are silently truncated.

The command `set-geometry w h` sets the geometry of a view to be `w`
wide and `h` high. Any parts of the view that end up outside of the
current output's bounding box are silently truncated.

View truncation is applied as shown below:

    +-------------+              +-------------+
    |   output    |              |   output    |
    |      +------+--+  ------>  |      +------+
    |      | view |  |           |      | view |
    +------+------+  |           +------+------+
           +---------+

-----

In addition to the above commands, this refactors the `move` and `resize` functions to use the `setPosition` and `setGeometry` commands. This way there is only one true way of setting position and geometry of windows.


ThreeFx:
This closes #368 btw :)

ifreund:
> It's nice for positioning some windows as part of a larger setup (e.g. a terminal running `find <zig files> | entr zig build`).

hmm, could you elaborate a bit more on this? From a code standpoint this PR seems solid, I'm just not yet convinced that this feature is widely useful due to the restriction to always targeting the focused view.

ifreund:
Hit a panic when attempting to move a view past the edge of the output in a nested session:
```

**Comment by ifreund** [comment]
I've come around on having this feature in river. It's not really that complex and lets us unify some code paths. We *should* also be able to get rid of the separate resize handling in `Cursor.processMotion()` eventually though that could be done in a follow up PR.

**Comment by Danacus** [comment]
To add to what @ThreeFx already said, this feature could be useful for floating terminals. For example I use a floating terminal with an mpc client that can toggled with a key binding.

```
scratch_tag=$((1 << 20 ))
mpc_tag=$((1 << 21 ))

riverctl map normal Mod4 W toggle-focused-tags ${scratch_tag}
riverctl map normal Mod4+Shift M toggle-focused-tags ${mpc_tag}

riverctl set-focused-tags ${scratch_tag}

riverctl spawn "alacritty"
sleep 0.1
riverctl toggle-float
riverctl resize horizontal 200
riverctl resize vertical 200

riverctl set-focused-tags ${mpc_tag}

riverctl spawn "alacritty -e mpc-rs"
sleep 0.1
riverctl toggle-float
riverctl resize horizontal -100
riverctl resize vertical -200
riverctl snap down

riverctl set-focused-tags 1
```

`set-position` and `set-geometry` would give more control compared to `snap` and `resize`.

ifreund:
> Is this feature still planned for River as I'd find it extremely useful ?

I'm not opposed to it, though I don't have plans to implement it myself. In any case this PR is unfortunately quite out of date now.

I'd recommend anyone interested on working on this to wait until I resolve the issues described in #828 to avoid conflicts and save time for both of us.

PR #459, opened 2021-10-30
Cursor: fix move/resize with high poll rate/low dpi mice

### LINKED ISSUE #255: Moving/Resizing views with high polling rate mouse doesn't work properly
I recently got a new mouse and the first thing I noticed is that moving views around the screen felt weird, kinda like it didn't want to move sometimes.

I believe this is because River converts delta_x and delta_y into ints to move the view and then moves the cursor based on the difference in the current and pending positions of the view.

This can also be replicated by moving your mouse really slowly.

ifreund:
Fixes #255

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -43,9 +43,21 @@ const XwaylandUnmanaged = @import("XwaylandUnmanaged.zig");
 const Mode = union(enum) {
     passthrough: void,
     down: *View,
-    move: *View,
+    move: struct {
+        view: *View,
+        /// View coordinates are stored as i32s as they are in logical pixels.
+        /// However, it is possible to move the cursor by a fraction of a
+        /// logical pixel and this happens in practice with low dpi, high
+        /// polling rate mice. Therefore we must accumulate the current
+        /// fractional offset of the mouse to avoid rounding down tiny
+        /// motions to 0.
+        delta_x: f64 = 0,
+        delta_y: f64 = 0,
+    },
     resize: struct {
         view: *View,
+        delta_x: f64 = 0,
+        delta_y: f64 = 0,
         /// Offset from the lower right corner of the view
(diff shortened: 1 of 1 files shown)


river/Cursor.zig
```diff
@@ -768,22 +780,33 @@ fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64,
                 self.wlr_cursor.y - @intToFloat(f64, output_box.y + view.current.box.y - view.surface_box.y),
             );
         },
-        .move => |view| {
-            view.move(@floatToInt(i32, delta_x), @floatToInt(i32, delta_y));
+        .move => |*data| {
+            dx += data.delta_x;
+            dy += data.delta_y;
+            data.delta_x = dx - @round(dx);
+            data.delta_y = dy - @round(dy);
```
Bonicgamer, line 787: We need to use trunc here instead since round can round up.
For instance with using round, if our "new" dx value were 0.9, it would round to 1.0 which would mean data's delta_x is going to be -0.1. If dx was 0.9 again, our "new" dx value would be 0.8 when it should be 1.8 (0.9 + 0.9 = 1.8 just in case my explanation wasn't clear).

```suggestion
            data.delta_x = dx - @trunc(dx);
            data.delta_y = dy - @trunc(dy);
```

Bonicgamer (changes requested):
Thanks for trying to fix this issue I was having! Most of it seems to work, but there was a slight thing that needs to be fix which I explained in the one of the comments/suggestions.
With the small tweak, it works perfectly for me!
This is my first time using the Github code review thing so I hope it works nice.

river/Cursor.zig
ifreund: Great catch, the semantics of `@trunc()` are indeed what I intended here!

PR #461, opened 2021-11-01
completions/fish: simplify

FollieHiyuki:
We can all agree that the 1st function looks really ugly :)

So I made it more simple to read.


ifreund (approved):
Looks like a nice cleanup, thanks!

PR #465, opened 2021-11-08
fix wlroots link in README.md

wannesdemeyer:
The GitHub repo is archived and redirects to [https://gitlab.freedesktop.org/wlroots/wlroots](https://gitlab.freedesktop.org/wlroots/wlroots)


ifreund (approved):
Thanks!

PR #473, opened 2021-11-21
Implement directional view focusing and swapping

Leon-Plickat:
An often requested feature, so I decided to go ahead and implement it.

This is a nice temporary solution to make layouts more flexible until we allow layout gens to control focus themselves, or maybe this even makes that unnecessary.

Either way, this makes a lot more workflows possible with river for comparatively little code.


Leon-Plickat:
To make this work I used the `util.safeIntDownCast()` function I originally came up with for #399. I feel like this can be done better with `@truncate()`; I'll look into that later.

Leon-Plickat:
This commit also introduces the `Vector` type, which I think will also allow me to refactor #399 a little bit.

niksingh710:
is this PR under review ? will this be added or working on other ways to implement it?

Leon-Plickat:
It's not under review. The code is a bit outdated; Should be an easy update, however I am not actively working on this right now.

The general idea is to make river extensible enough that this PR isn't necessary anymore, however that is pretty far off right now. So I am not sure what the fate of this PR is. I myself don't need it, I just hacked it together out of fun, so I am not actively engaged with getting this into river.

If this is a feature that interests you, I recommend pleading your case in IRC. If the feature gets a go, I'll pick up this PR again.

Leon-Plickat:
Ok, I resurrected the PR since I want this now.

river/command/view_operations.zig
```diff
@@ -0,0 +1,132 @@
...
+
+    if (try getTarget(seat, args[1], true)) |target| {
+        assert(!target.pending.float);
+        assert(!target.pending.fullscreen);
+        seat.focused.view.pending_wm_stack_link.swapWith(&target.pending_wm_stack_link);
+        server.root.applyPending();
+    }
+}
+
+fn getTarget(seat: *Seat, direction_str: []const u8, layout_only: bool) !?*View {
```
ifreund, line 64: layout_only could be an `enum {all, skip_float}` or similar to make the callsites a bit more readable.

I don't really recall why we have different behavior for the swap command in the first place...

river/command/view_operations.zig
```diff
@@ -0,0 +1,132 @@
...
+    if (seat.focused != .view) return null;
+    if (seat.focused.view.pending.fullscreen) return null;
+    if (layout_only and seat.focused.view.pending.float) return null;
+    const output = seat.focused_output orelse return null;
+
+    // If no currently no view is focused, just focus the first in the stack.
```
ifreund, line 70:
```suggestion
    // If no view is currently focused, focus the first in the stack.
```

river/Vector.zig
```diff
@@ -0,0 +1,62 @@
...
+
+const Vector = @This();
+
+x: i32,
+y: i32,
+
+pub fn positionOfWlrBox(box: wlr.Box) Vector {
+    return .{
+        .x = box.x + @divFloor(@intCast(u31, box.width), 2),
+        .y = box.y + @divFloor(@intCast(u31, box.height), 2),
```
ifreund, line 30: These `@intCast()`s are unnecessary.

river/Vector.zig
```diff
@@ -0,0 +1,62 @@
...
+    if ((math.absInt(self.y) catch return null) > (math.absInt(self.x) catch return null)) {
+        // Careful: We are operating in a Y-inverted coordinate system.
+        return if (self.y > 0) .down else .up;
+    } else {
+        return if (self.x > 0) .right else .left;
+    }
+}
+
+/// Returns the length of the vector.
+pub fn getLength(self: Vector) usize {
```
ifreund, line 56: I'd call these functions simply `direction()` and `length()`, the "get" adds no information here.

river/Vector.zig
```diff
@@ -0,0 +1,62 @@
...
+    const x = math.absInt(self.x) catch math.maxInt(i32);
+    const y = math.absInt(self.y) catch math.maxInt(i32);
+    const x_pow_2 = @intCast(usize, math.pow(i32, x, 2));
+    const y_pow_2 = @intCast(usize, math.pow(i32, y, 2));
+    return math.sqrt(x_pow_2 + y_pow_2);
+}
```
ifreund, line 62: Could this not be simplified to the following? I don't really see a need to return a `usize` either...
```zig
return math.sqrt((x *| x) +| (y *| y));
```

doc/riverctl.1.scd
```diff
@@ -35,8 +35,9 @@ over the Wayland protocol.
 	Focus the next or previous output, the closest output in any direction
 	or an output by name.
 
-*focus-view* *next*|*previous*
-	Focus the next or previous view in the stack.
+*focus-view* *next*|*previous*|*up*|*right*|*down*|*left*
```
ifreund, line 38: For consistency with the `move` command:
```suggestion
*focus-view* *next*|*previous*|*up*|*down*|*left*|*right*
```

doc/riverctl.1.scd
```diff
@@ -62,9 +63,9 @@ over the Wayland protocol.
 	*spawn* only takes a single argument. To spawn a command taking
 	multiple arguments, wrapping the command in quotes is recommended.
 
-*swap* *next*|*previous*
-	Swap the focused view with the next/previous visible non-floating
-	view. If the first/last view in the stack is focused, wrap.
+*swap* *next*|*previous*|*up*|*right*|*down*|*left*
```
ifreund, line 66: For consistency with the move command:
```suggestion
*swap* *next*|*previous*|*up*|*down*|*left*|*right*
```

river/Vector.zig
```diff
@@ -0,0 +1,62 @@
...
+const std = @import("std");
+const math = std.math;
+const wlr = @import("wlroots");
+
+const Vector = @This();
+
+x: i32,
+y: i32,
+
+pub fn positionOfWlrBox(box: wlr.Box) Vector {
```
ifreund, line 27: We no longer use non-wlr boxes.
```suggestion
pub fn positionOfBox(box: wlr.Box) Vector {
```

ifreund (commented):
Looks pretty good, just a bit of cleanup then this can be merged :)

river/command/view_operations.zig
Leon-Plickat: I think we came to the conclusion that swapping floating views is weird  and leads to unexpected UX especially when trying to swap a floating with a tiled view or the other way araound.

river/Vector.zig
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+        return if (self.x > 0) .right else .left;
+    }
+}
+
+/// Returns the length of the vector.
+pub fn length(self: Vector) u31 {
+    const x = @intCast(u31, math.absInt(self.x) catch math.maxInt(i32));
+    const y = @intCast(u31, math.absInt(self.y) catch math.maxInt(i32));
+    return math.sqrt((x *| x) +| (y *| y));
```
ifreund, line 59: The first 2 lines here are unnecessary I believe, `self.x *| self.x` will always be positive and handles the case of the minimum integer value correctly.

Leon-Plickat:
```
/usr/lib/zig/std/math/sqrt.zig:29:24: error: sqrt not implemented for signed integers 
            .signed => @compileError("sqrt not implemented for signed integers"), 
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ 
```

Leon-Plickat: Mathematically you are right, but the typesystem disagrees.

Leon-Plickat: Oh, unless you mean it with still having an intCast inside the sqrt, that works.

ifreund: Ah, I didn't realize `sqrt` doesn't support signed integers. Doing the `@intCast()` after the multiplication is cleaner though, thanks!

ifreund:
The alpine CI failure looks unrelated to this PR but serious nonetheless. At a guess, alpine has recently updated to a newer version of musl which has has caused incompatibility with the Zig 0.10 standard library...

Anyhow, the only thing blocking this PR from being merged is me finding time to build and test it briefly locally.

novakne:
FWIW I've been running it all day without any problem

river/command/view_operations.zig
```diff
@@ -0,0 +1,133 @@
...
+    // If no currently view is focused, focus the first in the stack.
+    if (seat.focused != .view) {
+        var it = output.pending.wm_stack.iterator(.forward);
+        return it.next();
+    }
+
+    // Logical direction, based on the view stack.
+    if (std.meta.stringToEnum(Direction, direction_str)) |direction| {
+        switch (direction) {
+            inline else => |dir| {
```
Leon-Plickat, line 80: Looking over this again, I don't see why there is a switch with an `inline else` here. I copied that from the current implementation, but actually looking at and thinking about this code, the only place `dir` is used is in the inner switch to get `it_dir`, so I'd be genuinely surprised if the outer switch actually has any benefits w.r.t. to machine code generation.

ifreund: Doesn't the `iterator()` function used in this branch take a comptime value?

Leon-Plickat: You're right, it does.

ifreund (approved):
Thanks for working on this!

Issue #448, opened 2021-10-08
[feature request] Could `focus-follows-cursor` detect output change?


nwg-piotr:
Hi,

I first installed river 2 days ago, to test how [my stuff](https://github.com/nwg-piotr/nwg-shell) behaves here, and I must say you're doing a great job!

As I use my mouse quite heavily on a triple-headed setup, `focus-follows-cursor normal` fits my needs best. It would be even more convenient if pointing an empty output (with no surface on it) changed focus as well. Otherwise I need to put the mouse down and switch outputs using the keyboard. Could it be changed?

nwg-piotr:
BTW: if there existed a command to focus outputs by name (not just `left` | `right` | `next` | `previous`), the problem could be temporarily resolved with a script...

keshto:
Would be nice to have a click option for changing the focus of the output, as I have the tendency to use my other monitor to put my cursor out of the way of the working output.

Leon-Plickat:
@keshto outputs can already be focused via mouse click. I don't think that patch made it into the last release though.

keshto:
@Leon-Plickat that explains that since I am using 0.1.2. I guess I could build from master if it is stable enough and test or just wait and open an issue if I don't see it in next release. Thanks for the info :+1:

edrex:
I'd like this too, and would be willing to implement it as a first patch.

Leon-Plickat:
@edrex you may want to wait with that patch until the scene-graph port is complete.

larstiq:
> @edrex you may want to wait with that patch until the scene-graph port is complete.

I didn't want to wait so implemented https://github.com/riverwm/river/pull/956/

PR #475, opened 2021-11-22
Accept specific output name for `focus-output` and `send-to-ouput` commands

pmkap:
This allows to focus a specific output or send the view to a specific output.

With riverctl it looks like this:
`send-to-output next|previous|up|right|down|left|name`
`focus-output next|previous|up|right|down|left|name`

After matching the argument for next/previous and up/right/etc., the output names itself are checked for a match.

Exmaples:
`riverctl focus-output WL-1`
`riverctl send-to-output HDMI-A-1`


nwg-piotr:
Works excellent, and should allow to write a script to workaround #448.

doc/riverctl.1.scd
```diff
@@ -53,6 +53,9 @@ over the Wayland protocol.
 *focus-output* *next*|*previous*|*up*|*right*|*down*|*left*
 	Focus the next or previous output or the closest output in any direction.
 
+*focus-output-name* _output_
+	Focus the a specific output by its name.
```
and3rson, line 57: Seems like a typo: "the a"

Grawp:
@pmkap Would you consider updating the PR so it builds with Zig 0.9.0? Thanks.
Btw. the PR itself is very helpful. I had been using it as patch for River builds in NixOS till 0.9.0 broke it.

pmkap:
@Grawp Sure, I'll update this somewhen in the next days.

Also, I'm thinking maybe it would be nicer not to have an extra command but just something like
`focus-output next|previous|output-name` and resp.
`send-to-output next|previous|output-name`

Or is there anything that speaks against this?

doc/riverctl.1.scd
```diff
@@ -50,8 +50,9 @@ over the Wayland protocol.
 	Remove an app-id or title from the float filter list. Note that this
 	affects only new views, not already existing ones.
 
-*focus-output* *next*|*previous*|*up*|*right*|*down*|*left*
-	Focus the next or previous output or the closest output in any direction.
+*focus-output* *next*|*previous*|*up*|*right*|*down*|*left*|*name*
```
Leon-Plickat, line 53: Should be `_name_`.

doc/riverctl.1.scd
```diff
@@ -68,9 +69,9 @@ over the Wayland protocol.
 	Snap the focused view to the specified screen edge. The view will
 	be set to floating.
 
-*send-to-output* *next*|*previous*|*up*|*right*|*down*|*left*
-	Send the focused view to the next or previous output or the closest
-	output in any direction.
+*send-to-output* *next*|*previous*|*up*|*right*|*down*|*left*|*name*
```
Leon-Plickat, line 72: same here.

Leon-Plickat (changes requested):

doc/riverctl.1.scd
pmkap: Ah yeah I can see why.

Leon-Plickat (approved):

pmkap:
Currently, when having only one output and using e.g `riverctl send-to-output next`, the view gets pushed on top resp. bottom of the stack. Also with this proposed PR the user can send the view to the current output. I wonder if this is a bit unintuitive and we should intercept this case: `seat.focused.view.output == destination_output`.

ifreund:
> Currently, when having only one output and using e.g `riverctl send-to-output next`, the view gets pushed on top resp. bottom of the stack. Also with this proposed PR the user can send the view to the current output. I wonder if this is a bit unintuitive and we should intercept this case: `seat.focused.view.output == destination_output`.

Yeah, this should do absolutely nothing when targeting the current output.

river/command/output.zig
```diff
@@ -93,6 +95,13 @@ fn getOutput(seat: *Seat, str: []const u8) !?*Output {
         ) orelse return null;
         return @intToPtr(*Output, wlr_output.data);
     } else {
+        // Check if an output matches by name
+        var it = server.root.outputs.first;
+        while (it) |node| : (it = node.next) {
+            if (std.mem.eql(u8, std.mem.sliceTo(node.data.wlr_output.name, 0), str)) {
```
ifreund, line 101: We actually don't need `sliceTo()` here as `wlr_output.name` is now a 0 terminated pointer not an array.
```suggestion
            if (std.mem.eql(u8, std.mem.span(node.data.wlr_output.name), str)) {
```

ifreund: And looks like I forgot to change the doc comment in zig-wlroots...

pmkap: Ah nice!

ifreund (approved):
LGTM, Thanks!

PR #476, opened 2021-11-22
Cursor: fix pointer input after drags in gtk3 clients

ifreund:
It seems that either GTK mishandles the case where wl_pointer.enter is
sent before wl_data_source.dnd_finished after performing a dnd
or that sending the events in that order is a protocol error on the
compositor's part.

Edit: the above analysis is at least partially incorrect.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -33,6 +33,7 @@ const util = @import("util.zig");
 
 const Box = @import("Box.zig");
 const Config = @import("Config.zig");
+const DragIcon = @import("DragIcon.zig");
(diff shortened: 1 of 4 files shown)


ifreund:
Hmm, I think this is actually better resolved with a minor change to wlroots: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3364

ifreund:
My wlroots patch above has been merged, so this bug will go away when river updates to the next wlroots release. Until then, I'll implement a workaround. The patch in this PR misses the point though and the workaround should be simpler.

PR #477, opened 2021-11-24
Implement basic touch support

Danacus:
I've implemented basic touch (finger) input by adding another another invisible cursor to the seat. This cursor never moves as it is not needed unless we wanted to emulate a mouse pointer for surfaces that don't support touch (all modern GTK3 and Qt5 programs do support touch). Events are propagated to clients and don't interact with the cursor.

Whenever `handleTouchDown` is called, the mouse cursor will be hidden until the mouse is moved.

I also changed the `surfaceAt` function in `Cursor` by adding a `surfaceAtPosition` function such that it can be used without a `Cursor`.


Leon-Plickat:
> I've implemented basic touch (finger) input by adding another another invisible cursor to the seat. 

Why? What advantages does this have over just sending touch events.

river/Touch.zig
```diff
@@ -0,0 +1,109 @@
...
+
+const Self = @This();
+
+const std = @import("std");
+const wlr = @import("wlroots");
+const wayland = @import("wayland");
+const wl = wayland.server.wl;
+const zwlr = wayland.server.zwlr;
+
+const c = @import("c.zig");
```
Leon-Plickat, line 26: You don't use anything from `c.zig`, so you do not need to import it.

river/Touch.zig
```diff
@@ -0,0 +1,109 @@
...
+const server = &@import("main.zig").server;
+const util = @import("util.zig");
```
Leon-Plickat, line 28: You also don't use anything from `util.zig`.

Leon-Plickat:
Have you tested this with multi touch? I am pretty sure that will break the integration with the invisible cursor.

Leon-Plickat:
Have you tested drag-and-drop between clients? For example, does this correctly work when you try to move a file from one filemanager view to another (nautilus should implement touch, if you need a test client)?

river/Touch.zig
```diff
@@ -0,0 +1,109 @@
...
+
+    var lx: f64 = undefined;
+    var ly: f64 = undefined;
+    self.wlr_cursor.absoluteToLayoutCoords(event.device, event.x, event.y, &lx, &ly);
+
+    if (Cursor.surfaceAtPosition(lx, ly)) |result| {
+        _ = self.seat.wlr_seat.touchNotifyDown(result.surface, event.time_msec, event.touch_id, result.sx, result.sy);
+    }
+
+    self.seat.cursor.hide();
```
Leon-Plickat, line 108: Why hide the main mouse cursor? It has nothing to do with the touch operation.

river/Touch.zig
```diff
@@ -0,0 +1,109 @@
...
+    event: *wlr.Touch.event.Up,
+) void {
+    const self = @fieldParentPtr(Self, "touch_up", listener);
+
+    self.seat.handleActivity();
+
+    _ = self.seat.wlr_seat.touchNotifyUp(event.time_msec, event.touch_id);
+}
+
+fn handleTouchDown(
```
Leon-Plickat, line 92: If this function is called and the coordinates are on an unfocused view, it should be focused by the seat. Unless I am missing something, that is currently not the case.

Danacus:
> > I've implemented basic touch (finger) input by adding another another invisible cursor to the seat.
> 
> ~Why? What advantages does this have over just sending touch events?~
> 
> ~I don't think this patch should touch `Cursor.zig` at all.~
> 
> ~Wayland allows clients to support touch inputs directly without going through cursor emulation and all relevant clients do so. Given that adding touch support to clients is almost trivial (I have done it myself) I don't see any reason to support clients that do not implement touch.~
> 
> Nevermind, sway does it too. I still think it's a bit of a hack though.

The thing about needing a cursor for touch seems to be something odd about wlroots. I didn't want a `wlr_cursor, because we don't need one for touch`, but that's where the events are defined, so I don't think there's another way.

The reason I had to touch `Cursor.zig` is so I wouldn't have to copy the `surfaceAt` code. Perhaps this code could be moved somewhere else?

Danacus:
> Have you tested this with multiple touchpoints?

Yes, I can zoom with 2 fingers. More than 2 touchpoints should work too.

river/Touch.zig
Danacus: It's something that annoys me about GNOME. When switching my laptop to tablet mode, it's a bit annoying to have a cursor in the middle of my screen that I can't move in any way because I can't use the mouse. Sway and KDE do hide the mouse cursor when using touch, but I understand that it seems a bit odd at first.

Danacus:
> Have you tested drag-and-drop between clients? For example, does this correctly work when you try to move a file from one filemanager view to another (nautilus should implement touch, if you need a test client)?

I can't really figure out what's up with drag and drop. It works on GTK apps like nautilus on sway, but not between windows. On GNOME it just doesn't work at all surprisingly. This might have something to do with sway actually moving the mouse cursor with touch (which feels like a hack). I wonder if touch drag and drop is even a thing.

river/Touch.zig
```diff
@@ -0,0 +1,117 @@
...
+    self.seat.handleActivity();
+
+    var lx: f64 = undefined;
+    var ly: f64 = undefined;
+    self.wlr_cursor.absoluteToLayoutCoords(event.device, event.x, event.y, &lx, &ly);
+
+    if (Cursor.surfaceAtPosition(lx, ly)) |result| {
+        switch (result.parent) {
+            .view => |view| {
+                self.seat.focusOutput(view.output);
```
Danacus, line 106: This should ideally only happen if the currently focused view is different from this view. However, including such statement causes problems where it's not possible to focus the surface which has the mouse pointer on it, because it thinks that view is already focused even though it's not. I'm not sure how to deal with this and my current solution is not great.

Leon-Plickat: You could try to have a new `surfaceAtPosition()` function in `Touch.zig`.

river/Touch.zig
```diff
@@ -102,6 +101,15 @@ fn handleTouchDown(
...
 
     if (Cursor.surfaceAtPosition(lx, ly)) |result| {
+        switch (result.parent) {
+            .view => |view| {
+                self.seat.focusOutput(view.output);
+                self.seat.focus(view);
+                server.root.startTransaction();
+            },
+            .layer_surface => {},
+            .xwayland_unmanaged => assert(build_options.xwayland),
```
Leon-Plickat, line 111: This assert is not needed. I don't see how it will ever be reached unless XWayland support is enabled, in which case it will never fail.

river/Touch.zig
Danacus: The `surfaceAtPosition` from `Cursor` works fine and adding one in `Touch.zig` would create a lot of code duplication which I wanted to avoid. The problem here is that I'm setting focus to the current surface without checking if that surface is already focused. Adding such check would cause the aforementioned problems.

river/Touch.zig
```diff
@@ -102,6 +101,15 @@ fn handleTouchDown(
     self.wlr_cursor.absoluteToLayoutCoords(event.device, event.x, event.y, &lx, &ly);
 
     if (Cursor.surfaceAtPosition(lx, ly)) |result| {
+        switch (result.parent) {
+            .view => |view| {
+                self.seat.focusOutput(view.output);
```
Leon-Plickat, line 106: `self.seat.focus(view)` will already focus the output the view is on if it is unfocused, so you do not need to do it manually here.

Leon-Plickat: You probably copied from `checkFocusFollowCursor()`, right? If you like, you can add a commit removing it there as well.

Danacus: I did copy it from there and I did also notice that `seat.focus` sets the output. I'll fix it in both places.

river/Touch.zig
Danacus: This is also just here because I copied it from somewhere else and wasn't sure what it's supposed to do so I left it.

river/Touch.zig
Leon-Plickat: Can't you just check the focus of the seat?

```zig
    if (Cursor.surfaceAtPosition(lx, ly)) |result| {
        switch (result.parent) {
            .view => |view| {
                 if (seat.focused == .view and seat.focused.view == view) {
                    self.seat.focusOutput(view.output);
                }
            }
```

Danacus: I was having problems with that earlier, but now it works fine.

river/Touch.zig
Danacus: I think the assert is just used to assert that that case is never reached. I don't think it's needed in `checkFocusFollowsCursor` either, but it's perhaps just a "coding style" thing?

river/Cursor.zig
```diff
@@ -444,6 +449,13 @@ const SurfaceAtResult = struct {
 pub fn surfaceAt(self: Self) ?SurfaceAtResult {
     const lx = self.wlr_cursor.x;
     const ly = self.wlr_cursor.y;
+    return surfaceAtPosition(lx, ly);
+}
+
+/// Find the surface on the given position if any, and return information about that
+/// surface and the given position in surface local coords.
+/// This function must be kept in sync with the rendering order in render.zig.
+pub fn surfaceAtPosition(lx: f64, ly: f64) ?SurfaceAtResult {
```
Leon-Plickat, line 458: Since this is now also used elsewhere in a different context, I am not really a fan of having it in `Cursor.zig`, seems a bit leaky abstraction-wise.

river/Seat.zig
```diff
@@ -436,6 +446,11 @@ fn addPointer(self: Self, device: *wlr.InputDevice) void {
     self.cursor.wlr_cursor.attachInputDevice(device);
 }
 
+fn addTouch(self: *Self, device: *wlr.InputDevice) void {
+    self.touch.wlr_cursor.attachInputDevice(device);
```
Leon-Plickat, line 450: I quickly checked Sway again and I think they do not have a separate cursor for touch. What are the advantages of having a separate cursor instead of just using the one that is already there?

Leon-Plickat (commented):
Probably last review from me.

I can't test it, but to me it looks sound overall.

Oh, and you probably want to squash all your "Touch: ..." commits. While git commits are supposed to be incremental, I'd consider implementing touch support a single such increment that needs no splitting.

river/Cursor.zig
Danacus: I also considered moving it somewhere else, but wasn't sure where.

river/Seat.zig
Leon-Plickat: And, since touch operations seem to be cursor operations in wlroots, why not have them in `Cursor.zig`?

Leon-Plickat:
Yeah, thinking about it I believe the touch handlers should be in `Cursor.zig` as well, since they are handlers for cursor events. In that case you can also keep `surfaceAtPosition` where it is now.

Danacus:
I guess that makes sense to follow the design of wlroots even though it feels a bit odd to me that you would need a cursor for touch. In Sway they also move the cursor on touch input, but I don't think this makes a lot of sense.

river/Seat.zig
Danacus: I was using the mouse cursor at first but changed that when you mentioned that touch is not supposed to move the cursor. However, since wlroots needs a cursor for the events, I needed to add a cursor anyway. So perhaps it makes sense to use the already existing mouse cursor since we only really need it for events.

Danacus:
Btw, thanks for taking your time to review my PR, I really appreciate it!

river/Touch.zig
Leon-Plickat:
```IRC
12:47   leon-p: ifreund: why is there an assert(build_options.xwayland) in checkFocusFollowsCursor()? If xwayland support is not enabled at compile time, that branch will never be reached anyway
14:05  ifreund: leon-p: exactly, if build_options.xwayland is false we want to mark the branch as unreachable
```

Leon-Plickat (approved):

river/Cursor.zig
```diff
@@ -192,6 +203,11 @@ pub fn setTheme(self: *Self, theme: ?[*:0]const u8, _size: ?u32) !void {
     }
 }
 
+/// Hides the cursor.
+pub fn hide(self: *Self) void {
+    self.wlr_cursor.setImage(0, 0, 0, 0, 0, 0, 0);
+}
+
```
ifreund, line 210: I think we would need to save whatever the current cursor image is to be able to properly restore it on the next pointer input. Cursor images may be set by clients.

Leaving this feature for a second PR would likely get this one merged more quickly.

river/Cursor.zig
```diff
@@ -444,6 +517,13 @@ const SurfaceAtResult = struct {
 pub fn surfaceAt(self: Self) ?SurfaceAtResult {
     const lx = self.wlr_cursor.x;
     const ly = self.wlr_cursor.y;
+    return surfaceAtPosition(lx, ly);
+}
+
+/// Find the surface on the given position if any, and return information about that
```
ifreund, line 523:
```suggestion
/// Find the surface at the given position if any, and return information about that
```

river/Cursor.zig
```diff
@@ -444,6 +517,13 @@ const SurfaceAtResult = struct {
 pub fn surfaceAt(self: Self) ?SurfaceAtResult {
     const lx = self.wlr_cursor.x;
     const ly = self.wlr_cursor.y;
+    return surfaceAtPosition(lx, ly);
```
ifreund, line 520:
```suggestion
    return surfaceAtPosition(self.wlr_cursor.y, self.wlr_cursor.x);
```

ifreund (changes requested):
Thanks for working on this!

> I guess it makes sense to follow the design of wlroots even though it feels a bit odd to me that you would need a cursor for touch. In Sway they also move the cursor on touch input, but I don't think this makes a lot of sense.

You don't need a cursor for touch, the events of the `wlr_touch` struct may be listened to directly. One reason to go through a wlr_cursor would be if we wanted to simulate pointer input on surfaces that don't subscribe to touch events (like sway does). Nobody's asked for this feature yet and I don't even have any touch hardware though.

Another reason to go through the wlr_cursor would be to support mapping input devices to/from outputs/regions, as wlr_cursor does the required conversions for us. I imagine that this is something we will eventually want to support.

Anyhow, I think putting this code in `Cursor.zig` is fine.

river/Seat.zig
```diff
@@ -436,6 +441,11 @@ fn addPointer(self: Self, device: *wlr.InputDevice) void {
     self.cursor.wlr_cursor.attachInputDevice(device);
 }
 
+fn addTouch(self: *Self, device: *wlr.InputDevice) void {
+    self.cursor.wlr_cursor.attachInputDevice(device);
+    self.has_touch = true;
```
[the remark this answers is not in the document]
Danacus: I'm not sure how to handle the destroy signal.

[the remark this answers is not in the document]
ifreund: The destroy signal on `wlr.InputDevice` is the one I was talking about here.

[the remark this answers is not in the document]
Danacus: Should `Seat.zig` handle this or should it be handled in a `Touch.zig` file similar to `Keyboard.zig`? Also, when a device is destroyed, should the capabilities of the seat be updated? Shouldn't this also be the case for when a keyboard is destroyed? I'm not sure how or if this gets done right now.

theSuess:
I've locally rebased these changes onto master and can confirm that touch works great on my Surface Pro 4! Would be awesome to have this merged

Danacus:
There are still some things that need to be finished, but I've been busy with university. I might finish this later.

Danacus:
It's mostly just some details as discussed in the threads on this issue. Functionally everything is done, it just needs to be finished such that maintainers accept the PR.

river/Seat.zig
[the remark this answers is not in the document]
ifreund: Hey, sorry it's taken so long to get back to you. It looks like we don't ever update the capabilities on removal for keybords/pointers currently. I think the right way to do this would be adding a `removeDevice()` function to `Seat.zig` and call that from `InputDevice.handleDestroy()` in `InputManager.zig`.

Anyhow, since it turns out we don't do this for *any* input devices currently feel free to leave out of this PR, I can implement that when I find the time to merge this if you like.

Danacus:
This is indeed a non-trivial addition. I once attempted this, but lost time and motivation.

These are the changes I proposed for `zig-wlroots`: https://github.com/swaywm/zig-wlroots/pull/35

And a very crude attempt to get pen input working: https://github.com/Danacus/river/tree/tablet_v2

The main struggle I had was that the implementation of `tablet-unstable-v2` in `wlroots` felt a bit unintuitive and confusing to me.

ifreund:
I've recently refactored river's input device handling to make touch support cleaner to implement. Due to that and other changes to the cursor code I decided to continue this effort in a new PR: https://github.com/riverwm/river/pull/632

Any testing would be highly appreciated, I still don't have any touch hardware.

ifreund:
Thanks for all your work here @Danacus, this feature has been merged in #632!

PR #481, opened 2021-11-29
layer-shell: tighten up handling of large margins

### LINKED ISSUE #468: Layer surface vertical wrap around
I'm seeing what I consider to be weird behavior when mapping many layer surfaces, at increasing y-coordinates. A picture says, perhaps not all, but a lot:
many-layer-surfaces

This is fnott showing notifications. I've triggered them in order ("one", "two", "three", ...). Note how the bottom one is "ten", and then there's **one** final notification at the top, "eleven". It looks like it has wrapped around.

After "eleven", I also create the notifications "twelve" and "thirteen". These are closed (`layer_surface.close()`) before being mapped. I.e. river rejects their first configuring commit.

I don't mind the surfaces being closed immediately. That's what I would expect to happen anyway. But the one that wraps around doesn't look intentional?

(note that the latest release of fnott isn't great at handling this situation - running latest master is advised if you want to use fnott for testing).

River version: 0.2.0-dev-f3475c5

ifreund:
This should fix #468.

### DIFF
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -308,23 +308,15 @@ fn arrangeLayer(
 
         // Horizontal alignment
         const horizontal_margin_size = current_state.margin.left + current_state.margin.right;
-        if (horizontal_margin_size >= bounds.width) {
-            // TODO find a better solution
-            // We currently have not reached a conclusion on how to gracefully
-            // handle this case yet, so we just close the surface. That will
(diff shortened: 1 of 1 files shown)


novakne:
Looks like it works as expected now https://0x0.st/-hot.mp4

PR #484, opened 2021-12-10
improve cursor output focusing on button press

Leon-Plickat:
This patch allows to focus outputs by clicking on the empty background and by
clicking on layer surfaces without keyboard interactivity. This makes it
possible to use the cursor to focus outputs with no visible views.

This also fixes problems with pointer interactive layer surfaces (for example
launchers and docks) on non-focused outputs.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -255,16 +255,25 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
                 }
             },
             .layer_surface => |layer_surface| {
+                self.seat.focusOutput(layer_surface.output);
                 // If a keyboard inteactive layer surface has been clicked on,
                 // give it keyboard focus.
                 if (layer_surface.wlr_layer_surface.current.keyboard_interactive == .exclusive) {
-                    self.seat.focusOutput(layer_surface.output);
                     self.seat.setFocusRaw(.{ .layer = layer_surface });
+                } else {
+                    self.seat.focus(null);
                 }
+                server.root.startTransaction();
             },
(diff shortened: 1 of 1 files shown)


river/Cursor.zig
```diff
@@ -255,16 +255,24 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
                 }
             },
             .layer_surface => |layer_surface| {
+                self.seat.focusOutput(layer_surface.output);
+                self.seat.focus(null);
```
ifreund, line 259: I think we should move this line to the else branch of the if statement below.

river/Cursor.zig
```diff
@@ -255,16 +255,24 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
...
-                    self.seat.focusOutput(layer_surface.output);
                     self.seat.setFocusRaw(.{ .layer = layer_surface });
                 }
+                server.root.startTransaction();
             },
             .xwayland_unmanaged => assert(build_options.xwayland),
         }
         _ = self.seat.wlr_seat.pointerNotifyButton(event.time_msec, event.button, event.state);
+    } else if (server.root.output_layout.outputAt(self.wlr_cursor.x, self.wlr_cursor.y)) |wlr_output| {
+        // If the user clicked on an empty output, focus it.
```
ifreund, line 271:
```suggestion
        // If the user clicked on empty space of an output, focus it.
```

ifreund (commented):
This is a great feature, thanks!

ifreund (approved):
LGTM, Thanks!

PR #495, opened 2021-12-25
Install session desktop file

kmicklas:
Display managers like GDM use these desktop files to show a session selector in the greeter.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -157,6 +157,8 @@ pub fn build(b: *zbs.Builder) !void {
         b.installFile(file, "share/pkgconfig/river-protocols.pc");
     }
 
+    b.installFile("river.desktop", "share/wayland-sessions/river.desktop");
+
     if (man_pages) {
         const scdoc_step = ScdocStep.create(b);
         try scdoc_step.install();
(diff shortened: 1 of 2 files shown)


ifreund (commented):
Hi, I appreciate the PR but am not sure I want to merge this.

I personally don't use any "display manager"/"session manager" and just start river from a tty. From my perspective, this `.desktop` file seems like a piece of configuration for your specific display manger, not something universally useful that river should distribute. For example, [greetd](https://sr.ht/~kennylevinsen/greetd/) is a session manager that does not require any .desktop file.

Is GDM's usage of .desktop files for wayland sessions even specified anywhere? As I understand it they are intended for applications that run inside a session not for wayland compositors themselves.

If distro maintainers want to patch in a .desktop file for river to better integrate with their distributions set up that's perfectly fine by me.

kmicklas:
I'm not sure this is properly documented for GDM or elsewhere.

As you say, it's trivial for distros to add it, so feel free to close this if you'd rather take that approach.

My only real reason for creating this was that Sway does it (https://github.com/swaywm/sway/issues/100), and this caused it to work out of the box in GDM on my distro (NixOS) while River did not.

ifreund:
Ok, well let's not ship GDM specific configuration files then, especially since they are trivial for users/distribution packagers to add if desired.

PR #501, opened 2022-01-04
Xwayland: Handle minimize request

### LINKED ISSUE #500: Black screen on some Wine games after they lose focus
This is the same issue as https://github.com/swaywm/sway/issues/4324, which was resolved by https://github.com/swaywm/sway/pull/5566.

novakne:
Fix X11 clients getting stuck minimized, and displaying a black screen
after they lose focus.

Fixes #500

### DIFF
--- a/river/XwaylandView.zig
+++ b/river/XwaylandView.zig
@@ -48,6 +48,8 @@ set_title: wl.Listener(*wlr.XwaylandSurface) = wl.Listener(*wlr.XwaylandSurface)
 set_class: wl.Listener(*wlr.XwaylandSurface) = wl.Listener(*wlr.XwaylandSurface).init(handleSetClass),
 request_fullscreen: wl.Listener(*wlr.XwaylandSurface) =
     wl.Listener(*wlr.XwaylandSurface).init(handleRequestFullscreen),
+request_minimize: wl.Listener(*wlr.XwaylandSurface.event.Minimize) =
+    wl.Listener(*wlr.XwaylandSurface.event.Minimize).init(handleRequestMinimize),
 
 pub fn init(self: *Self, view: *View, xwayland_surface: *wlr.XwaylandSurface) void {
     self.* = .{ .view = view, .xwayland_surface = xwayland_surface };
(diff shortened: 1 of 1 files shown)


river/XwaylandView.zig
```diff
@@ -269,3 +280,12 @@ fn handleRequestFullscreen(listener: *wl.Listener(*wlr.XwaylandSurface), xwaylan
...
     self.view.applyPending();
 }
+
+fn handleRequestMinimize(
+    listener: *wl.Listener(*wlr.XwaylandSurface.event.Minimize),
+    event: *wlr.XwaylandSurface.event.Minimize,
+) void {
+    const self = @fieldParentPtr(Self, "request_minimize", listener);
+    self.view.pending.minimized = event.minimize;
+    self.view.applyPending();
```
ifreund, line 290: This should just call `xwayland_surface.setMinimized()` directly. This is an implementation detail of xwayland and should not leak out in to the rest of the code.

This also needs a comment explaining why we need to do this.

ifreund (changes requested):
Thanks for taking a look at this!

river/XwaylandView.zig
```diff
@@ -98,6 +103,10 @@ pub fn setFullscreen(self: Self, fullscreen: bool) void {
     self.xwayland_surface.setFullscreen(fullscreen);
 }
 
+pub fn setMinimized(self: Self, minimized: bool) void {
+    self.xwayland_surface.setMinimize(minimized);
+}
```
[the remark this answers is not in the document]
novakne: Whoops my bad I deleted the function that called this since it was useless and forgot to delete it

river/XwaylandView.zig
```diff
@@ -269,3 +276,14 @@ fn handleRequestFullscreen(listener: *wl.Listener(*wlr.XwaylandSurface), xwaylan
     self.view.pending.fullscreen = xwayland_surface.fullscreen;
     self.view.applyPending();
 }
+
+/// Some X11 clients get stuck in minimized state when they lose focus, which
+/// cause black screen, therefore we just disable it when a request is received
+/// as river doesn't support it.
+fn handleRequestMinimize(
```
ifreund, line 283: This comment and the code below do not match the findings and fix in the linked sway issue/PR...

ifreund (approved):
Thanks for the fix! I've just improved the comments slightly before pushing this to master.

ifreund:
It seems github doesn't like what I did so this doesn't like what I did so this doesn't say "merged." Fear not though, it was merged.

PR #504, opened 2022-01-07
Implement a scroll-factor command

mmskv:
This PR implements a `scroll-factor` input setting that allows to control scroll speed

See https://github.com/swaywm/sway/pull/3018:

> This PR adds a scroll_factor input command whose value is used as a multiplier for axis event deltas received from a device.  Put simply, this means that scrolling speed can now be adjusted, a configuration feature that libinput itself lacks. At least on my laptop's touchpad, such a change is sorely needed.
>
>For example, adding scroll_factor 0.5 under a device's input section will make scrolling half as fast.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -40,6 +40,8 @@ const View = @import("View.zig");
 const ViewStack = @import("view_stack.zig").ViewStack;
 const XwaylandUnmanaged = @import("XwaylandUnmanaged.zig");
 
+const InputConfig = @import("InputConfig.zig");
+
 const Mode = union(enum) {
     passthrough: void,
     down: struct {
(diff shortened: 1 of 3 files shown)


river/Cursor.zig
```diff
@@ -247,12 +249,14 @@ fn handleAxis(listener: *wl.Listener(*wlr.Pointer.event.Axis), event: *wlr.Point
 
     self.seat.handleActivity();
 
+    const scroll_factor = server.input_manager.input_configs.scroll_factor || 1.0;
```
mmskv, line 252: How do I query a device specific scroll factor value?

itshog:
Hey, any chance to revive this PR for 0.3 release? On my laptop scrolling with the touchpad is basically unusable with the default scroll factor of libinput. I know i could use something like [this](https://gitlab.com/warningnonpotablewater/libinput-config), but it feels kind of a hacky workaround, especially since other wayland compositors (Sway and Hyprland come to mind) implement a "native" solution.

Anyway, thanks for the amazing project!

Leon-Plickat:
This is a rather simple PR, so anyone could revive it if they wanted to.

Although I should add that unwanted scroll-behaviour is *usually* a client bug. A client should react different to high-res scroll devices than to "normal" scrolling.

ifreund:
This never got finished and is now quite out of date. Feel free to open a new PR to continue this work.

PR #505, opened 2022-01-08
contrib: add desktop file

NickHastings:
I recently made a river Debian package and included a desktop file to be installed in /usr/share/wayland-sessions. I noticed that the river-git AUR package does the same thing, So, to avoid duplicating work and to keep things consistent between distributions, I thought it would be worthwhile if upstream river shipped with a desktop file that all packagers could make use of. This desktop file in this branch is almost identical to the one from AUR river-git differing only in the capitalization for "River" and in the addition of a newline at the end of the file.

### DIFF
--- a/contrib/river.desktop
+++ b/contrib/river.desktop
@@ -1,0 +1,5 @@
+[Desktop Entry]
+Name=River
+Comment=A dynamic tiling Wayland compositor
+Exec=river
+Type=Application


novakne:
See #495 for why there are no plans to add it.

ifreund (approved):
Hmm, I think having this in the contrib directory and not officially supported is a good compromise actually. It won't be installed by `zig build` automatically but distributions wanting to use it can very simply install it manually.

Thanks!

NickHastings:
Excellent! That's one thing I can cross off the list of things that the "Debian High Command" don't like about the package I'm building!

FYI  https://mentors.debian.net/package/river/ and https://lintian.debian.org/tags/maintainer-desktop-entry

PR #506, opened 2022-01-08
command: Implement cycle-focused-tags

NickHastings:
I've been using river for a few days and have been very impressed. There seems to be just one feature that my fingers muscle memory (picked up from around 10 years of using awesome)  just won't forget.

This branch implements a function that lets users cycle through tags, either next or previous, with at keybinding. The configuration option "cycle-focused-tags" takes two arguments: the first being the direction next|previous, the second being the maximum number tags at which the cycling should wrap back to the first tag (or to the last tag from the first tag).

A sample from a users ~/.config/river/init could look like:

```
riverctl map normal Mod4+Shift Up cycle-focused-tags previous 9
riverctl map normal Mod4+Shift Down cycle-focused-tags next 9
```

This has been tested and seems to be working well even when multiple tags are selected. However, the code is somewhat ugly. This is the first time I've written anything in zig, and I'm sure I've done things in a less than ideal why. In particular, I found that I could not do the bitwise calculations how I would in other languages, since the RHS of the << operator seems to need to be known at compile time.

I'm sure I must have breached coding style conventions etc. Please let me know what I can do to make this code better.

PR #507, opened 2022-01-09
contrib: Add script to cycle through the focused tags

NickHastings:
As discussed in https://github.com/riverwm/river/pull/506 I've reimplemented the functionality to cycle through the focussed tags using a python script.

The script takes two arguments: the first being the direction next|previous, the second being the maximum number of tags at which the cycling should wrap back to the first tag (or to the last tag from the first tag).
The script can be called using spawn in the users init file. For example:
```
riverctl map normal Mod4 Up spawn "path-to/river/contrib/cycle-focused-tags previous 9"
riverctl map normal Mod4 Down spawn "path-to/river/contrib/cycle-focused-tags next 9"
```
This has been tested and seems to be working well even when multiple tags are selected.

### DIFF
--- a/contrib/cycle-focused-tags
+++ b/contrib/cycle-focused-tags
@@ -1,0 +1,185 @@
+#!/usr/bin/env python3
+
+import sys
+from pywayland.client import Display
+from pywayland.protocol.wayland import WlOutput
+from pywayland.protocol.wayland import WlSeat
+try:
+    from pywayland.protocol.river_control_unstable_v1 import ZriverControlV1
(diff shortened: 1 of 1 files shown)


NickHastings:
Sorry, history got messed up. I'll fix it soon.

NickHastings:
I spent my lunch break putting this script into a little [package](https://github.com/NickHastings/riverwm-utils). It's also on [PyPi](https://pypi.org/project/riverwm-utils), so can be installed with `python3 -m pip install riverwm-utils`. I added it to the [recommended software](https://github.com/riverwm/river/wiki/Recommended-Software) on the wiki. I'll close this now.

PR #509, opened 2022-01-11
command/map: Warn users on keybinding overwritten

### LINKED ISSUE #482: Alert user when overwriting keybindings
I really like that `riverctl map` overwrites existing keybindings with no fuss, but have been got a couple of times by accidentally overwriting my own keybindings in the init file. 

I think logging to the terminal when `riverctl map` overwrites an existing keybinding would be helpful and make it easier to recognize when this happens.

novakne:
Just display a message warning the users, don't throw errors and still map the new keybinding.

closes #482

### DIFF
--- a/river/command/map.zig
+++ b/river/command/map.zig
@@ -46,9 +46,13 @@ pub fn map(
 
     if (optionals.release and optionals.repeat) return Error.ConflictingOptions;
 
-    const mode_id = try modeNameToId(allocator, args[1 + offset], out);
-    const modifiers = try parseModifiers(allocator, args[2 + offset], out);
-    const keysym = try parseKeysym(allocator, args[3 + offset], out);
+    const mode_raw = args[1 + offset];
+    const modifiers_raw = args[2 + offset];
+    const keysym_raw = args[3 + offset];
+
+    const mode_id = try modeNameToId(allocator, mode_raw, out);
+    const modifiers = try parseModifiers(allocator, modifiers_raw, out);
+    const keysym = try parseKeysym(allocator, keysym_raw, out);
 
     const mode_mappings = &server.config.modes.items[mode_id].mappings;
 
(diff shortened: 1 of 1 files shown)


ifreund (commented):
Seems like a nice addition, could you please run `zig fmt`?

novakne:
> Seems like a nice addition, could you please run `zig fmt`?

Whoops my bad, removed a comma to see if it was under 100 columns and forgot to add it back, done.

river/command/map.zig
```diff
@@ -58,6 +58,12 @@ pub fn map(
     if (mappingExists(mode_mappings, modifiers, keysym, optionals.release)) |current| {
         mode_mappings.items[current].deinit();
         mode_mappings.items[current] = new;
+        // Warn user if they overwrote an existing keybinding using riverctl.
+        out.* = try std.fmt.allocPrint(
+            allocator,
+            "overwrote an existing keybinding: {s} {s}",
+            .{ args[2], args[3] },
```
ifreund, line 65: Does hardcoding the indexes 2 and 3 here work properly if e.g. -release is passed?

novakne: Hmm yeah didn't though of that, it does not, would you prefer that I make it works or just do not print the keys?

novakne: Yeah nvm the offset is already here, I thought I would need to add it, so wasn't sure if that was worth it, I'll do this

river/command/map.zig
```diff
@@ -46,9 +46,13 @@ pub fn map(
 
     if (optionals.release and optionals.repeat) return Error.ConflictingOptions;
 
-    const mode_id = try modeNameToId(allocator, args[1 + offset], out);
-    const modifiers = try parseModifiers(allocator, args[2 + offset], out);
-    const keysym = try parseKeysym(allocator, args[3 + offset], out);
+    const mode_name = args[1 + offset];
+    const modifiers_name = args[2 + offset];
+    const keysym_name = args[3 + offset];
```
ifreund, line 51: We use `raw` elsewhere in the codebase to describe unprocessed user input like this, I think it would be nice to do the same here.

river/command/map.zig
```diff
@@ -58,6 +62,13 @@ pub fn map(
     if (mappingExists(mode_mappings, modifiers, keysym, optionals.release)) |current| {
         mode_mappings.items[current].deinit();
         mode_mappings.items[current] = new;
+        // Warn user if they overwrote an existing keybinding using riverctl.
+        const opts = if (optionals.release) "-release " else "";
+        out.* = try std.fmt.allocPrint(
+            allocator,
+            "overwrote an existing keybinding: {s}{s} {s}",
+            .{ opts, modifiers_name, keysym_name },
```
ifreund, line 70: let's include mode as well for completeness.

river/command/map.zig
ifreund: It's not multiple keys though, it's a single keysym.

ifreund (approved):
LGTM, Thanks!

PR #510, opened 2022-01-13
rivertile: Use saturating addtion

novakne:
As far as I could tell this is the only occurrence where we can replace it. Grepping around, the only other case of `maxInt(T)` are use to set a value. Grepping for `math` also doesn't return any `math.mul/add/...`

### DIFF
--- a/rivertile/main.zig
+++ b/rivertile/main.zig
@@ -170,11 +170,7 @@ const Output = struct {
                             return;
                         };
                         switch (raw_arg[0]) {
-                            '+' => output.main_count = math.add(
-                                u32,
-                                output.main_count,
-                                @intCast(u32, arg),
-                            ) catch math.maxInt(u32),
+                            '+' => output.main_count +|= @intCast(u32, arg),
                             '-' => {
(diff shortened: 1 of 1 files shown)


rivertile/main.zig
```diff
@@ -170,11 +170,7 @@ const Output = struct {
                             return;
                         };
                         switch (raw_arg[0]) {
-                            '+' => output.main_count = math.add(
-                                u32,
-                                output.main_count,
-                                @intCast(u32, arg),
-                            ) catch math.maxInt(u32),
+                            '+' => output.main_count = output.main_count +| @intCast(u32, arg),
```
ifreund, line 173:
```suggestion
                            '+' => output.main_count +|= @intCast(u32, arg),
```

ifreund (changes requested):
Nice, we can make it even better though :)

rivertile/main.zig
novakne: Ah yeah, obviously, done!

novakne:
Fixed a typo in the commit message

ifreund (approved):
LGTM, thanks!

PR #520, opened 2022-01-29
Update cursor image when xcursor theme is changed and no client sets it

Leon-Plickat:


ifreund (approved):
LGTM, thanks for the fix!

PR #525, opened 2022-02-01
Don't stop applying input configs after first match.

Leon-Plickat:
A conversation on IRC showed me that device names aren't so unique after all and that multiple devices with the exact same name can be encountered in the real world.

Currently, river will apply an output config to the first input device with a matching device name. Input devices sharing the same name will not get the configuration because the loop `break`'s, which this patch changes.

Maybe we can instead make the names more unique. This however needs to be implemented in a way so that device names still stay consistent between sessions, so for example we can not just prefix them with a counter. Perhaps we can prefix them with the device type (like `touchpad:000:000:company-touchpad-something`), but that would still break if there are multiple devices with the same name of the same type. Also we'd loose input device name compatibility with Sway, which I think would be unfortunate.

Input and thoughts on this are welcome.

### DIFF
--- a/river/InputManager.zig
+++ b/river/InputManager.zig
@@ -46,11 +46,15 @@ pub const InputDevice = struct {
     identifier: []const u8,
 
     pub fn init(self: *InputDevice, device: *wlr.InputDevice) !void {
-        // The identifier is formatted exactly as in Sway
         const identifier = try std.fmt.allocPrint(
             util.gpa,
-            "{}:{}:{s}",
-            .{ device.vendor, device.product, mem.trim(u8, mem.span(device.name), &ascii.spaces) },
+            "{s}-{}-{}-{s}",
+            .{
+                @tagName(device.type),
+                device.vendor,
+                device.product,
+                mem.trim(u8, mem.span(device.name), &ascii.spaces),
+            },
         );
         for (identifier) |*char| {
             if (char.* == ' ' or !std.ascii.isPrint(char.*)) {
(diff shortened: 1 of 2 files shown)


adamcstephens:
> Perhaps we can prefix them with the device type (like `touchpad:000:000:company-touchpad-something`)

Sway already does something similar to this. For example here are two of my devices from my former sway config:

1739:52619:SYNA8004:00_06CB:CD8B_Touchpad
1452:613:Apple_Inc._Magic_Trackpad_2

Leon-Plickat:
I am pretty sure the `_Touchpad` and the `_Trackpad` are part of what the device advertises as its name, not anything Sway does. I checked the way Sway generates names when I implemented input configuration and it did not do anything like that. And I am unaware of Sways input device naming scheme changing since then.

adamcstephens:
> Currently you should see the exact same input device names in river and Sway.

You're correct, I'm wrong. Sway does identify touchpads as a different type then general pointers, but that's a different topic.

ifreund:
I'm fine with river's input device naming scheme diverging from sway's if it makes sense to do so, which it seems to me it does.

+1 from me on prefixing input device names with the device type.

Leon-Plickat:
Changed the name format to `[type]-[vendor id]-[product id]-[name]`. I replaced `:` with `-` since I think it's a bit more readable.

river/InputManager.zig
```diff
@@ -49,8 +49,13 @@ pub const InputDevice = struct {
         // The identifier is formatted exactly as in Sway
```
ifreund, line 49: We should probably remove this comment in the commit changing the formatting.

ifreund (approved):
LGTM, thanks!

PR #532, opened 2022-02-07
Mode: Use ArrayListUnmanaged to save memory

novakne:
- Use ArrayListUnmanaged for `Mode.mappings` and `Mode.pointer_mappings` and also for `Config.Mode`.

- Remove `Mode.init()` as the array list can be directly initialized in `Config.init()`, could be added back if any objection. But imo unless I missed something this doesn't seems to be really useful:
``` zig 
pub fn init() Self {
    return .{
        .mappings = .{},
        .pointer_mappings = .{},
    };
}
```

### DIFF
--- a/river/command/declare_mode.zig
+++ b/river/command/declare_mode.zig
@@ -39,9 +39,9 @@ pub fn declareMode(
 
     if (config.mode_to_id.get(new_mode_name) != null) return;
 
-    try config.modes.ensureUnusedCapacity(1);
+    try config.modes.ensureUnusedCapacity(util.gpa, 1);
     const owned_name = try util.gpa.dupe(u8, new_mode_name);
     errdefer util.gpa.free(owned_name);
     try config.mode_to_id.putNoClobber(owned_name, config.modes.items.len);
-    config.modes.appendAssumeCapacity(Mode.init());
+    config.modes.appendAssumeCapacity(.{});
 }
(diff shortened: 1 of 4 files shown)


river/command/declare_mode.zig
```diff
@@ -39,9 +39,9 @@ pub fn declareMode(
 
     if (config.mode_to_id.get(new_mode_name) != null) return;
 
-    try config.modes.ensureUnusedCapacity(1);
+    try config.modes.ensureUnusedCapacity(util.gpa, 1);
```
novakne, line 42: Here `util.gpa`, as in the lines below, is used when in `command/map.zig` (see below) the `allocator` args is used, I feel like this is kinda bad consistency but not sure which one to chose.

river/command/map.zig
```diff
@@ -73,7 +73,7 @@ pub fn map(
         // possible crash if the Mapping ArrayList is reallocated, stop any
         // currently repeating mappings.
         seat.clearRepeatingMapping();
-        try mode_mappings.append(new);
+        try mode_mappings.append(allocator, new);
```
novakne, line 76: Here and the next changed line are what I'm talking about above, with `allocator` vs `util.gpa`

ifreund: The allocator arg exists for historical reasons, today we always just pass `util.gpa`. I'd prefer if we just use `util.gpa` directly here.

It would be nice to remove the allocator arg from the command interface entirely, but that belongs in a different patch if you feel like it.

novakne: I Changed here to `util.gpa` and I'll remove the allocator args in an other PR

ifreund (approved):
LGTM, Thanks!

PR #533, opened 2022-02-07
code: Cleanup use of std library for consistancy

novakne:
So I come accross some case where there was in the same file for example `std.mem.foo()` but also `const mem = std.mem;` at the beginning of the file, so I tried to cleanup a bit everything, either adding `const foo = std.foo` if the lib was called a lot of time or just fix when both `std.foo` and `foo()` were used.

Sorry for the not so great commit message, I didn't know how to describe this tbh.


ifreund (approved):
LGTM, Thanks!

PR #534, opened 2022-02-08
WIP: Cursor (auto)hiding

tiosgz:
This is still in the stage when i’m figuring out how things could work, but i’ve stumbled on one thing i’m not sure i can solve myself. Otherwise do not waste time reviewing yet, unless you want to stop me from possibly going in the wrong direction.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -86,6 +86,17 @@ const Image = enum {
     @"se-resize",
 };
 
+/// Last image the cursor had before it was hidden
+const SavedImage = union(enum) {
+    builtin: Image,
+    client: struct {
+        client: *wlr.Seat.Client,
+        surface: ?*wlr.Surface,
+        hotspot_x: i32,
+        hotspot_y: i32,
+    },
+};
+
 const default_size = 24;
 
 const log = std.log.scoped(.cursor);
@@ -99,10 +110,12 @@ pointer_gestures: *wlr.PointerGesturesV1,
 xcursor_manager: *wlr.XcursorManager,
 
 image: Image = .unknown,
+
 auto_hide_timer: *wl.EventSource,
 /// Only ever useful if image == .none to determine whether the cursor has been
 /// hidden automatically, or explicitly
 auto_hidden: bool = false,
+saved_image: SavedImage = .{ .builtin = .left_ptr },
 
 constraint: ?*wlr.PointerConstraintV1 = null,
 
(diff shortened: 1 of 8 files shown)


river/Cursor.zig
```diff
@@ -744,25 +856,29 @@ pub fn enterMode(self: *Self, mode: enum { move, resize }, view: *View) void {
...
         },
         .move => {},
         .resize => |resize| resize.view.setResizing(false),
     }
 
     self.mode = .passthrough;
-    self.passthrough(event.time_msec);
+    // this is only a temporary solution. a fake event has to be created if
+    // there is none (TODO)
+    if (event != null) self.passthrough(event.?.time_msec);
```
tiosgz, line 877: This seems to be a problem when showing the cursor manually (via riverctl, see `command/config.zig`). Do you have an idea how to solve this?

Now thinking about it, *maybe* the issue could be that i don’t refocus the view under cursor though...

tiosgz: Sorry for rubber-ducking, the main issue i had seems to be solved.

But still i’m not sure what to do if there’s no event to pass to `leaveMode`. Make it up?

river/Cursor.zig
```diff
@@ -162,9 +184,16 @@ pub fn init(self: *Self, seat: *Seat) !void {
...
+    // Ensure that it is safe to call remove()
+    self.saved_image_surface_destroy.link = .{
+        .prev = &self.saved_image_surface_destroy.link,
+        .next = &self.saved_image_surface_destroy.link,
+    };
 }
 
 pub fn deinit(self: *Self) void {
+    self.auto_hide_timer.remove();
+    // why are no listeners removed?
```
tiosgz, line 196: Is there some logic behind this, or are no listeners removed because it hasn’t triggered any segfault yet? Should i remove `saved_image_surface_destroy`?

river/command.zig
```diff
@@ -70,6 +70,8 @@ const command_impls = std.ComptimeStringMap(
         .{ "send-layout-cmd",           @import("command/layout.zig").sendLayoutCmd },
         .{ "send-to-output",            @import("command/output.zig").sendToOutput },
         .{ "send-to-previous-tags",     @import("command/tags.zig").sendToPreviousTags },
+        // TODO: or hide-cursor never|now|delay msec?
+        .{ "set-cursor-state",          @import("command/config.zig").setCursorState },
```
tiosgz, line 74: Which variant is better? `hide-cursor never|now|delay`, or `set-cursor-state shown|hidden|auto-hide`?

river/command/config.zig
```diff
@@ -93,6 +93,44 @@ pub fn borderColorUrgent(
...
+    if (new_state != .@"auto-hide" and args.len > 2) return Error.TooManyArguments;
+
+    switch (new_state) {
+        .shown => {
+            server.config.cursor_auto_hide_delay = 0;
+            seat.cursor.resumeFromAutoHide();
+        },
+        .hidden => {
+            server.config.cursor_auto_hide_delay = 0;
+            if (seat.cursor.mode != .passthrough) return;
```
tiosgz, line 115: Should this mode requirement be documented in the manpage?

tiosgz (commented):
From my side, only these few questions remain to be answered (iirc) before the PR is ready.

tiosgz:
Apparently there’s #552 which seems to provide more functionality. Closing in favor of that, feel free to reopen if you want this PR instead.

PR #536, opened 2022-02-08
Libintput switch event handling and mapping

pmkap:
Use-case examples and suggested riverctl interface
`map-switch normal lid close spawn "wlr-randr --output eDP-1 --off"`
`map-switch normal lid open spawn "wlr-randr --output eDP-1 --on"`

Necessary wlroots bindings: https://github.com/swaywm/zig-wlroots/pull/39 MERGED

I used this to test the events: https://gist.github.com/pmkap/c9c31d5b0d7698507032df34393f00da
Working:
- [x] lid close
- [x] lid open
- [x] tablet on
- [x] tablet off

I choose not to implement the `lid:toggle` (resp. `tablet:toggle`) present in sway.


pmkap:
I'm wondering why wlroots adds a `TOGGLE` state here https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_switch.h#L36 

Looking at libinput there are only ON and OFF states: https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/src/libinput.h#L670

Wlroots never uses this third `TOGGLE` state from what I can see: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/backend/libinput/switch.c#L44

So river can never receive a toggle event the way I see it.

If I understand correctly, Sway keeps track of the current state to realize a toggle binding.
Sway example binding for this: `bindswitch lid:toggle exec echo "Lid moved"`

I'm thinking maybe we should remove the possibility to map to toggle. I cannot think of a use case where you would want to map the same action to 'lid close' AND 'lid open' ('tablet on' and 'tablet off' resp). And if, it could be done with two mappings.

pmkap:
Something else to think about: When river is started with e.g. laptop lid closed, there are no mappings present yet and the event is discarded. I'm not sure if there is a clean way to address this edge case.

Leon-Plickat:
> Something else to think about: When river is started with e.g. laptop lid closed, there are no mappings present yet and the event is discarded. I'm not sure if there is a clean way to address this edge case.

Realistically? With a configuration mechanism that is a bit more sophisticated than executing shell commands.

In other words, let's just accept that the edge case exists for now. Key-/Pointer- and Switch-maps may get moved out of river into the layout generators at some point and then it's no longer our problem.

Leon-Plickat:
>  I'm wondering why wlroots adds a TOGGLE state here https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_switch.h#L36

Maybe ask in `#sway-devel`

Leon-Plickat:
> I think, maybe we should not implement this for river. I cannot think of a use case where you would want to map the same action to 'lid close' AND 'lid open' ('tablet on' and 'tablet off' resp). And if, it could be done with two mappings.

+1

river/Seat.zig
```diff
@@ -412,6 +435,7 @@ pub fn addDevice(self: *Self, device: *wlr.InputDevice) void {
     self.wlr_seat.setCapabilities(.{
         .pointer = true,
         .keyboard = self.keyboards.len > 0,
+        // TODO: switch capability
```
ifreund, line 438: there is no switch capability or protocol that sends switch events to clients afaik.

river/Seat.zig
```diff
@@ -350,29 +359,42 @@ pub fn handleMapping(
...
+) void {
+    const modes = &server.config.modes;
+    for (modes.items[self.mode_id].switch_mappings.items) |mapping| {
+        if (std.meta.eql(mapping.switch_type, switch_type) and std.meta.eql(mapping.switch_state, switch_state)) {
+            self.runCommand(mapping.command_args);
+        }
+    }
+}
+
+fn runCommand(self: *Self, command_args: []const [:0]const u8) void {
```
ifreund, line 383: the diff would be cleaner if we kept this variable as just `args`.
```suggestion
fn runCommand(self: *Self, args: []const [:0]const u8) void {
```

river/Seat.zig
```diff
@@ -55,6 +56,9 @@ cursor: Cursor = undefined,
 /// Mulitple keyboards are handled separately
 keyboards: std.TailQueue(Keyboard) = .{},
 
+/// There can be two switches: lid switch and tablet mode switch
+switches: std.TailQueue(Switch) = .{},
```
ifreund, line 60: Does this comment mean that this list of switches can have at most two items? If so it might better to just have `lid_switch` and `tablet_switch` fields. If not I think it should be reworded to avoid confusion.

river/command/map.zig
```diff
@@ -148,6 +184,21 @@ fn mappingExists(
...
 }
 
+/// Returns the index of the SwitchMapping with matching switch_type and switch_state, if any.
+fn switchMappingExists(
+    switch_mappings: *std.ArrayListUnmanaged(SwitchMapping),
+    switch_type: Switch.Type,
+    switch_state: Switch.State,
+) ?usize {
+    for (switch_mappings.items) |mapping, i| {
+        if (std.meta.eql(mapping.switch_type, switch_type) and std.meta.eql(mapping.switch_state, switch_state)) {
```
ifreund, line 194: I don't think you need `std.meta.eql()` for `switch_type`, it's an enum so `==` should work fine.

river/command/map.zig
```diff
@@ -210,6 +261,58 @@ fn parseModifiers(modifiers_str: []const u8, out: *?[]const u8) !wlr.Keyboard.Mo
...
+    } else if (mem.eql(u8, switch_type_str, "tablet")) {
+        return .tablet;
+    }
+    out.* = try std.fmt.allocPrint(
+        util.gpa,
+        "invalid switch '{s}', must be 'lid' or 'tablet'",
+        .{switch_type_str},
+    );
+    return Error.Other;
+}
```
ifreund, line 279: This function could be replaced with `std.meta.stringToEnum(Switch.Type, str) orelse { ... }`

river/command/map.zig
```diff
@@ -210,6 +261,58 @@ fn parseModifiers(modifiers_str: []const u8, out: *?[]const u8) !wlr.Keyboard.Mo
...
+                return Switch.State{ .lid = .open };
+            } else if (mem.eql(u8, switch_state_str, "close")) {
+                return Switch.State{ .lid = .close };
+            }
+            out.* = try std.fmt.allocPrint(
+                util.gpa,
+                "invalid lid state '{s}', must be 'close' or 'open'",
+                .{switch_state_str},
+            );
+            return Error.Other;
```
ifreund, line 298: could use `std.meta.stringToEnum()` here and below as well.

ifreund (commented):
This is looking pretty good! Here are few comments to clean up some bits of the code.

river/Seat.zig
pmkap: There are probably no devices that have more than one of each, but one could have additional switches in terms of uinput devices. Would we need an extra Switch in this case?

river/command/map.zig
pmkap: For now I just changed the contents of the function and not the function itself to avoid duplication in `mapSwitch` and `unmapSwitch`. However it's just a few lines so maybe you prefer them inline?

ifreund (approved):
LGTM, sorry for the long wait :/

Just letting CI run on the newly squashed commit then I'll merge.

pmkap:
> LGTM, sorry for the long wait :/

No worries at all!

Issue #546, opened 2022-02-19
Rivertile crashes with integer overflow if main-ratio is changed


KSPAtlas:
I added `-main-ratio 55` to `rivertile`s args, however, when trying to open a terminal, the first one opens fine, but the second causes `rivertile` to panic (integer overflow) and `SIGABRT`. Running it even with `-main-ratio 60` causes it to crash, but no args works fine. I have checked and no other args are causing this behavior.

OS: Void Linux aarch64-musl
Host: Raspberry Pi 4b
River version: 0.1.0

Leon-Plickat:
Would be interesting to see the crash log as well as the dimensions of your screen.

Leon-Plickat:
Can you check if #547 fixes the issue for you?

KSPAtlas:
> Would be interesting to see the crash log as well as the dimensions of your screen.

1920x1080

Leon-Plickat:
Ah, I was under the impression the screen would be small, like a lot f Raspberry Pi screens are. In thic case the crashlog will be even more important, so please show that so we can find out what's causing the bug.

You can kill the currently running rivertile and start a new instance in a terminal, then reproduce the crash, to get the backtrace printed by zig. River should handle the crash and re-launching of rivertile just fine.

KSPAtlas:
> Ah, I was under the impression the screen would be small, like a lot of Raspberry Pi screens are. In this case the crashlog will be even more important, so please show that so we can find out what's causing the bug.
> 
> You can kill the currently running rivertile and start a new instance in a terminal, then reproduce the crash, to get the backtrace printed by zig. River should handle the crash and re-launching of rivertile just fine.

I have done that before, and there was no back trace:
```
thread 4024 panic: integer overflow
Aborted
```

ifreund:
Should be fixed by e67a9423a81359bd9d3bc33215b24222a0c85775

PR #547, opened 2022-02-19
Make rivertile not crash on small screens and/or when very large values for padding, etc., are used

Leon-Plickat:
In response to #546 

This will probably require more than one patch.


rivertile/main.zig
```diff
@@ -88,6 +88,10 @@ var default_main_ratio: f64 = 0.6;
 /// We don't free resources on exit, only when output globals are removed.
 const gpa = std.heap.c_allocator;
 
+/// Arbitrary minimal view size. This allows us to assert a minimum width and
+/// height, making the layout math nicer.
+const min_view_size = 10;
```
ifreund, line 93: It would be a lot simpler to just do `ev.usable_width -| 2 *outer_padding` IMO. I kinda hate the constants like this that we already have and would like to get rid of them.

Leon-Plickat: I'm not a fan of the constants either, however `ev.usable_width -| 2 * outer_padding` would still allow a usable width/height of `0`, which is nonsensical. In theory we should be able to rely on clients to have a sane minimum view size, but in practice this is not the case, so we probably need an arbitrary minimum view size  constant *somewhere*. And it does remove an edge case. However I do think river is the better place to enforce this than rivertile, so I'll update this patch accordingly.

Leon-Plickat:
I tested it and rivertile will not crash when the usable width/height is 0.

ifreund (approved):
LGTM thanks!

I added a commit that takes this a step further to get rid of nearly all the `@intCast()`s which seems to clean up the code nicely. I wonder if we should start using  `u31`s for widths/heights inside river's codebase as well.

To truly be confident that we handle all inputs correctly we'd need to fuzz test this code, but this is a good start.

Issue #453, opened 2021-10-20
Feature request: Ability to hide the cursor (akin to unclutter)


sagehane:
I couldn't find this feature on river, while sway seems to have it under `hide_cursor` here: https://github.com/swaywm/sway/blob/master/sway/sway-input.5.scd#seat-configuration.

I also couldn't find any wm-agnostic tool for this, so I figured it needs to be supported on river's side?

Leon-Plickat:
Yes, this is something a compositor needs to implement itself, since a Wayland client can only set the cursor image when the pointer is over on of its surfaces.

Should not be too hard to implement, basically just hooking into the idle system.

tiosgz:
> Tested on `0.2.0-dev-a8491eb` from archlinux aur.

Can you try git master? I think e0246c04104fad95573f07c3bfc4547c7ce1d116 could fix it.

ifreund:
Indeed, that commit fixed the issue described. I now see that I forgot to close this issue.

PR #552, opened 2022-02-27
Add command to hide cursor while typing or after a timeout

Duncaen:
From the riverctl.1:

```
*hide-cursor* *timeout* _timeout_
	Hide the cursor if it wasn't moved in the last _timeout_ milliseconds, until
	it is moved again.
	The default value is 0, which disables automatically hiding the cursor.

*hide-cursor* *jitter* _hide_delta_ _unhide_delta_
	- _hide_delta_: How many logical pixels the cursor has to move to reset the
	  hide timeout.
	- _unhide_delta_: How many logical pixels the cursor has to move to show the
	  cursor again.
	The default value is for both is 0, which disables the jitter.

*hide-cursor* *jitter* _delta_
	Sets both *hide-cursor* *jitter* _hide_delta_ and _unhide_delta_ to _delta_.

*hide-cursor* *when-typing* *enabled*|*disabled*
	Hide the cursor when typing starts, the cursor has to move to be shown again.
```

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -288,6 +288,15 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
 	If the view to be focused is on an output that does not have focus,
 	focus is switched to that output.
 
+*hide-cursor* *timeout* _timeout_
+	Hide the cursor if it wasn't moved in the last _timeout_ milliseconds
+	until it is moved again. The default value is 0, which disables
+	automatically hiding the cursor. Show the cursor again on any movement.
+
+*hide-cursor* *when-typing* *enabled*|*disabled*
+	Hide the cursor when pressing any non-modifier key. Show the cursor
+	again on any movement.
+
 *set-cursor-warp* *disabled*|*on-output-change*
 	Set the cursor warp mode. There are two available modes:
 
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -36,6 +36,11 @@ pub const WarpCursorMode = enum {
     @"on-output-change",
 };
 
+pub const HideCursorWhenTypingMode = enum {
+    disabled,
+    enabled,
+};
+
 /// Color of background in RGBA (alpha should only affect nested sessions)
 background_color: [4]f32 = [_]f32{ 0.0, 0.16862745, 0.21176471, 1.0 }, // Solarized base03
 
@@ -84,6 +89,12 @@ repeat_rate: u31 = 25,
 
 /// Keyboard repeat delay in milliseconds
 repeat_delay: u31 = 600,
+
+/// Cursor hide timeout in milliseconds
+cursor_hide_timeout: u31 = 0,
+
+/// Hide the cursor while typing
+cursor_hide_when_typing: HideCursorWhenTypingMode = .disabled,
 
 pub fn init() !Self {
     var self = Self{
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -102,6 +102,10 @@ constraint: ?*wlr.PointerConstraintV1 = null,
 /// Number of distinct buttons currently pressed
 pressed_count: u32 = 0,
 
+hide_cursor_timer: *wl.EventSource,
+
+hidden: bool = false,
+
 axis: wl.Listener(*wlr.Pointer.event.Axis) = wl.Listener(*wlr.Pointer.event.Axis).init(handleAxis),
 frame: wl.Listener(*wlr.Cursor) = wl.Listener(*wlr.Cursor).init(handleFrame),
 button: wl.Listener(*wlr.Pointer.event.Button) =
@@ -136,12 +140,16 @@ pub fn init(self: *Self, seat: *Seat) !void {
     const xcursor_manager = try wlr.XcursorManager.create(null, default_size);
     errdefer xcursor_manager.destroy();
 
+    const event_loop = server.wl_server.getEventLoop();
     self.* = .{
         .seat = seat,
         .wlr_cursor = wlr_cursor,
         .pointer_gestures = try wlr.PointerGesturesV1.create(server.wl_server),
         .xcursor_manager = xcursor_manager,
+        .hide_cursor_timer = try event_loop.addTimer(*Self, handleHideCursorTimeout, self),
     };
+    errdefer self.hide_cursor_timer.remove();
+    try self.hide_cursor_timer.timerUpdate(server.config.cursor_hide_timeout);
     try self.setTheme(null, null);
 
     // wlr_cursor *only* displays an image on screen. It does not move around
@@ -165,6 +173,7 @@ pub fn init(self: *Self, seat: *Seat) !void {
 }
 
 pub fn deinit(self: *Self) void {
+    self.hide_cursor_timer.remove();
     self.xcursor_manager.destroy();
     self.wlr_cursor.destroy();
 }
@@ -250,6 +259,7 @@ fn handleAxis(listener: *wl.Listener(*wlr.Pointer.event.Axis), event: *wlr.Point
     const self = @fieldParentPtr(Self, "axis", listener);
 
     self.seat.handleActivity();
+    self.unhide();
 
     // Notify the client with pointer focus of the axis event.
     self.seat.wlr_seat.pointerNotifyAxis(
@@ -265,6 +275,7 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
     const self = @fieldParentPtr(Self, "button", listener);
 
     self.seat.handleActivity();
+    self.unhide();
 
     if (event.state == .released) {
         assert(self.pressed_count > 0);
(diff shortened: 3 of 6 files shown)


river/command/cursor.zig
```diff
@@ -0,0 +1,58 @@
...
+            4 => {
+                server.config.cursor_hide_jitter = try std.fmt.parseInt(u32, args[2], 10);
+                server.config.cursor_unhide_jitter = try std.fmt.parseInt(u32, args[3], 10);
+            },
+            else => return Error.TooManyArguments,
+        }
+    } else if (std.mem.eql(u8, "timeout", args[1])) {
+        if (args.len < 3) return Error.NotEnoughArguments;
+        if (args.len > 3) return Error.TooManyArguments;
+        server.config.cursor_hide_timeout = try std.fmt.parseInt(u31, args[2], 10);
```
ifreund, line 49: If the timeout is currently 0 and `riverctl hide-cursor timeout 1000` is run but the cursor is not moved, the cursor is never hidden since we don't actually start the timer here.

Likewise, if the cursor is hidden with an absurdly high `unhide_jitter` it is currently impossible to unhide the cursor through setting a lower `unhide_jitter` and moving the cursor. The only way to get the cursor shown again current is to click.

I think both of these should be resolved by calling `Cursor.unhide()` for the cursor of every seat in `server.input_manager.seats` at the end of this function.

river/command/cursor.zig
```diff
@@ -0,0 +1,58 @@
...
+    } else if (std.mem.eql(u8, "when-typing", args[1])) {
+        if (args.len < 3) return Error.NotEnoughArguments;
+        if (args.len > 3) return Error.TooManyArguments;
+        server.config.cursor_hide_when_typing = std.meta.stringToEnum(Config.HideCursorWhenTypingMode, args[3]) orelse
```
ifreund, line 53:
```suggestion
        server.config.cursor_hide_when_typing = std.meta.stringToEnum(Config.HideCursorWhenTypingMode, args[2]) orelse
```

river/Cursor.zig
```diff
@@ -165,10 +176,17 @@ pub fn init(self: *Self, seat: *Seat) !void {
...
 
 pub fn deinit(self: *Self) void {
+    self.hide_cursor_timer.remove();
     self.xcursor_manager.destroy();
     self.wlr_cursor.destroy();
 }
 
+fn getHideTimeout(self: *Self) u31 {
+    if (self.pressed_count > 0)
+        return 0;
```
ifreund, line 186: Style nit: I prefer either
```zig
if (self.pressed_count > 0) return 0;
```
or
```zig
if (self.pressed_count > 0) {
    return 0;
}
```
I'm sure you've seen some existing old examples of this style around the codebase, but I've been consistently updating them as I touch that code. I should really get around to codifying this in CONTRIBUTING.md.

river/Cursor.zig
```diff
@@ -487,6 +507,35 @@ fn handleRequestSetCursor(
...
+    };
+}
+
+pub fn unhide(self: *Self) void {
+    self.jitter = 0;
+    self.hide_cursor_timer.timerUpdate(self.getHideTimeout()) catch {
+        log.err("failed to update cursor hide timeout", .{});
+    };
+    if (!self.hidden)
+        return;
```
ifreund, line 526:
```suggestion
    if (!self.hidden) return;
```

river/Keyboard.zig
```diff
@@ -82,6 +82,10 @@ fn handleKey(listener: *wl.Listener(*wlr.Keyboard.event.Key), event: *wlr.Keyboa
 
     self.seat.clearRepeatingMapping();
 
+    if (server.config.cursor_hide_when_typing == .enabled) {
+        self.seat.cursor.hide();
+    }
+
```
ifreund, line 88: This can cause the cursor to be hidden while cursor buttons are pressed. We should be able to fix this by moving the `self.pressed_count > 0` check from `getHideTimeout()` to `hide()`.

river/Cursor.zig
```diff
@@ -762,6 +811,18 @@ fn leaveMode(self: *Self, event: *wlr.Pointer.event.Button) void {
...
 fn processMotion(self: *Self, device: *wlr.InputDevice, time: u32, delta_x: f64, delta_y: f64, unaccel_dx: f64, unaccel_dy: f64) void {
+    const jitter_limit = if (self.hidden)
+        server.config.cursor_unhide_jitter
+    else
+        server.config.cursor_hide_jitter;
+    self.jitter += @floatToInt(u32, @fabs(delta_x) + @fabs(delta_y));
+    if (self.jitter > jitter_limit) {
+        self.unhide();
+    } else if (self.hidden) {
+        // If the cursor is hidden and the jitter limit is not reached, ignore the motion.
```
ifreund, line 822: With the other suggested changes, we can add some assertions here:
```suggestion
        // If the cursor is hidden and the jitter limit is not reached, ignore the motion.
        assert(self.pressed_count == 0);
        assert(self.mode == .passthrough);
```

ifreund (changes requested):
Thanks for the patch, this looks quite good already but there are a couple of minor tweaks/fixes I think we should make before merging.

I'm also a little curious what hardware setup you use that the jitter options make sense for? I wouldn't have thought to add them myself but I don't mind having them as they're quite simple.

river/command/cursor.zig
Duncaen: I solved this by iterating over seats and calling `cursor.unhide()` to reset the timeout, this seems like the simplest solution.

river/Keyboard.zig
Duncaen: Resolved it using the suggested solution. Hiding the cursor while typing is weird, i.e. hitting mod to move/resize a view will currently hide the cursor until the button is pressed which looks strange, maybe modifiers shouldn't hide the cursor.

Duncaen:
> I'm also a little curious what hardware setup you use that the jitter options make sense for? I wouldn't have thought to add them myself but I don't mind having them as they're quite simple.

Not sure if I have any hardware that requires it, its just something I've always had enabled with unclutter.

I was thinking it might be useful on notebooks to avoid unhiding it while typing and accidentally hitting the touchpad, but I just tested that and it would require jitter values that are too high to actually feel usable.

ifreund:
> I was thinking it might be useful on notebooks to avoid unhiding it while typing and accidentally hitting the touchpad, but I just tested that and it would require jitter values that are too high to actually feel usable.

Hmm, if there's no concrete use-case then perhaps we should leave it out for now to keep the configuration interface simpler? I personally don't really notice any difference in practice between jitter settings that are small enough to not be annoying and no jitter at all.

river/Keyboard.zig
ifreund: I agree, it would be better if modifiers didn't hide the cursor. Unfortunately `libxkbcommon` doesn't expose anything nice to check this in its public API afaict. It does however have this internal function:
https://github.com/xkbcommon/libxkbcommon/blob/a2507c0866543542e1bf310ca27695d7b33c83a1/src/keysym.c#L240-L248. Copying just the first part of that check (`(keysym >= XKB_KEY_Shift_L && keysym <= XKB_KEY_Hyper_R)`) should be sufficient for our use-case I believe.

river/Cursor.zig
```diff
@@ -487,6 +498,42 @@ fn handleRequestSetCursor(
...
+        if (self.seat.wlr_seat.pointer_state.focused_surface != result.surface) {
+            switch (result.parent) {
+                .view => {
+                    self.seat.wlr_seat.pointerNotifyEnter(result.surface, result.sx, result.sy);
+                },
+                .layer_surface => {},
+                .xwayland_unmanaged => assert(build_options.xwayland),
+            }
+        }
+    }
```
ifreund, line 528: This could be replaced with a call to `self.updateState()`, which would also be a bit more correct I believe.

river/Config.zig
```diff
@@ -85,6 +90,18 @@ repeat_rate: u31 = 25,
...
 repeat_delay: u31 = 600,
 
+/// Cursor hide timeout in milliseconds
+cursor_hide_timeout: u31 = 0,
+
+/// How many pixels the cursor has to be moved to reset the hide tiemout.
+cursor_hide_jitter: u32 = 0,
+
+/// How many pixels the cursor has to be moved to unhide it.
+cursor_unhide_jitter: u32 = 0,
```
ifreund, line 100: Looks like these should have been removed in 21e418f6fbd0b0ec5ef91a5f7d0a6a93dcad3060

doc/riverctl.1.scd
```diff
@@ -267,6 +267,24 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
...
+*hide-cursor* *jitter* _hide_delta_ _unhide_delta_
+	- _hide_delta_: How many logical pixels the cursor has to move to reset the
+	  hide timeout.
+	- _unhide_delta_: How many logical pixels the cursor has to move to show the
+	  cursor again.
+	The default value is for both is 0, which disables the jitter.
+
+*hide-cursor* *jitter* _delta_
+	Sets both *hide-cursor* *jitter* _hide_delta_ and _unhide_delta_ to _delta_.
+
```
ifreund, line 284:
```suggestion
```

sagehane:
I presume this PR resolves https://github.com/riverwm/river/issues/453? If so, thanks.

ifreund:
> I presume this PR resolves #453? If so, thanks.

Indeed it does :)

ifreund (approved):
Thank you again for this patch, sorry it took me so long to get around to merging it.

I've tweaked the code in Keyboard.zig slightly to hide the cursor only on pressing a non-modifier key and squashed the commits. I'll just let CI run to make sure it's green before merging :)

PR #553, opened 2022-03-02
view: stop enforcing custom minimum size

tiosgz:
I have encountered a crash (failing assert) if a view specified a fixed
size less than this minimum, and according to ifreund this behavior was
planned to be removed, anyway.


ifreund (approved):
LGTM, thanks!

While testing this I learned that
1. Is is infact possible to resize a view that is 1x1 pixel in size with patience and a steady hand.
2. The "hotspot" of my cursor is the tip of the black cursor itself, not the white border around the black.

This makes me want to make view borders valid to grab for move/resize as well, but that doesn't have to block this merge.

PR #558, opened 2022-03-05
Fix builds manifest

novakne:


novakne:
@ifreund so yeah the ci was build against river master everytime, as I messed up a link for river repo in the manifests, my bad.

I test pushing a commit that should not build with 2 new manifest (freebsd, arch) and one as before (alpine) and as expected, you can see in build.sr.ht that only the alpine one didn't failed, so it should be all good now.

The [`sources:`](https://builds.sr.ht/api/jobs/709211/manifest) is right and you see that it also checkout the right [commit](https://builds.sr.ht/~ifreund/job/709211#setup-437) to build river

.builds/alpine.yml
```diff
@@ -20,9 +20,9 @@ packages:
   - wget
   - xz
 sources:
+  - https://github.com/riverwm/river
   - https://gitlab.freedesktop.org/wayland/wayland.git
   - https://gitlab.freedesktop.org/wlroots/wlroots.git
-  - https://github.com/riverwm/river.git
```
ifreund, line 25: What exactly is the fix here? The ordering change? The removal of the `.git` suffix? both?

novakne: oh yeah sorry it's only the removing of the `.git` suffix, the ordering was the first thing I tried and didn't change it back, I can revert it if you want, I don't think the order is relevant, although I think I prefer it in this order

ifreund: Should we get rid of the `.git` suffix for the other sources as well? It would be nice to explain what exactly the fix is in the commit message.

novakne: Let me know if that new commit message clear things up, the documents for builds.sr.ht is not really clear for suffix and details like this, but as far I could see looking around for build manifest exemples, no suffix for the repo you want to test and suffix for deps seems like the usual way.

ifreund (approved):
LGTM, Thanks!

PR #569, opened 2022-03-13
xwayland: handle keyboard demand

Shinyzenith:


Shinyzenith:
This patch crashes on attempting to run Xwayland apps so I thought it's best to open the pr and mark it as a draft so more experienced developers can chime in with possible fixes.

PR #577, opened 2022-03-31
layout-independent mappings

pmkap:
After some discussions, we came to the conclusion, that layout-independent mappings are best achieved with the ability to pin a certain layout to a mapping.

E.g. something like `riverctl map -layout "French" normal Super Y spawn foot`.

When this mapping is checked, the specified layout will be used to translate the received key instead of the currently active layout.

The details are yet to be decided.

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -193,13 +193,19 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
 *enter-mode* _name_
 	Switch to given mode if it exists.
 
-*map* [_-release_|_-repeat_] _mode_ _modifiers_ _key_ _command_
+*map* [_-release_|_-repeat_|_-layout_ _index_] _mode_ _modifiers_ _key_ _command_
 	Run _command_ when _key_ is pressed while _modifiers_ are held down
 	and in the specified _mode_.
 
 	- _-release_: if passed activate on key release instead of key press
 	- _-repeat_: if passed activate repeatedly until key release; may not
 	  be used with -release
+	- _-layout_: if passed, a specific layout is pinned to the mapping.
+	  When the mapping is checked against a pressed key, this layout is
+	  used to translate the key independent of the active layout
+	- _index_: zero-based index of a layout set with the environment variable
+	  *XKB_DEFAULT_LAYOUT*; see *river*(1) for an example; if the index is
+	  out of range, the _-layout_ option will have no effect
 	- _mode_: name of the mode for which to create the mapping
 	- _modifiers_: one or more of the modifiers listed above, separated
 	  by a plus sign (+).
--- a/river/Keyboard.zig
+++ b/river/Keyboard.zig
@@ -85,53 +85,39 @@ fn handleKey(listener: *wl.Listener(*wlr.Keyboard.event.Key), event: *wlr.Keyboa
     // Translate libinput keycode -> xkbcommon
     const keycode = event.keycode + 8;
 
-    // TODO: These modifiers aren't properly handled, see sway's code
     const modifiers = wlr_keyboard.getModifiers();
     const released = event.state == .released;
 
-    var handled = false;
+    const xkb_state = wlr_keyboard.xkb_state orelse return;
+    const keysyms = xkb_state.keyGetSyms(keycode);
 
-    var non_modifier_pressed = false;
-
-    // First check translated keysyms as xkb reports them
-    for (wlr_keyboard.xkb_state.?.keyGetSyms(keycode)) |sym| {
-        if (!released and !isModifier(sym)) non_modifier_pressed = true;
-
-        // Handle builtin mapping only when keys are pressed
-        if (!released and handleBuiltinMapping(sym)) {
-            handled = true;
-            break;
-        } else if (self.seat.handleMapping(sym, modifiers, released)) {
-            handled = true;
+    // Hide cursor when typing
+    for (keysyms) |sym| {
+        if (server.config.cursor_hide_when_typing == .enabled and
+            !released and
+            !isModifier(sym))
+        {
+            self.seat.cursor.hide();
             break;
         }
     }
 
-    // If not yet handled, check keysyms ignoring modifiers (e.g. 1 instead of !)
-    // Important for mappings like Mod+Shift+1
-    if (!handled) {
-        const layout_index = wlr_keyboard.xkb_state.?.keyGetLayout(keycode);
-        for (wlr_keyboard.keymap.?.keyGetSymsByLevel(keycode, layout_index, 0)) |sym| {
-            // Handle builtin mapping only when keys are pressed
-            if (!released and handleBuiltinMapping(sym)) {
-                handled = true;
-                break;
-            } else if (self.seat.handleMapping(sym, modifiers, released)) {
-                handled = true;
-                break;
-            }
-        }
+    // Handle builtin mapping, only when keys are pressed
+    for (keysyms) |sym| {
+        if (!released and handleBuiltinMapping(sym)) return;
     }
 
-    if (!handled) {
-        // Otherwise, we pass it along to the client.
+    // Handle user-defined mapping
+    if (!self.seat.handleMapping(
+        keycode,
+        modifiers,
+        released,
+        xkb_state,
+    )) {
+        // If key was not handled, we pass it along to the client.
         const wlr_seat = self.seat.wlr_seat;
         wlr_seat.setKeyboard(self.input_device);
         wlr_seat.keyboardNotifyKey(event.time_msec, event.keycode, event.state);
-    }
-
-    if (non_modifier_pressed and server.config.cursor_hide_when_typing == .enabled) {
-        self.seat.cursor.hide();
     }
 }
 
--- a/river/Seat.zig
+++ b/river/Seat.zig
@@ -342,17 +342,20 @@ pub fn handleViewUnmap(self: *Self, view: *View) void {
     if (self.focused == .view and self.focused.view == view) self.focus(null);
 }
 
-/// Handle any user-defined mapping for the passed keysym and modifiers
+/// Handle any user-defined mapping for passed keycode, modifiers and keyboard state
 /// Returns true if the key was handled
 pub fn handleMapping(
     self: *Self,
-    keysym: xkb.Keysym,
+    keycode: xkb.Keycode,
     modifiers: wlr.Keyboard.ModifierMask,
     released: bool,
+    xkb_state: *xkb.State,
 ) bool {
     const modes = &server.config.modes;
+    // In case more than on mapping matches, all of them are activated
+    var handled = false;
     for (modes.items[self.mode_id].mappings.items) |*mapping| {
-        if (std.meta.eql(modifiers, mapping.modifiers) and keysym == mapping.keysym and released == mapping.release) {
+        if (mapping.match(keycode, modifiers, released, xkb_state)) {
             if (mapping.repeat) {
                 self.repeating_mapping = mapping;
                 self.mapping_repeat_timer.timerUpdate(server.config.repeat_delay) catch {
@@ -360,10 +363,10 @@ pub fn handleMapping(
                 };
             }
             self.runCommand(mapping.command_args);
-            return true;
+            handled = true;
         }
     }
-    return false;
+    return handled;
 }
 
 /// Handle any user-defined mapping for switches
--- a/river/command/map.zig
+++ b/river/command/map.zig
@@ -40,7 +40,7 @@ pub fn map(
     args: []const [:0]const u8,
     out: *?[]const u8,
 ) Error!void {
-    const optionals = parseOptionalArgs(args[1..]);
+    const optionals = try parseOptionalArgs(args[1..]);
     // offset caused by optional arguments
     const offset = optionals.i;
     if (args.len - offset < 5) return Error.NotEnoughArguments;
@@ -57,7 +57,14 @@ pub fn map(
 
     const mode_mappings = &server.config.modes.items[mode_id].mappings;
 
-    const new = try Mapping.init(keysym, modifiers, optionals.release, optionals.repeat, args[4 + offset ..]);
+    const new = try Mapping.init(
+        keysym,
+        modifiers,
+        optionals.release,
+        optionals.repeat,
+        optionals.layout_index,
+        args[4 + offset ..],
+    );
     errdefer new.deinit();
 
     if (mappingExists(mode_mappings, modifiers, keysym, optionals.release)) |current| {
@@ -316,32 +323,41 @@ const OptionalArgsContainer = struct {
     i: usize,
     release: bool,
     repeat: bool,
+    layout_index: ?u32,
 };
 
 /// Parses optional args (such as -release) and return the index of the first argument that is
 /// not an optional argument
 /// Returns an OptionalArgsContainer with the settings set according to the args
-/// Errors cant occur because it returns as soon as it gets an unknown argument
-fn parseOptionalArgs(args: []const []const u8) OptionalArgsContainer {
+fn parseOptionalArgs(args: []const []const u8) !OptionalArgsContainer {
     // Set to defaults
     var parsed = OptionalArgsContainer{
         // i is the number of arguments consumed
         .i = 0,
         .release = false,
         .repeat = false,
+        .layout_index = null,
     };
 
-    var i: usize = 0;
-    for (args) |arg| {
-        if (mem.eql(u8, arg, "-release")) {
+    var j: usize = 0;
+    while (j < args.len) : (j += 1) {
+        if (mem.eql(u8, args[j], "-release")) {
             parsed.release = true;
-            i += 1;
-        } else if (mem.eql(u8, arg, "-repeat")) {
+            parsed.i += 1;
+        } else if (mem.eql(u8, args[j], "-repeat")) {
             parsed.repeat = true;
-            i += 1;
+            parsed.i += 1;
+        } else if (mem.eql(u8, args[j], "-layout")) {
+            j += 1;
+            if (j == args.len) return Error.NotEnoughArguments;
+            // To keep things simple here, we do not check if the layout index
+            // is out of range. We rely on xkbcommon to handle this case:
+            // xkbcommon will simply use the active layout instead, leaving
+            // this option without effect
+            parsed.layout_index = try std.fmt.parseInt(u32, args[j], 10);
+            parsed.i += 2;
         } else {
             // Break if the arg is not an option
-            parsed.i = i;
             break;
         }
     }
@@ -354,7 +370,7 @@ fn parseOptionalArgs(args: []const []const u8) OptionalArgsContainer {
 /// Example:
 /// unmap normal Mod4+Shift Return
 pub fn unmap(seat: *Seat, args: []const [:0]const u8, out: *?[]const u8) Error!void {
-    const optionals = parseOptionalArgs(args[1..]);
+    const optionals = try parseOptionalArgs(args[1..]);
     // offset caused by optional arguments
     const offset = optionals.i;
     if (args.len - offset < 4) return Error.NotEnoughArguments;
--- a/completions/bash/riverctl
+++ b/completions/bash/riverctl
@@ -56,7 +56,7 @@ function __riverctl_completion ()
 			"focus-output"|"focus-view"|"send-to-output"|"swap") OPTS="next previous" ;;
 			"move"|"snap") OPTS="up down left right" ;;
 			"resize") OPTS="horizontal vertical" ;;
-			"map") OPTS="-release -repeat" ;;
+			"map") OPTS="-release -repeat -layout" ;;
 			"unmap") OPTS="-release" ;;
 			"attach-mode") OPTS="top bottom" ;;
 			"focus-follows-cursor") OPTS="disabled normal" ;;
--- a/completions/fish/riverctl.fish
+++ b/completions/fish/riverctl.fish
@@ -70,7 +70,7 @@ complete -c riverctl -x -n '__fish_seen_subcommand_from resize'               -a
 complete -c riverctl -x -n '__fish_seen_subcommand_from snap'                 -a 'up down left right'
 complete -c riverctl -x -n '__fish_seen_subcommand_from send-to-output'       -a 'next previous'
 complete -c riverctl -x -n '__fish_seen_subcommand_from swap'                 -a 'next previous'
-complete -c riverctl -x -n '__fish_seen_subcommand_from map'                  -a '-release -repeat'
+complete -c riverctl -x -n '__fish_seen_subcommand_from map'                  -a '-release -repeat -layout'
 complete -c riverctl -x -n '__fish_seen_subcommand_from unmap'                -a '-release'
 complete -c riverctl -x -n '__fish_seen_subcommand_from attach-mode'          -a 'top bottom'
 complete -c riverctl -x -n '__fish_seen_subcommand_from focus-follows-cursor' -a 'disabled normal'
--- a/completions/zsh/_riverctl
+++ b/completions/zsh/_riverctl
@@ -169,7 +169,7 @@ _riverctl()
                 snap) _alternative 'arguments:args:(up down left right)' ;;
                 send-to-output) _alternative 'arguments:args:(next previous)' ;;
                 swap) _alternative 'arguments:args:(next previous)' ;;
-                map) _alternative 'arguments:optional:(-release -repeat)' ;;
+                map) _alternative 'arguments:optional:(-release -repeat -layout)' ;;
                 unmap) _alternative 'arguments:optional:(-release)' ;;
                 attach-mode) _alternative 'arguments:args:(top bottom)' ;;
                 focus-follows-cursor) _alternative 'arguments:args:(disabled normal)' ;;
--- a/river/Mapping.zig
+++ b/river/Mapping.zig
@@ -26,6 +26,10 @@ keysym: xkb.Keysym,
 modifiers: wlr.Keyboard.ModifierMask,
 command_args: []const [:0]const u8,
 
+// This is set for mappings with layout-pinning
+// If set, the layout with this index is always used to translate the given keycode
+layout_index: ?u32,
+
 /// When set to true the mapping will be executed on key release rather than on press
 release: bool,
 
@@ -37,6 +41,7 @@ pub fn init(
     modifiers: wlr.Keyboard.ModifierMask,
     release: bool,
     repeat: bool,
+    layout_index: ?u32,
     command_args: []const []const u8,
 ) !Self {
     const owned_args = try util.gpa.alloc([:0]u8, command_args.len);
@@ -50,6 +55,7 @@ pub fn init(
         .modifiers = modifiers,
         .release = release,
         .repeat = repeat,
+        .layout_index = layout_index,
         .command_args = owned_args,
     };
 }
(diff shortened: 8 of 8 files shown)


MaxVerevkin:
Running `riverctl map -to-code normal Super 9 set-focused-tags 256` results in a crash.

Log: https://paste.rs/LPo
Config: https://paste.rs/Aa9

pmkap:
@MaxVerevkin Thanks! I'm looking into it.

pmkap:
@MaxVerevkin I cannot reproduce this. Could you provide a backtrace?

MaxVerevkin:
> Could you provide a backtrace?

How can I do this :) ?

pmkap:
@MaxVerevkin `coredumpctl debug` starts gdb with the last coredump, then you cat get the backtrace with `bt`.

MaxVerevkin:
@pmkap Like this?

image

pmkap:
@MaxVerevkin Yes, thank you. I wasn't able to find the problem yet. 

However, I'm thinking about restructuring the `handleMapping` and `handleKey` functions for some clearness. I will come back to your crash after this.

pmkap:
@MaxVerevkin I found the crash. Should be resolved now. Thank you!

MaxVerevkin:
@pmkap It doesn't crash, but it still doesn't work:

```
$ riverctl map -to-code normal Super 1 set-focused-tags 1
error: couldn't translate keysym to keycode
```

Oh, and I forgot to mention that I have this in my `.zprofile`:

```
export XKB_DEFAULT_LAYOUT="us(workman),ru"
```

pmkap:
@MaxVerevkin Is your `1` on the first level? Because only the first level is checked.

MaxVerevkin:
@pmkap No, I have to press Shift to get number one. Ideally I would expect -to-code to work like rivers --to-code: it just uses qwerty.

MaxVerevkin:
> Sway also only checks the first level for the keysym

Yes, but it uses qwerty (actually IDK if it's a feature or a bug).

MaxVerevkin:
Just sad that I can't do [like this](https://github.com/riverwm/river/blob/ae349b0ce45b44b79a0a93e89c2cd05d24bb63d4/example/init#L74-L89) with my layout :laughing:

MaxVerevkin:
Anyway, thanks @pmkap for this PR :)

pmkap:
@MaxVerevkin How about this?
```sh
count=0
for i in Exclam At Dollar Percent .....
do
    tags=$((1 << count))
    count=$((count + 1))
    riverctl map normal Super $i set-focused-tags $tags
    ...
done
```

We *could* of course also check the the second layer for a translation but IMO it's not necessary.

novakne:
I do it like [this](https://git.sr.ht/~novakane/dotfiles/tree/main/item/.config/river/init#L120-133) with bash and an azerty keyboard

MaxVerevkin:
Thanks for the suggestions. I'm not sure what I'm doing wrong, but
```
$ riverctl map normal Super dollar set-focused-tags 8
overwrote an existing keybinding: normal Super dollar
$ riverctl map -to-code normal Super dollar set-focused-tags 8
error: couldn't translate keysym to keycode
```

pmkap:
@MaxVerevkin Hm strange. The others do work? I think I need your layout to reproduce this. Would you share it?

MaxVerevkin:
Everything except for dollar and parenthesis. 

I have this in my `/usr/share/X11/xkb/symbols/us` (there must be a better way to override layouts, but idk): https://paste.rs/f9S

pmkap:
@MaxVerevkin Thanks for your help!
The problem is, that more than one keysym is found for the keycode. Sway has exactly the same problem (I copied that part over from sway)

I'll think on how to solve this.

MaxVerevkin:
@pmkap Tested the latest commit, works perfectly!

river/Config.zig
```diff
@@ -85,10 +86,27 @@ repeat_rate: u31 = 25,
...
+    const rules = xkb.RuleNames{
+        .rules = null,
+        .model = null,
+        .layout = null,
+        .variant = null,
+        .options = null,
+    };
+    const context = xkb.Context.new(.no_flags) orelse return error.XkbContextFailed;
+    defer context.unref();
+    const keymap = xkb.Keymap.newFromNames(context, &rules, .no_flags) orelse return error.XkbKeymapFailed;
```
ifreund, line 102: Passing null here is treated the same as passing a all-null `xkb.RuleNames` struct.
```suggestion
    const keymap = xkb.Keymap.newFromNames(context, null, .no_flags) orelse return error.XkbKeymapFailed;
```

ifreund (commented):
This is looking good, nice cleanup of the `Keyboard.zig` code. Unfortunately I've introduced some conflicts by merging the `hide-cursor` PR, could you please resolve those?

pmkap:
> This is looking good, nice cleanup of the `Keyboard.zig` code. Unfortunately I've introduced some conflicts by merging the `hide-cursor` PR, could you please resolve those?

Sure!

doc/riverctl.1.scd
```diff
@@ -193,10 +193,12 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
 *enter-mode* _name_
 	Switch to given mode if it exists.
 
-*map* [_-release_|_-repeat_] _mode_ _modifiers_ _key_ _command_
+*map* [_-to-code_|_-release_|_-repeat_] _mode_ _modifiers_ _key_ _command_
```
ifreund, line 196: What do you think about making this flag `-positional` instead? I think that more intuitively explains that this is mapping a position on the physical keyboard rather than the logical keysym received from xkbcommon.

river/command/map.zig
```diff
@@ -31,6 +31,21 @@ const SwitchMapping = @import("../SwitchMapping.zig");
 const Switch = @import("../Switch.zig");
 const Seat = @import("../Seat.zig");
 
+const KeycodeMatch = struct {
+    keysym: xkb.Keysym,
+    keycode: xkb.Keycode,
+};
+
+/// Iterate through keycodes in the keymap to find ones matching the specified keysym.
+/// If more than one keycode matches, use the smallest.
```
ifreund, line 40: It'd be nice to explain why the smallest is used in this comment, perhaps with an example.

river/Seat.zig
```diff
@@ -342,28 +342,53 @@ pub fn handleViewUnmap(self: *Self, view: *View) void {
...
+    if (mapping_match) |mapping| {
+        if (mapping.repeat) {
+            self.repeating_mapping = mapping;
+            self.mapping_repeat_timer.timerUpdate(server.config.repeat_delay) catch {
+                log.err("failed to update mapping repeat timer", .{});
+            };
+        }
+        self.runCommand(mapping.command_args);
+        return true;
+    } else return false;
```
ifreund, line 391: Since the other branch of the if has braces I'd rather this one have them too to keep things visually balanced.
```suggestion
    } else {
        return false;
    }
```

ifreund (commented):
Here's a round of code style suggestions and an alternative naming proposal. The logic of the code itself already looks great :)

river/Keyboard.zig
```diff
@@ -85,54 +80,56 @@ fn handleKey(listener: *wl.Listener(*wlr.Keyboard.event.Key), event: *wlr.Keyboa
...
+        modifiers_raw,
+        keysyms_translated,
+        modifiers_translated,
+        released,
+    )) return;
+
+    // If key was not handled, we pass it along to the client.
+    const wlr_seat = self.seat.wlr_seat;
+    wlr_seat.setKeyboard(self.input_device);
+    wlr_seat.keyboardNotifyKey(event.time_msec, event.keycode, event.state);
```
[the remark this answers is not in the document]
pmkap: > What do you think?

I agree.

river/command/map.zig
pmkap: Sure I'll add an example to the comment.

What do you think of making keycodes a linked list and add all found keycodes to the mapping? This would probably be the most "correct" way.

The cases where I found second keycodes were all quite high keycodes that do not appear on normal keyboards, e.g. `KEY_KPLEFTPAREN` or even more obscure ones. (This was with French AZERTY layout). But users could have custom layouts where they have a keysym on multiple keys and they want the mapping for all of them.

doc/riverctl.1.scd
pmkap: I mainly choose `-to-code` because it is already present in sway, `-positional` is also fine for me.

pmkap: @ifreund Another option could be `-independent` for layout-independent. I also like `-static` or `-fixed`.

river/command/map.zig
ifreund: Hmm, I'm not a huge fan of introducing more complexity for an edge case I'm not sure we'll ever see in practice.

I was just looking around the xkbcommon API a bit and found the `xkb_keymap_key_by_name()` function which translates a string name to a keycode. Its implementation looks similar to what you have here, but with something additional about aliases. Perhaps we should just use that function?
```c
XKB_EXPORT xkb_keycode_t
xkb_keymap_key_by_name(struct xkb_keymap *keymap, const char *name)
{
    struct xkb_key *key;
    xkb_atom_t atom;

    atom = xkb_atom_lookup(keymap->ctx, name);
    if (atom) {
        xkb_atom_t ratom = XkbResolveKeyAlias(keymap, atom);
        if (ratom)
            atom = ratom;
    }
    if (!atom)
        return XKB_KEYCODE_INVALID;

    xkb_keys_foreach(key, keymap) {
        if (key->name == atom)
            return key->keycode;
    }

    return XKB_KEYCODE_INVALID;
}
```

ifreund: It seems to me that the simplest way forward with this would be to put everything in libxkbcommon's hands and use `xkb_keymap_key_by_name()` for keycode based mappings and `xkb_keysym_from_name()` for keysym based mappings. I don't think we should ever parse or store a keysym for keycode based mappings as doing a keysym to keycode is potentially not one-to-one and seems overcomplex to me.

ifreund: Hmm, that might be a bad idea actually. It's not clear to me that keycodes consistently have good user-facing names, it seems to depend on the keymap.

pmkap: Hm then we could just do a direct keycode mapping like sway's `bindcode`?

The huge advantage of keysym-to-keycode mapping IMO is that the user doesn't need to look up keycodes but just add the flag to an existing mapping. I think this was the reason why sway decided to add `--to-code` although it was not truly necessary because they already had `bindcode`.

ifreund: Did a bit more poking around. It looks like since the only input backend we/wlroots currently support is libinput which uses evdev the keycodes we receive from wlroots are always linux input event codes as defined in `linux/input-event-codes.h`. We already use these codes for mouse buttons, so I wouldn't mind using them here as well.

Thanks for reminding me that sway has bincode as well, I haven't used sway in quite a while. It looks like the way they implement it isn't very user-friendly though as it takes an xkb keycode number directly. We could instead make our version accept linux input event code names as I mentioned above, parsing with `libevdev_event_code_from_name()` and adding 8 to convert from linux input event code to xkb keycode.

As for whether we should have an equivalent to sway's `bindsym --to-code`, `bindcode`, or both let's remember that converting keysym to keycode as is done by `bindsym --to-code` can be ambiguous. This makes me think that we must eventually implement a `bindcode` equivalent in order to support the maximal number of use-cases. Furthermore, if we implement a `bindcode` equivalent that is not painful to use, I don't think we need a redundant `bindsym --to-code` equivalent.

What are your thoughts there? Thanks for helping me figure this stuff out :)

pmkap: Yeah, I thought about the same when implementing this. I was also sceptical about the redundancy of having both methods. Eventually, I chose `-to-code` over a direct keycode mapping because of the simplicity for the user, despite the potential ambiguity of the translation.

I do agree that evdev code names are nicer than the plain numbers sway uses for `bindcode`. But still users would need `sudo evtest` or something similar to find the event code name in a lot of cases. 

True, to support every use case, we need an equivalent to sway's `bindcode`. But do these use cases ever emerge? I'm not sure...

My concern is that with exclusive direct keycode mappings (be it with event code names or xkb keycodes), we end up with lots of complaints from people who doesn't want to lookup the event code names for all mappings. Like here https://github.com/swaywm/sway/issues/2999 and here https://github.com/swaywm/sway/pull/3058 .

Me personally, I'm fine with either solution (or even both) and I don't mind implementing a direct keycode mapping if you decide so.

> What are your thoughts there? Thanks for helping me figure this stuff out :)

You're welcome, I'm happy to be a part of this :)

ifreund: Cool, I think I'd like to stick to binding event codes directly by name and see how that works out. The implementation is far simpler and there's less that can go wrong from a technical perspective. It seems that emersion's initial reaction matched mine as well.

The core complaint in those sway discussions seems to be the readability of the config file, which I think will be mostly alleviated by using `KEY_A` instead of `30` for example. We could probably drop the `KEY_` prefix as well, not 100% confident that won't cause issues though.

As for the `riverctl` interface, I'm thinking it'd be more user friendly to have a separate `map-keycode` command, which makes me think we should rename `map` to `map-keysym`.
```
       map-keysym [-release|-repeat] mode modifiers keysym command
       map-keycode [-release|-repeat] mode modifiers keycode command
```

If this proves to be too hard to use, I'd rather explore adding a way for riverctl to print out the keysyms/keycodes being pressed before adding (flawed) keysym to keycode translation.

Leon-Plickat:
> If this proves to be too hard to use, I'd rather explore adding a way
> for riverctl to print out the keysyms/keycodes being pressed before
> adding (flawed) keysym to keycode translation.

wev already displays both keysym and keycode I think.

As for riverctl, we should look into whether this is possible without
connecting to river. Involving the server means either having a
surface or exposing what is effectively a keylogger protocol.

river/command/map.zig
pmkap: I think this would be very usable. If we dropped the `KEY_` prefix and make it case-insensitive, it would result in the same name as the keysym for most mappings in the default init file. So users with other layouts could use the default init file without changing much.

But then we maybe should do the same with mouse buttons for consistency.

pmkap:
> wev already displays both keysym and keycode I think. 

AFAIK not, one would need to use `sudo libinput debug-events --show-keycodes` or `sudo evtest`.

river/command/map.zig
ifreund: > I think this would be very usable. If we dropped the `KEY_` prefix and make it case-insensitive, it would result in the same name as the keysym for most mappings in the default init file. So users with other layouts could use the default init file without changing much.

Sounds good to me, using `std.ascii` to uppercase whatever string the user provides and then prepending `KEY_` should be robust. Good point about the mouse buttons, let's leave that for a follow up patch if this one goes well.

The main challenge remaining is documentation and spending some more thought on whether `map-keysym`/`map-keycode` are the best names we can come up with. `map-keycode` *kinda* sounds like it expects a numerical code but I don't have a better suggestion.

pmkap: > Sounds good to me, using `std.ascii` to uppercase whatever string the user provides and then prepending `KEY_` should be robust. Good point about the mouse buttons, let's leave that for a follow up patch if this one goes well.

Ah yeah thinking about this, I remember there was a feature request some time ago (I think on IRC) to map mouse events to commands. With a dedicated `map-keycode` and corresponding `KeycodeMapping` type, I think this could become feasible. If we want to leave the door open for this we cannot drop the `BTN_` *and* `KEY_` prefixes. 

> The main challenge remaining is documentation and spending some more thought on whether `map-keysym`/`map-keycode` are the best names we can come up with. `map-keycode` _kinda_ sounds like it expects a numerical code but I don't have a better suggestion.

The only alternative name I can come up with on the fly is `map-event` or. Or we simplify as `map-sym` and `map-code`. I also feel like we could leave `map` as is.

pmkap: or even `map-evdev`

ifreund: I think leaving `map` as is and introducing `map-code` is what I'd like to go with. There's no strong need for a breaking change here and `map-code` avoids the term "keycode" which usually refers to an xkbcommon keycode not a linux input event code.

pmkap: Yes I agree!

MaxVerevkin:
Maybe this is a more "correct" / universal implementation, but I liked `-to-code` more. For example, I could do

```
riverctl map -to-code normal Super T spawn 'alacritty'
```

Now I have to do

```
riverctl map-code normal Super F spawn 'alacritty'
```

Notice that it's 'F', not 'T'. A possible workaround (on the user side) is to define a bunch of variables, [like I'm doing with sway](https://github.com/MaxVerevkin/dotfiles/blob/0bff59e3d67aa92834355691106212446056965b/sway/.config/sway/config#L20-L47), but still, it requires some work.

ifreund:
@MaxVerevkin Could you give some details about your setup? What layout does your physical keyboard have? At what level have you configured the workman layout? Through xkbcommon?

MaxVerevkin:
@ifreund 

> What layout does your physical keyboard have?

It's plain qwerty.

> At what level have you configured the workman layout? Through xkbcommon?

I use this to configure layouts: `export XKB_DEFAULT_LAYOUT="us(workman),ru"`

ifreund:
> I use this to configure layouts: `export XKB_DEFAULT_LAYOUT="us(workman),ru"`

Ah thanks, I was kinda assuming that being able to configure keybindings using the letters on the physical keyboard would be acceptable for all use-cases.

One more question, why do you need the manual `set $key_t f` variables for sway? What's different about the semantics of their `--to-code` compared to what pmkap developed here?

MaxVerevkin:
> letters on the physical keyboard

I don't actually have physical labels and I don't remember qwerty layout :)

> One more question, why do you need the manual set $key_t f variables for sway? What's different about the semantics of their --to-code compared to what pmkap developed here?

Actually, I don't know, and I was surprised when I was told that sway should automatically convert stuff. I have a ` ~/.config/sway/config.d/` directory with various configs, one of which sets the layout to workman, and one sets the keybindings. Apparently the layout was set after keybindings...

MaxVerevkin:
That makes sense, thanks for the explanation! I'll try it out later.

ifreund:
I think I see the possibility for a better, or at least more powerful/explicit, interface here. What if we could specify a keyboard layout as an optional argument to the `map` command? Then when translating input event code to keysym to check for an active mapping we would use the specified layout instead of the currently active layout. If no layout was specified then we would just use the currently active one.

If implementing this functionality like that doesn't have other issues I'm not aware of, this would eliminate the need for ambiguous `keysym` -> `keycode` translation and I'm *pretty* sure support all possible uses cases.

pmkap:
> I think I see the possibility for a better, or at least more powerful/explicit, interface here. What if we could specify a keyboard layout as an optional argument to the `map` command? Then when translating input event code to keysym to check for an active mapping we would use the specified layout instead of the currently active layout. If no layout was specified then we would just use the currently active one.
> 
> If implementing this functionality like that doesn't have other issues I'm not aware of, this would eliminate the need for ambiguous `keysym` -> `keycode` translation and I'm _pretty_ sure support all possible uses cases.

I like the idea, but to cater to *everybody's* needs we would probably need more fields as described here https://xkbcommon.org/doc/current/structxkb__rule__names.html , and not only the layout?

Leon-Plickat:
If I am not mistaken, implementing such a flag for `map` likely shares some code with setting the active keymap via `riverctl`. So maybe it makes sense to work on that first?

pmkap:
> If I am not mistaken, implementing such a flag for `map` likely shares some code with setting the active keymap via `riverctl`. So maybe it makes sense to work on that first?

I think the common code here would mostly be the parsing of the `xkb_rule_names`.

ifreund:
> I like the idea, but to cater to _everybody's_ needs we would probably need more fields as described here https://xkbcommon.org/doc/current/structxkb__rule__names.html , and not only the layout?

Hmm, that's a very good question, I'm not sure we do actually. As I understand it, the use-case for keysym mappings that use a layout other than the currently active layout is to allow for switching the layout dynamically at runtime while keeping the same keybindings relative to the physical keyboard that one expects using one's primary layout. 

As far as I know, xkbcommon doesn't provide a builtin way to modify the other options in the rule names struct with a key combination at runtime, just switch between layouts. I suppose we may want to add such functionality to river as well in the future, but it doesn't seem to be a major pain point so far, xkbcommon's XKB_DEFAULT_FOO environment variables have proven to be more flexible than I had initially expected.

As always though, I could be missing something or just be wrong :D It's certainly happened before in this discussion, who knew that keyboard configuration was so complex :/

ifreund:
> If I am not mistaken, implementing such a flag for `map` likely shares some code with setting the active keymap via `riverctl`. So maybe it makes sense to work on that first?

I looked very briefly at xkbcommon's API in relation to this and I don't think so. How we interpret the keycode for our key bindings is orthogonal to the keymap we send to clients. If I'm not mistaken, all we would need to do to implement what I've proposed would be to translate layout name to index with `xkb_keymap_layout_get_index()` and then use `xkb_keymap_key_get_syms_by_level()` instead of `xkb_keymap_key_get_syms()`.

pmkap:
Just to clarify: Do you want to

1. let the user define a whole new keymap with new `xkb_rule_names` for translating the mappings, or
2. let the user choose *one* group of the default layout, e.g. when `XKB_DEFAULT_LAYOUT=us,de,fr` choose one of the three?

MaxVerevkin:
> 1. let the user define a whole new keymap with new xkb_rule_names for translating the mappings, or

If this is going to be the case, will it look something like this?

```
riverctl map -layout=us(workman) normal Super T spawn alacritty
riverctl map -layout=us normal Super 1 set-focused-tags 1
```

ifreund:
> Just to clarify: Do you want to
>     1. let the user define a whole new keymap with new `xkb_rule_names` for translating the mappings, or 
>     2. let the user choose _one_ group of the default layout, e.g. when `XKB_DEFAULT_LAYOUT=us,de,fr` choose one of the three for translation? 
> Edit: I'd prefer 1.

Option 2 there is what I had in mind, but I'm certainly open to have my mind changed, I obviously don't yet know what the best path forward here is. Why would you prefer option 1?

pmkap:
> Option 2 there is what I had in mind, but I'm certainly open to have my mind changed. I obviously don't yet know what the best path forward here is. Why would you prefer option 1?

Mainly because I thought eventually we set the layout (or rather all `xkb_rule_names`) through riverctl and this would make option 2 impossible. Like described in `sway-input(5)`. 

But I am not convinced anymore. Like you said above, the XKB_DEFAULT_* environment variable scheme seems to be fully sufficient.

pmkap:
> If this is going to be the case, will it look something like this?
> 
> ```
> riverctl map -layout=us(workman) normal Super T spawn alacritty
> riverctl map -layout=us normal Super 1 set-focused-tags 1
> ```

Yes that's how I understand it. But the layout needs to be set by `XKB_DEFAULT_LAYOUT`.

Just one question here: Is the `(workman)` part of `us(workman)` in your example the `variant` as set by `XKB_DEFAULT_VARIANT`?

MaxVerevkin:
> Is the (workman) part of us(workman) in your example the variant as set by XKB_DEFAULT_VARIANT?

No, I use `XKB_DEFAULT_LAYOUT="us(workman)"`.

pmkap:
These two things seem to be somewhat equivalent (see https://xkbcommon.org/doc/current/structrxkb__layout.html)

* XKB_DEFAULT_LAYOUT="us(workman)"
* XKB_DEFAULT_LAYOUT="us" XKB_DEFAULT_VARIANT="workman"

pmkap:
> I looked very briefly at xkbcommon's API in relation to this and I don't think so. How we interpret the keycode for our key bindings is orthogonal to the keymap we send to clients. If I'm not mistaken, all we would need to do to implement what I've proposed would be to translate layout name to index with `xkb_keymap_layout_get_index()` and then use `xkb_keymap_key_get_syms_by_level()` instead of `xkb_keymap_key_get_syms()`

I was reading a bit in the xkbcommon docs. One thing that apparently would be a bit unintuitive with `xkb_keymap_layout_get_index()` is that it needs the full layout name.

E.g. when using `XKB_DEFAULT_LAYOUT=us,de`, one would need to call `xkb_keymap_layout_get_index("English (US)")` to get index 0 and `xkb_keymap_layout_get_index("German")` to get 1.

Maybe we could let the users specify the group index instead of confusing them with different naming schemes?

Leon-Plickat:
On Wed Apr 20, 2022 at 11:05 AM CEST, Peter Kaplan wrote:
> These two things seem to be somewhat equivalent (see
> https://xkbcommon.org/doc/current/structrxkb__layout.html)
>
> * XKB_DEFAULT_LAYOUT="us(workman)"
> * XKB_DEFAULT_LAYOUT="us" XKB_DEFAULT_VARIANT="workman"

That mirrors XKB internals. Variants are implemented as just "normal"
layouts that include the "original" and overwrite a few definitions.
`<layout>(<variant>)` is the canonical name of variant layout
definitions.

pmkap:
> That mirrors XKB internals. Variants are implemented as just "normal" layouts that include the "original" and overwrite a few definitions. `<layout>(<variant>)` is the canonical name of variant layout definitions.

Ah yeah I was confused by that.

Also, the names used with the environment variables are only to locate the layout in the filesystem and they are not stored in the resulting xkb_keymap, if I understand correctly.

Leon-Plickat:
On Wed Apr 20, 2022 at 12:13 PM CEST, Peter Kaplan wrote:
> Also, the names used with the environment variables are only to locate
> the layout in the filesystem and they are not stored in the resulting
> xkb_keymap, if I understand correctly.

Not quite. XKB has a few locations where layout defintion files are
stored. It parses all of them, always, and builds an internal layout
database. One of those files can in fact hold multiple layouts, as is
common with variants.

I can't remember what happens server-side and what client-side though,
all my knowledge on this comes from implementing my own layout variant.

river/Keyboard.zig
```diff
@@ -85,54 +78,36 @@ fn handleKey(listener: *wl.Listener(*wlr.Keyboard.event.Key), event: *wlr.Keyboa
...
-        } else if (self.seat.handleMapping(sym, modifiers, released)) {
-            handled = true;
-            break;
+    // Hide cursor when typing
+    for (keysyms) |sym| {
+        if (server.config.cursor_hide_when_typing == .enabled and
+            !released and
+            !isModifier(sym))
+        {
+            self.seat.cursor.hide();
```
MaxVerevkin, line 91: There should be a `break;`, I think

pmkap: Yes, thanks :)

MaxVerevkin:
Tested, works for me :)

river/command/map.zig
```diff
@@ -316,32 +323,58 @@ const OptionalArgsContainer = struct {
...
+        } else if (mem.eql(u8, args[j], "-repeat")) {
             parsed.repeat = true;
-            i += 1;
+            parsed.i += 1;
+        } else if (mem.eql(u8, args[j], "-layout")) {
+            // TODO: Figure out how we want the interface to look like.
+            // Currently, this uses the full layout name e.g. "English (Colemak)",
+            // because this is what `xkb_keymap_layout_get_index()` provides.
+            // But maybe the short name e.g. "us(colemak)" would be better because
+            // it's already used for XKB_DEFAULT_LAYOUT. We could also parse both.
```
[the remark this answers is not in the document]
pmkap: I also like the short names better. I've just looked in how to do this. According to this https://xkbcommon.org/doc/current/xkbcommon_8h.html#ad37512642806c55955e1cd5a30efcc39 , the short names are not present in the compiled keymap. So I think we would need to create a temporary keymap from the argument and use `xkb_keymap_layout_get_name(0)` on this to get the full name 

Would that overhead be acceptable to you or do you maybe have a better idea?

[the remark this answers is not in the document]
MaxVerevkin: > Hmm, I guess a third option would be to use layout indexes directly here.

Isn't that's how it was implemented at some point? If this is the simplest option, then why not? I don't think there is a reason to define a mapping for a layout that is not defined in `XKB_DEFAULT_LAYOUT`.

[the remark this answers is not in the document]
pmkap: > Isn't that's how it was implemented at some point? If this is the simplest option, then why not? I don't think there is a reason to define a mapping for a layout that is not defined in `XKB_DEFAULT_LAYOUT`.

Yeah I had this before I implemented parsing the layout name.

IMO, the advantage of a name is that the user can change the environment variable (e.g. swap order of layouts) without the need to touch the river init.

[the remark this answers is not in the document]
pmkap: It would still not be possible to use a layout that's not in XKB_DEFAULT_LAYOUT.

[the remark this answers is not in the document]
ifreund: > It would still not be possible to use a layout that's not in XKB_DEFAULT_LAYOUT.

I think this is actually a pretty strong argument in favor of sticking to layout indexes here as it makes it clear that this is selecting from the list of layouts defined by `XKB_DEFAULT_LAYOUT`. Accepting a name does make it feel like any layout name should work and I don't think we want to support that.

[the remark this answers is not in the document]
pmkap: Alright then, I reverse this :) Saves us a lot of parsing code :)

river/Seat.zig
```diff
@@ -342,17 +342,49 @@ pub fn handleViewUnmap(self: *Self, view: *View) void {
...
+    // map -layout 1 normal Super+Shift Ampersand ...
+    // map -layout 1 normal Super 1 ... #Shift is consumed to produce 1
+    //
+    // The question now is: Do we want to have a defined behaviour which one of these
+    // mappings would be triggered in such a case?
+    // If so, we would need to loop over the mappings-list multiple times. Up to 4
+    // times if we want to set a clear priority-list.
+
+    // For now, do everything in one wipe, which means any of these mappings could be
+    // the one that is triggered
```
ifreund, line 385: Hmm, if there's an easy way to detect and forbid conflicting mappings as in the example my first choice would be to forbid this entirely. If not, I think the simplest and least surprising thing we can do would be to activate *both* mappings. That is, instead of returning early from the for loop below we would continue in case there is another match. I'd be fine with leaving the ordering between the mappings undefined for simplicity. How's that sound?

pmkap: Sounds good. Implemented it that way.. However, only one mapping can be repeated. We could make `seat.repeating_mapping` a linked list if we cared about this edge case...

pmkap: Ah yeah makes sense :)

ifreund (approved):
LGTM! Thanks all your work on the many iterations it took to get this use-case supported. I'm quite happy with how it's turned out :)

MaxVerevkin:
I'm glad I helped :) Thanks for this really nice feature :+1:

PR #585, opened 2022-04-11
river: implement smart borders

tleydxdy:
Turn off borders when the stacked view takes up all the usable space.

### DIFF
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -41,11 +41,20 @@ pub const HideCursorWhenTypingMode = enum {
     enabled,
 };
 
+pub const SmartBordersMode = enum {
+    disabled,
+    /// Turn off border when the view takes up all usable space
+    enabled,
+};
+
 /// Color of background in RGBA (alpha should only affect nested sessions)
 background_color: [4]f32 = [_]f32{ 0.0, 0.16862745, 0.21176471, 1.0 }, // Solarized base03
 
 /// Width of borders in pixels
 border_width: u32 = 2,
+
+/// The selected smart_border mode
+smart_borders: SmartBordersMode = .disabled,
 
 /// Color of border of focused window in RGBA
 border_color_focused: [4]f32 = [_]f32{ 0.57647059, 0.63137255, 0.63137255, 1.0 }, // Solarized base1
--- a/river/LayoutDemand.zig
+++ b/river/LayoutDemand.zig
@@ -85,15 +85,20 @@ pub fn pushViewDimensions(self: *Self, output: *Output, x: i32, y: i32, width: u
     // The client pushed too many dimensions
     if (self.views < 0) return;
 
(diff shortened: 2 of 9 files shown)


tleydxdy:
A few things still need to be worked on:
- maybe the mode need to have more descriptive names than "enabled" (in case there are other smart borders style than what's implemented in the future)
- I'm not sure if I catched all the docs/completion stuff or if they are ordered correctly

river/Box.zig
```diff
@@ -40,3 +40,7 @@ pub fn toWlrBox(self: Self) wlr.Box {
         .height = @intCast(c_int, self.height),
     };
 }
+
+pub fn eql(self: Self, other: Self) bool {
```
pmkap, line 44: This is not necessary since `std.meta.eql` can be used.

river/Config.zig
```diff
@@ -41,12 +41,21 @@ pub const HideCursorWhenTypingMode = enum {
...
+};
+
 /// Color of background in RGBA (alpha should only affect nested sessions)
 background_color: [4]f32 = [_]f32{ 0.0, 0.16862745, 0.21176471, 1.0 }, // Solarized base03
 
 /// Width of borders in pixels
 border_width: u32 = 2,
 
+/// The selected smart_border mode
+smart_borders: SmartBordersMode = .enabled,
```
pmkap, line 57: I think this should be disabled by default.

river/LayoutDemand.zig
```diff
@@ -85,15 +85,20 @@ pub fn pushViewDimensions(self: *Self, output: *Output, x: i32, y: i32, width: u
...
-    // Here we apply the offset to align the coords with the origin of the
-    // usable area and shrink the dimensions to accomodate the border size.
-    const border_width = server.config.border_width;
-    self.view_boxen[self.view_boxen.len - @intCast(usize, self.views)] = .{
-        .x = x + output.usable_box.x + @intCast(i32, border_width),
-        .y = y + output.usable_box.y + @intCast(i32, border_width),
-        .width = if (width > 2 * border_width) width - 2 * border_width else width,
-        .height = if (height > 2 * border_width) height - 2 * border_width else height,
-    };
+    if (server.config.smart_borders == .enabled and x == 0 and y == 0 and width == output.usable_box.width and height == output.usable_box.height) {
```
pmkap, line 88: Isn't `x == 0 and y == 0` redundant?

ifreund:
Sorry, but I'm not interested in maintaining this feature.

This will be possible to achieve cleanly without modifying river's code eventually if my ideas on window management APIs pan out. Until then it's important for me to keep the amount of configuration options river has as small as possible to prevent inhibiting refactoring.

savolla:
This feature is a must. I can't figure out how to implement "cleanly" using current version of river...

PR #589, opened 2022-04-17
Cursor: fix cursor doesn't remain hidden under some situations

tleydxdy:


ifreund:
What situations? How can I reproduce the bug here and why is this the proper fix?

tleydxdy:
two situation I found was when the client sets a custom cursor, e.g. wait for cursor to be hidden and switch to a tag with foot under the cursor, and two when focus is cleared, e.g. switch to a empty tag.

ifreund:
Thanks for the bug report and attempted fix! See e0246c04104fad95573f07c3bfc4547c7ce1d116 for what I came up with.

PR #590, opened 2022-04-17
Cursor: add on-focus-change option 

tleydxdy:
Warp the cursor to the center of the focused view if it is not in the bounding box of that view.

This helps the user to keep track of their cursor when they mostly use the keyboard and the cursor becomes hidden most of the time, also helps trackpad/trackpoint users.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -105,6 +105,7 @@ pressed_count: u32 = 0,
 hide_cursor_timer: *wl.EventSource,
 
 hidden: bool = false,
+may_need_warp: bool = false,
 
 axis: wl.Listener(*wlr.Pointer.event.Axis) = wl.Listener(*wlr.Pointer.event.Axis).init(handleAxis),
 frame: wl.Listener(*wlr.Cursor) = wl.Listener(*wlr.Cursor).init(handleFrame),
@@ -1050,6 +1051,9 @@ pub fn checkFocusFollowsCursor(self: *Self) void {
 /// the target view of a cursor operation potentially being moved to a non-visible tag,
 /// becoming fullscreen, etc.
 pub fn updateState(self: *Self) void {
+    if (self.may_need_warp) {
+        self.warp();
+    }
     if (self.shouldPassthrough()) {
         self.mode = .passthrough;
         var now: os.timespec = undefined;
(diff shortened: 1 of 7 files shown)


tleydxdy:
not sure if https://github.com/tleydxdy/river/commit/558ca5e6efc75e006a052d900fd17bf4c4315780 this is a good idea to add to this. It make it so my cursor isn't warped when I click on the bar to switch tags etc.

Toesmasher:
Would love to see this merged. Is it possible to also add a 'both' or 'all' option to set-cursor-warp? That is, warp the cursor to the new output on an output swap, and to new focus on a focus change?

tleydxdy:
that is how it works currently

adamcstephens:
This is a crucial feature for me, and I've considered switching away from river without it. I use keyboard view switching heavily, but on occasion my trackpoint on my laptop will register a movement while typing. This jumps focus back to wherever the mouse was last, generally in an editor, and then I'm typing random characters into whatever file I was editing.

Leon-Plickat:
Hey everyone. Just dropping by to let y'all know that a lack of attention for this PR does not equal a lack of interest, but rather a lack of time to take a proper look at it. As always, please use the fancy reaction things on the top comment instead of writing +1 comments; Keeping the discussion on-topic makes the review process *a lot* simpler. Thanks!

river/Cursor.zig
```diff
@@ -1095,3 +1099,44 @@ fn passthrough(self: *Self, time: u32) void {
...
+        .disabled => return,
+        .@"on-output-change" => layout_box,
+        .@"on-focus-change" => switch (self.seat.focused) {
+            .layer, .lock_surface, .none => layout_box,
+            .view => |view| wlr.Box{
+                .x = layout_box.x + view.current.box.x,
+                .y = layout_box.y + view.current.box.y,
+                .width = @intCast(c_int, view.current.box.width),
+                .height = @intCast(c_int, view.current.box.height),
+            },
```
ifreund, line 1119: These `@intCast()`s are no longer needed.

river/Cursor.zig
```diff
@@ -105,6 +105,7 @@ pressed_count: u32 = 0,
 hide_cursor_timer: *wl.EventSource,
 
 hidden: bool = false,
+needWarp: bool = false,
```
ifreund, line 108: Struct fields should be snake case. I think `keyboard_focus_changed` might also be more helpful to future readers.

river/Cursor.zig
```diff
@@ -1095,3 +1099,44 @@ fn passthrough(self: *Self, time: u32) void {
...
+        },
+    };
+    const usable_box = self.seat.focused_output.usable_box;
+    const usable_layout_box = wlr.Box{
+        .x = layout_box.x + usable_box.x,
+        .y = layout_box.y + usable_box.y,
+        .width = usable_box.width,
+        .height = usable_box.height,
+    };
+    if (!layout_box.containsPoint(self.wlr_cursor.x, self.wlr_cursor.y) or (usable_layout_box.containsPoint(self.wlr_cursor.x, self.wlr_cursor.y) and !box.containsPoint(self.wlr_cursor.x, self.wlr_cursor.y))) {
```
ifreund, line 1135: This should be split across multiple lines for readability, `zig fmt` allows line breaks after binary operators such as `and`/`or`.

Also, what is the purpose of the useable_layout_box check here?

river/Seat.zig
```diff
@@ -273,6 +277,8 @@ pub fn setFocusRaw(self: *Self, new_focus: FocusTarget) void {
             PointerConstraint.warpToHint(&self.cursor);
             constraint.sendDeactivated();
             self.cursor.constraint = null;
+        } else {
+            self.cursor.warp();
```
ifreund, line 281: Why does this call warp directly instead of setting the flag?

river/Cursor.zig
```diff
@@ -1095,3 +1099,44 @@ fn passthrough(self: *Self, time: u32) void {
         self.clearFocus();
     }
 }
+
+pub fn warp(self: *Self) void {
+    self.needWarp = false;
+    if (self.seat.focused_output == &server.root.noop_output) return;
+    // Warp pointer to center of the focused view/output (In layout coordinates) if enabled.
+    var layout_box: wlr.Box = undefined;
```
ifreund, line 1107: `output_layout_box` would be a bit more descriptive if you don't mind the verbosity.

river/Cursor.zig
```diff
@@ -1095,3 +1099,44 @@ fn passthrough(self: *Self, time: u32) void {
...
+    server.root.output_layout.getBox(self.seat.focused_output.wlr_output, &layout_box);
+    var box = switch (server.config.warp_cursor) {
```
ifreund, line 1109:
```suggestion
    const box = switch (server.config.warp_cursor) {
```
`target_box` might also be a bit more helpful as a name here.

ifreund (changes requested):
Hello, sorry for the long wait. I haven't had as much time for river recently as I would like. 

Anyhow, this feature looks fine and nearly merge ready. Just a few style issues/cleanup to take care of.

river/Cursor.zig
tleydxdy: I think it might be more versatile to name it this way in case there's some other case that needs mouse warpping

river/Cursor.zig
tleydxdy: it prevents the mouse to be warpped if it was outside the usable area, this prevents the mouse being yanked when say you click the bar to select tag

river/Seat.zig
tleydxdy: here it's for when the output is empty, so it won't trigger an layout event

ifreund: Whether the output is empty or not is irrelevant, focus is double buffered state that is applied through the transaction system.

Leon-Plickat:
A comment explaining that would be nice though.

ifreund (approved):
Thanks for the patch!

Rebased on to master and made a few minor tweaks to comments and naming before merging :)

PR #598, opened 2022-04-29
Fix typo in natural-scroll

Cephian:
Run the following commands to see where the typo appears:
```bash
riverctl input my_device natural-scroll enabled
riverctl list-input-configs
```


ifreund (approved):
Thanks!

PR #599, opened 2022-04-30
ci: Remove some unused packages

novakne:
Add quiet flag to wget

### DIFF
--- a/.builds/alpine.yml
+++ b/.builds/alpine.yml
@@ -16,7 +16,6 @@ packages:
   - xwayland
   - meson
   - scdoc
-  - tar
   - wget
   - xz
 sources:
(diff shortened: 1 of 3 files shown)


.builds/alpine.yml
```diff
@@ -16,7 +16,6 @@ packages:
   - xwayland
   - meson
   - scdoc
-  - tar
```
ifreund, line 19: it seems strange to remove this when we use tar below to extract the zig tarball...

novakne: I imagine it is already a deps of something, I saw that is required by abuild, the program to build Alpine package.
I've been using it like so on my project and also for nfm without any problem, and as you can see the ci for this PR is succesfull, but if you prefer I can add it back.

tiosgz: I believe that you should explicitly install/require everything that you explicitly use. (And leave implicit everything that you don’t use explicitly.)

ifreund: Yeah, I'd prefer to keep everything we directly use explicitly listed. That way our CI doesn't break if the transitive dependencies brought in change.

Issue #604, opened 2022-05-13
"mod+Shift Space" binding also triggers "mod Space"


dnkl:
With this `init`:

```sh
#!/bin/bash

mod="Mod4"
riverctl map normal $mod Space spawn fuzzel
riverctl map normal $mod+Shift Space toggle-float

riverctl default-layout rivertile
exec rivertile
```
pressing `Mod4+Shift+Space` **both** toggles float, **and** spawns fuzzel.

River: [0.2.0-dev-90b3764](https://github.com/riverwm/river/commit/90b37645cf7744d8cc2a000fd3d74caf95bce204)

dnkl:
XKB settings:
```sh
export XKB_DEFAULT_LAYOUT=se,us
export XKB_DEFAULT_OPTIONS=ctrl:swapcaps_and_switch_layout
```

ifreund:
This is caused by #577.

Before that patch we would check keybinds first with modifiers taken into account. Then, if no match was found, we would check keybindings against the base layer as if modifiers had no effect. This allows the user to map `Mod4+Shift 1` instead of `Mod4+Shift !` on a US layout for example.

In #577 this behavior was changed to always check both and run all matching mappings even if the check with modifiers taken into account was successful.

I think the old behavior is more intuitive in this case, cc @pmkap.

pmkap:
Apparently, I didn't take all things into account with my patch. I'm going to look into it.

pmkap:
Apparently, some layouts behave differently when pressing `Shift Space`. It seems that [`xkb_state_key_get_consumed_mods2()`](https://xkbcommon.org/doc/current/group__state.html#ga5b119b99542fcad1787119191ba9e286) reports `Shift` as consumed in this case, so that `Shift Space` translates to `Space`. Not only on the Swedish layout but also others. Finnish, Norwegian and Indian for example.

Before #577, we didn't have this problem because we didn't translate (consume) the modifiers. This was a `TODO` https://github.com/riverwm/river/blob/ae349b0ce45b44b79a0a93e89c2cd05d24bb63d4/river/Keyboard.zig#L88 

Sway translates the modifiers, so `Mod4+Shift Space` triggers a mapping for `Mod4 Space` on the above mentioned layouts. Only that sway never executes two mappings so probably nobody ever noticed this. 

So the way I see it we can either not translate modifiers like it was before #577. Or we need to check mappings without translating the keys *first*.

ifreund:
[...]

PR #606, opened 2022-05-15
mappings: Do not translate keys with xkb

pmkap:
When checking keys for matching mappings, previously we did two checks:

1. Keysyms translated by xkb.
2. Raw keysyms

This commit removes the first check, so only the second is checked.

As discussed in #604

### DIFF
--- a/river/Mapping.zig
+++ b/river/Mapping.zig
@@ -69,7 +69,7 @@ pub fn deinit(self: Self) void {
 pub fn match(
     self: Self,
     keycode: xkb.Keycode,
-    modifiers_raw: wlr.Keyboard.ModifierMask,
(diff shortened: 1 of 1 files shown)


pmkap:
@ifreund
What do think of an additional comment to explain this? Maybe something like

```
// We deliberately choose not to translate keysyms and modifiers with xkb,
// because of strange behavior that xkb shows for some layouts when
// pressing Shift Space.
```

ifreund:
>  What do think of an additional comment to explain this?

Yes, that'd be great. It be nice to mention the example of the Swedish layout explicitly. Also as far as we know `Shift Space` is only one example of this strange behavior, there might be more.

ifreund (approved):
LGTM, Thanks!

PR #607, opened 2022-05-24
example/init: fix typos

jnvrs:


ifreund (approved):
Thanks for the patch!

Issue #455, opened 2021-10-28
DMENU ignores keystrokes if no Xwayland clients are present and in focus.


Shinyzenith:
Description: dmenu doesn't accept input unless a wayland client exists on the screen which is actively checking for keystrokes. Example, if dmenu_run is being executed on a tag where no other app is open, none of the keystrokes will be accepted, works when any other app is opened ( example, terminal ) .
( Sorry for incorrect terminology, if any)

Any possible workaround? ( dmenu works flawlessly under sway so I assume one already exists. )

HeavyRain266:
If you want native dmenu replacement, I suggest you to use [bemenu](https://github.com/Cloudef/bemenu) instead of bothering with launcher programs running through XWayland.

Shinyzenith:
I know but bemenu has some missing features, hence this gh issue.

Shinyzenith:
I'm trying to fix this by getting the surface and then calling wlr.XdgSurface.fromWlrSurface but I keep getting this error, if anyone can provide a fix that would be amazing,

Shinyzenith:
![image](https://user-images.githubusercontent.com/60808802/158050094-2c0ce502-3124-4abf-9a74-9283857b2f65.png)

![image](https://user-images.githubusercontent.com/60808802/158050097-db2863d3-70f9-46ea-ab9c-be352024e0bd.png)

Shinyzenith:
https://github.com/riverwm/river/pull/569 should hopefully close this issue.

PR #608, opened 2022-05-28
xwayland-unmanaged: handle focus on map/cursor button

zakvf:
This should resolve https://github.com/riverwm/river/issues/455 by adding keyboard focus on map for unmanaged Xwayland views that want it (based on the relevant TODO, using wlroot's heuristic). It also adds keyboard focus when clicking on an unmanaged Xwayland view, for some consistency with layer surfaces. This was initially based on the closed PR https://github.com/riverwm/river/pull/569.

I have tested this with dmenu, as in the linked issue, and polybar (at top and bottom of screen), and it works on my end. However, there is a problem: I am unable to select an unmanaged Xwayland surface if I have another view in fullscreen. This appears to be an issue with `surfaceAt()` failing in `xwaylandUnmanagedSurfaceAt()` when in fullscreen, but I am not sure.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -579,7 +579,7 @@ pub fn surfaceAt(self: Self) ?SurfaceAtResult {
     if (layerSurfaceAt(output.getLayer(.overlay).*, ox, oy)) |s| return s;
 
     if (fullscreen_view) |view| {
-        if (build_options.xwayland) if (xwaylandUnmanagedSurfaceAt(ly, lx)) |s| return s;
+        if (build_options.xwayland) if (xwaylandUnmanagedSurfaceAt(lx, ly)) |s| return s;
         var sx: f64 = undefined;
         var sy: f64 = undefined;
         if (view.surfaceAt(ox, oy, &sx, &sy)) |found| {
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -315,7 +315,13 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
                     self.seat.focus(null);
                 }
             },
-            .xwayland_unmanaged => assert(build_options.xwayland),
+            .xwayland_unmanaged => |xwayland_unmanaged| {
+                if (build_options.xwayland) {
+                    self.seat.setFocusRaw(.{ .xwayland_unmanaged = xwayland_unmanaged });
+                } else {
+                    unreachable;
+                }
+            },
         }
         _ = self.seat.wlr_seat.pointerNotifyButton(event.time_msec, event.button, event.state);
 
--- a/river/XwaylandUnmanaged.zig
+++ b/river/XwaylandUnmanaged.zig
@@ -78,19 +78,32 @@ fn handleMap(listener: *wl.Listener(*wlr.XwaylandSurface), xwayland_surface: *wl
 
     xwayland_surface.surface.?.events.commit.add(&self.commit);
 
-    // TODO: handle keyboard focus
-    // if (wlr_xwayland_or_surface_wants_focus(self.xwayland_surface)) { ...
+    if (self.xwayland_surface.overrideRedirectWantsFocus()) {
+        server.input_manager.defaultSeat().setFocusRaw(.{ .xwayland_unmanaged = self });
+    }
 }
 
 /// Called when the surface is unmapped and will no longer be displayed.
 fn handleUnmap(listener: *wl.Listener(*wlr.XwaylandSurface), _: *wlr.XwaylandSurface) void {
     const self = @fieldParentPtr(Self, "unmap", listener);
 
-    // Remove self from the list of unmanged views in the root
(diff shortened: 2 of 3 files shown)


river/Cursor.zig
```diff
@@ -315,7 +315,13 @@ fn handleButton(listener: *wl.Listener(*wlr.Pointer.event.Button), event: *wlr.P
...
                 }
             },
-            .xwayland_unmanaged => assert(build_options.xwayland),
+            .xwayland_unmanaged => |xwayland_unmanaged| {
+                if (build_options.xwayland) {
+                    self.seat.setFocusRaw(.{ .xwayland_unmanaged = xwayland_unmanaged });
+                } else {
+                    unreachable;
+                }
+            },
```
ifreund, line 324:
```suggestion
            .xwayland_unmanaged => |xwayland_unmanaged| {
                assert(build_options.xwayland);
                self.seat.setFocusRaw(.{ .xwayland_unmanaged = xwayland_unmanaged });
            },
```

ifreund (commented):
> However, there is a problem: I am unable to select an unmanaged Xwayland surface if I have another view in fullscreen. This appears to be an issue with surfaceAt() failing in xwaylandUnmanagedSurfaceAt() when in fullscreen, but I am not sure.

Do you have a simple program that can be used to reliably reproduces this? The code in `surfaceAt()` looks correct to me.

river/Cursor.zig
zakvf: I had this originally, but `zig build` failed with the following error (including in the CI):

```
./river/Cursor.zig:320:64: error: expected type '*XwaylandUnmanaged', found 'void'
                self.seat.setFocusRaw(.{ .xwayland_unmanaged = xwayland_unmanaged });
```

zakvf:
> Do you have a simple program that can be used to reliably reproduces this? The code in `surfaceAt()` looks correct to me.

I did this just by fullscreening foot and running dmenu/polybar. (I don't believe starting before or after the fullscreen change affected the focus on click.) With debug logging, I saw that it was entering the `xwaylandUnmanagedSurfaceAt()` function (both when fullscreen and not, as expected) and that the surface was available. But the function did not return an Xwayland unmanaged surface result when a view was in fullscreen, even when with the same screen coordinates it worked otherwise (i.e. when not in fullscreen).

I will check it again now.

river/Cursor.zig
ifreund: Ah, yeah that makes sense. Sorry for the noise.

zakvf:
> I will check it again now.

I think I have found the issue---but I am not sure why it is happening. The dmenu/polybar surface appears to rotate 90 degrees when a view is in fullscreen. For me, when I spawned them at the top of the screen, on fullscreen their 'collision box' ran down the left side of the screen. When I spawned either of them instead at the bottom of the screen, their collision box ran down the centre-right of the screen. In the case of polybar, notably, the clickable centre module moved down towards the bottom edge of the collision box instead of being at the horizontal centre of the screen.

ifreund:
Ah, yep. Looks like I swapped the x/y arguments here:
```
        if (build_options.xwayland) if (xwaylandUnmanagedSurfaceAt(ly, lx)) |s| return s;
```

ifreund:
Thanks for debugging that, feel free to add the fix to this PR as a separate commit :)

zakvf:
> Ah, yep. Looks like I swapped the x/y arguments here:

That was a lot simpler than I expected!

> feel free to add the fix to this PR as a separate commit :)

Added, and works on my end now. Thanks!

river/XwaylandUnmanaged.zig
```diff
@@ -78,19 +78,35 @@ fn handleMap(listener: *wl.Listener(*wlr.XwaylandSurface), xwayland_surface: *wl
...
     xwayland_surface.surface.?.events.commit.add(&self.commit);
 
-    // TODO: handle keyboard focus
-    // if (wlr_xwayland_or_surface_wants_focus(self.xwayland_surface)) { ...
+    if (wlr.XwaylandSurface.overrideRedirectWantsFocus(self.xwayland_surface)) {
+        if (server.input_manager.seats.first) |seat_node| {
+            const seat = &seat_node.data;
+            seat.setFocusRaw(.{ .xwayland_unmanaged = self });
+        }
+    }
```
ifreund, line 86:
```suggestion
    if (self.xwayland_surface.overrideRedirectWantsFocus()) {
        server.input_manager.defaultSeat().setFocusRaw(.{ .xwayland_unmanaged = self });
    }
```

ifreund (commented):
Just one style cleanup then this should be ready to merge!

river/XwaylandUnmanaged.zig
zakvf: Done :)

ifreund (approved):
LGTM, Thanks for the PR!

Shinyzenith:
@zakvf thank you for patching this. My knowledge of server side wayland is still immature so Iearnt quite a bit from this patch.

Issue #609, opened 2022-05-29
xwayland views do not support override-redirect changes


zakvf:
Some of my Xwayland applications start as override-redirect views, then change to normal views (specifically, the games Kero Blaster and ZeroRanger (the demo) running through Wine). https://github.com/WayfireWM/wayfire/issues/684 and https://github.com/swaywm/sway/issues/4600 are about the same unhandled override-redirect change. I don't think river handles this, though, so it is difficult to interact with these windows since they are unmanaged.

ifreund:
Fixed in 6ef97eea241ecc5bec32e0cf820c3644692f5554

PR #610, opened 2022-05-29
xwayland: allow views to change override-redirect setting

zakvf:
Closes https://github.com/riverwm/river/issues/609. This is based on https://github.com/swaywm/sway/pull/5650, so it recreates the Xwayland view as the opposite type (`XwaylandView`/`XwaylandUnmanaged`) on a change to the override-redirect setting. The creation of the new view is done through the server. Because the unmapping and destruction of the old view and the mapping of the new view are not tied to their respective listeners in this case, I have moved their behaviour out of their handler functions, which now call the new map/unmap/destroy functions.

I tested this with both games I mentioned in the linked issue (Kero Blaster and the ZeroRanger demo, both running through Wine) and it works as expected for me.

### DIFF
--- a/river/Server.zig
+++ b/river/Server.zig
@@ -35,6 +35,7 @@ const Root = @import("Root.zig");
 const StatusManager = @import("StatusManager.zig");
 const View = @import("View.zig");
 const ViewStack = @import("view_stack.zig").ViewStack;
+const XwaylandView = @import("XwaylandView.zig");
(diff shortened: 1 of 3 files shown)


ifreund:
Thanks for the patch! I ended up implementing this a bit differently in 6ef97eea241ecc5bec32e0cf820c3644692f5554 as I saw the opportunity to pair this with some refactoring I'd been meaning to get around to in 960afa89ec41315640afd4dc2d9d84835923c44d.

PR #611, opened 2022-05-29
Eat mapped keys

### LINKED ISSUE #444: Don't send key release (/press) events for press (/release) binds
IRC:

```
[11:30:16] leon-p: I wonder if we should stop sending key release events to clients for keys which had a keymap on press.
[11:31:04] leon-p: some clients (looking at you, firefox) also handle release events even if they never got the press event, which gets a bit awkward
[11:31:22] emersion: you mean when switching keymap?
[11:33:04] leon-p: no, "keymap" is rivers word for "keybind"
[11:33:58] emersion: eh that's confusing :P
[11:34:00] leon-p: my specific problem: I have something bound to Mod4+Escape and when I watch videos fullscreen in firefox, FF leaves fullscreen when just receiving the "release" event from the bind
[11:34:31] emersion: but yeah, if you eat the press event, i think it's a good thing to also eat the release event
[11:40:39] dnkl: Second that
[11:42:02] leon-p: who wants to open the issue? :P
[11:43:21] novakane: well you're the “collaborator” so :P
[11:52:06] ifreund: leon-p: we call it a mapping not a keymap no?
[11:52:52] ifreund: skipping the release event if river eats the press makes sense. We can't really do the same the other way around thoug
[11:53:15] ifreund: that is, we can't skip the press event if river eats the release do to a mapping
[11:54:03] pkap joined the channel.
[11:55:58] leon-p: why not? When a press event is handled river can find out if there is a release mapping
[11:57:04] leon-p: it only gets messy when the order in which keys get released is different then the press order, but that is for both cases
[11:59:01] leon-p: f.e. a user presses Super, Alt, j, then releases Alt, j, super. We can not just block a Mod4+Mod1+j release being send, as it would be a Mod4+j release. We would need to block all j releases.
[11:59:19] ifreund: ok yeah forget what I said, though it's technically possible for mappings to change between press and release 
```

tiosgz:
For a normal (on press) mapping, do not send the press and release events and remove the key from `wl_keyboard::enter`. For a `-release`-only mapping, send the key to the client as if nothing was mapped.

Fixes #444

Draft status because cleanup (removing debug logs, TODO/FIXME) is still needed. Functionality-wise, the PR is done.

### DIFF
--- a/river/Seat.zig
+++ b/river/Seat.zig
@@ -31,6 +31,7 @@ const DragIcon = @import("DragIcon.zig");
 const Cursor = @import("Cursor.zig");
 const InputManager = @import("InputManager.zig");
 const Keyboard = @import("Keyboard.zig");
+const KeycodeSet = @import("KeycodeSet.zig");
 const Switch = @import("Switch.zig");
 const Mapping = @import("Mapping.zig");
 const LayerSurface = @import("LayerSurface.zig");
@@ -253,12 +254,20 @@ pub fn setFocusRaw(self: *Self, new_focus: FocusTarget) void {
 
         // Send keyboard enter/leave events and handle pointer constraints
         if (target_surface) |wlr_surface| {
-            if (self.wlr_seat.getKeyboard()) |keyboard| {
+            if (self.wlr_seat.getKeyboard()) |wlr_keyboard| {
+                var keycodes = KeycodeSet{
+                    .items = wlr_keyboard.keycodes,
+                    .len = wlr_keyboard.num_keycodes,
+                };
+
+                const keyboard = @intToPtr(*Keyboard, wlr_keyboard.data);
+                keycodes.subtract(keyboard.eaten_keycodes);
+
                 self.wlr_seat.keyboardNotifyEnter(
                     wlr_surface,
-                    &keyboard.keycodes,
-                    keyboard.num_keycodes,
-                    &keyboard.modifiers,
+                    &keycodes.items,
+                    keycodes.len,
+                    &wlr_keyboard.modifiers,
                 );
             } else {
                 self.wlr_seat.keyboardNotifyEnter(wlr_surface, null, 0, null);
@@ -349,8 +358,25 @@ pub fn handleViewUnmap(self: *Self, view: *View) void {
     if (self.focused == .view and self.focused.view == view) self.focus(null);
 }
 
+/// Is there a user-defined mapping for passed keycode, modifiers and keyboard state?
+pub fn hasMapping(
+    self: *Self,
+    keycode: xkb.Keycode,
+    modifiers: wlr.Keyboard.ModifierMask,
+    released: bool,
+    xkb_state: *xkb.State,
+) bool {
+    const modes = &server.config.modes;
+    for (modes.items[self.mode_id].mappings.items) |*mapping| {
+        if (mapping.match(keycode, modifiers, released, xkb_state)) {
+            return true;
+        }
+    }
+    return false;
+}
+
 /// Handle any user-defined mapping for passed keycode, modifiers and keyboard state
-/// Returns true if the key was handled
+/// Returns true if at least one mapping was run
 pub fn handleMapping(
     self: *Self,
     keycode: xkb.Keycode,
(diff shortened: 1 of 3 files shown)


river/Seat.zig
```diff
@@ -342,31 +351,45 @@ pub fn handleViewUnmap(self: *Self, view: *View) void {
...
     xkb_state: *xkb.State,
-) bool {
+) !*std.SinglyLinkedList(*Mapping) {
+    const MappingList = std.SinglyLinkedList(*Mapping);
+    const list = try util.gpa.create(MappingList);
+    list.* = .{};
+    errdefer {
+        while (list.popFirst()) |node| util.gpa.destroy(node);
+        util.gpa.destroy(list);
+    }
```
ifreund, line 368: I don't think these allocations are necessary. Rather they are an indication that we should restructure the `getMappings()` API.

It seems to me that given the current usage in `Keyboard.handleKey()` we could have one function to check if any mapping is matched. If that function returns true, add the keycode to the eaten keycodes set. Then finally run something similar to the old version of this function to run all matching mappings, asserting that at least one mapping was run.

river/set.zig
```diff
@@ -0,0 +1,51 @@
...
+                if (self.len > 0) self.items[idx] = self.items[self.len];
+
+                return true;
+            };
+
+            return false;
+        }
+
+        /// Removes other's contents from self (if present)
+        pub fn subtract(self: *Self, other: *Self) void {
```
ifreund, line 47:
```suggestion
        pub fn subtract(self: *Self, other: Self) void {
```

river/set.zig
```diff
@@ -0,0 +1,51 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+const assert = @import("std").debug.assert;
+
+pub fn Set(comptime T: type, comptime capacity: usize) type {
```
ifreund, line 19: This doesn't really need to be generic to be honest, I wouldn't mind calling it KeycodeSet and hardcoding u32 as the element type plus a reasonable capacity.

river/set.zig
```diff
@@ -0,0 +1,51 @@
...
+    return struct {
+        const Self = @This();
+
+        items: [capacity]T = undefined,
+        len: usize = 0,
+
+        pub fn add(self: *Self, new: T) void {
+            for (self.items[0..self.len]) |item| if (new == item) return;
+
+            assert(self.len < capacity);
```
ifreund, line 29: Is this ok to assert? Where is the current limit of 32 enforced?

ifreund (commented):
The overall approach looks good. I've left comments where I think things could be cleaned up/improved.

river/set.zig
tiosgz: 32 is a reasonable limit copied [from wlroots](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_keyboard.h#L65) (i don’t know if there’s also a hard limit somewhere further down the stack).

There were three options:
- assume that it’s really never crossed and let runtime safety checks handle it
- assert and raise the limit once it’s crossed
- misbehave and don’t eat the key

Anyway, a comment would indeed do no harm.

ifreund: > assume that it’s really never crossed and let runtime safety checks handle it

Code written this way is by definition buggy IMO. Runtime safety checks (UB in release fast/small builds remember) should never be triggered by correct code.

> misbehave and don’t eat the key

This is what wlroots does, and I think it makes sense for river to do the same here since we depend on wlroots. Adding a comptime assertion that we match wlroot's limit wouldn't be a bad idea either:

```zig
comptime {
     assert(capacity == std.meta.fieldInfo(wlr.Keyboard, .keycodes).Array.len);
}
```

ifreund: An error log message if we do happen to misbehave due to this limit would make sense as well.

tiosgz: Sorry, i can’t seem to make sense of this part of zig. It’s giving me `error: no member named 'Array' in struct 'std.builtin.StructField'`. How should it be?

ifreund: Yes that's right, sorry for the confusion.

ifreund (approved):
I'm quite happy with this solution, nice work!

PR #612, opened 2022-05-29
Xwayland: add handling of override redirect listeners

zakvf:
Handling of the `set_override_redirect` listener in `create()` and `handleDestroy()` was missed in https://github.com/riverwm/river/commit/6ef97eea241ecc5bec32e0cf820c3644692f5554, so Xwayland views don't actually switch between being managed and unmanaged.


ifreund (approved):
Whoops, my bad. Thanks for the fix!

PR #614, opened 2022-05-31
river-status: add mode event to seat status

ifreund:
This allows clients such as a status bar to display the currently active
mode.


tiosgz:
`StatusManager.zig:42` still reports v2.

ifreund:
> `StatusManager.zig:42` still reports v2.

That's what I get for not testing... Fixed, Thanks!

PR #617, opened 2022-06-04
Abort if default init is not executable

Leon-Plickat:
since it's such a common mistake.

### DIFF
--- a/river/main.zig
+++ b/river/main.zig
@@ -140,6 +140,12 @@ fn defaultInitPath() !?[:0]const u8 {
     };
 
     os.accessZ(path, os.X_OK) catch |err| {
+        if (err == error.PermissionDenied) {
+            if (os.accessZ(path, os.R_OK)) {
+                std.log.err("failed to run init executable {s}: the file is not executable", .{path});
+                os.exit(1);
+            } else |_| {}
+        }
         std.log.err("failed to run init executable {s}: {s}", .{ path, @errorName(err) });
         util.gpa.free(path);
         return null;


river/main.zig
```diff
@@ -140,9 +140,20 @@ fn defaultInitPath() !?[:0]const u8 {
     };
 
     os.accessZ(path, os.X_OK) catch |err| {
-        std.log.err("failed to run init executable {s}: {s}", .{ path, @errorName(err) });
-        util.gpa.free(path);
-        return null;
+        switch (err) {
+            if (err == error.NotExecutable) {
+            }
```
ifreund, line 145:
```suggestion
```

river/main.zig
```diff
@@ -140,9 +140,20 @@ fn defaultInitPath() !?[:0]const u8 {
...
+            error.PermissionDenied => {
+                std.log.err("failed to run init executable {s}: the file is not executable", .{path});
+                util.gpa.free(path);
```
ifreund, line 148: No need to free if exiting
```suggestion
```

river/main.zig
```diff
@@ -140,9 +140,17 @@ fn defaultInitPath() !?[:0]const u8 {
     };
 
     os.accessZ(path, os.X_OK) catch |err| {
-        std.log.err("failed to run init executable {s}: {s}", .{ path, @errorName(err) });
-        util.gpa.free(path);
-        return null;
+        switch (err) {
+            error.PermissionDenied => {
+                std.log.err("failed to run init executable {s}: the file is not executable", .{path});
```
ifreund, line 145: Hmm, I don't think this message is always accurate. EPERM might be returned because one of the directories in the path isn't readable by the user.

Leon-Plickat: If we test for `R_OK` first and then for `X_OK`, we could handle that case as well.

river/main.zig
```diff
@@ -139,12 +139,17 @@ fn defaultInitPath() !?[:0]const u8 {
...
         std.log.err("failed to run init executable {s}: {s}", .{ path, @errorName(err) });
         util.gpa.free(path);
         return null;
     };
 
+    os.accessZ(path, os.X_OK) catch {
+        std.log.err("failed to run init executable {s}: the file is not executable", .{path});
+        os.exit(1);
+    };
+
```
ifreund, line 152: Instead of always doing two syscalls, could we only do the `R_OK` check in the error path?

river/main.zig
```diff
@@ -140,6 +140,11 @@ fn defaultInitPath() !?[:0]const u8 {
     };
 
     os.accessZ(path, os.X_OK) catch |err| {
+        if (err == error.PermissionDenied) blk: {
+            os.accessZ(path, os.R_OK) catch break :blk;
+            std.log.err("failed to run init executable {s}: the file is not executable", .{path});
+            os.exit(1);
+        }
```
ifreund, line 147: Sorry for the bikeshedding, but this `break :blk` control flow feels a bit convoluted. I think an `if` statement would be better here:
```suggestion
        if (err == error.PermissionDenied) {
            if (os.accessZ(path, os.R_OK)) {
                std.log.err("failed to run init executable {s}: the file is not executable", .{path});
                os.exit(1);
            } else |_| {}
        }
```
Other than that this looks ready to merge :)

Leon-Plickat: Huh, I always forget `if` works on error unions as well.

ifreund (approved):
LGTM, Thanks!

PR #619, opened 2022-06-05
 Fix the mod key issues with the default init script

Infinitybeond1:
This pr solves the issue with the default init script calling 'Super' as a modifer which doesnt work, now the script maps the variable '$Super' to Mod4 which solves the problem

### DIFF
--- a/example/init
+++ b/example/init
@@ -8,113 +8,116 @@
 # See the river(1), riverctl(1), and rivertile(1) man pages for complete
 # documentation.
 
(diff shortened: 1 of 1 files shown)


novakne:
That is not an error,  in your issue you mention that you use void wich has river  0.1.1, the `Super` modifier was added later, you need to use the example init from 0.1.1, this one works as intended with river master.

Infinitybeond1:
Ohh sorry for the inconvenience

PR #653, opened 2022-07-13
rivertile: Add command to mutate padding during runtime

skovati:
Hi all! This PR adds a new command to `rivertile`, `padding` that allows updating padding during runtime. This was a feature I had wanted from river for a while, and had maintained a separate patch, but figured I should upstream this feature if it was desired.

My implementation treats padding as a boolean, allowing `send-layout-cmd rivertile padding {on,off,toggle}` to toggle padding (both view _and_ outer) on and off. I debated implementing this command to take an integer as input, similar to how main-count and ratio work, and would be happy to rework the PR if this format is preferred.

I'm new to `river` development and zig in general, so please do respond with (constructive) criticism! Of course, if this feature is out of scope for a minimal layout generator like `rivertile`, I absolutely understand.

### DIFF
--- a/doc/rivertile.1.scd
+++ b/doc/rivertile.1.scd
@@ -62,6 +62,10 @@ These commands may be sent to rivertile at runtime with the help of
 	_value_. Note that the ratio will always be clamped to the range
 	0.1 to 0.9.
 
+*padding* [*on*|*off*|*toggle*]
+	Set or toggle view and outer padding. On or off will set the padding
+	accordingly, while toggle will switch between the two.
+
 # EXAMPLES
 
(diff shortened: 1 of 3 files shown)


skovati:
Updated to run each commit with `zig fmt`

skovati:
Updated to set padding values before toggling `padding`, otherwise toggling once had no effect.

ifreund (commented):
Sorry, but I consider `rivertile` feature complete at this point. I encourage people wanting more features to use one of the more featureful options listed on [this wiki page](https://github.com/riverwm/river/wiki/Community-Layouts) or write their own.

The implementation quality of your patch on the other hand looks good, so nice work there!

PR #654, opened 2022-07-14
river: keep view within bounds when sending to output

Leon-Plickat:
There is still a problem with floating fullscreen views.

### DIFF
--- a/river/View.zig
+++ b/river/View.zig
@@ -283,8 +283,33 @@ pub fn sendToOutput(self: *Self, destination_output: *Output) void {
         self.foreign_toplevel_handle.?.outputEnter(destination_output.wlr_output);
     }
 
+    self.output = destination_output;
+    const dimensions = destination_output.getEffectiveResolution();
+
+    if (self.pending.float) {
+        // Adapt dimensions of view to new output. Only necessary when floating,
+        // because for tiled views the output will be rearranged, taking care
+        // of this.
+        if (self.pending.fullscreen) self.pending.box = self.post_fullscreen_box;
+        const border_width = if (self.draw_borders) @intCast(i32, server.config.border_width) else 0;
+        self.pending.box.width = math.min(
+            self.pending.box.width,
+            @intCast(i32, dimensions.width) - (2 * border_width),
+        );
+        self.pending.box.height = math.min(
+            self.pending.box.height,
+            @intCast(i32, dimensions.height) - (2 * border_width),
+        );
+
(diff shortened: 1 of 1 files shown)


ifreund (commented):
Why isn't this implemented in `View.sendToOutput()`?

> There is still a problem with floating fullscreen views.

What's the problem exactly? Is this something present on master you don't intend to fix in this patch? Is it a regression?

Leon-Plickat:
If you'd rather have this in View.sendToOutput() I can move it there. Wasn't really sure where to put it.

The problem with fullscreen floating views exists on master and I do intend to fix it; Forgot to mark the PR as draft. The "problem" is simply that a different box needs to be changed.

ifreund:
>If you'd rather have this in View.sendToOutput() I can move it there. Wasn't really sure where to put it.

Yeah it'd be better to have there. This isn't the only place we move views between outputs, it's also done when evacuating an output due to it being unplugged for example and we should apply the same logic to view sizes in that situation.

> The problem with fullscreen floating views exists on master and I do intend to fix it; Forgot to mark the PR as draft. The "problem" is simply that a different box needs to be changed.

Ah makes sense, thanks.

river/View.zig
```diff
@@ -283,16 +283,44 @@ pub fn sendToOutput(self: *Self, destination_output: *Output) void {
...
+            self.pending.box.width,
+            @intCast(i32, dimensions.width) - (2 * border_width),
+        );
+        self.pending.box.height = math.min(
+            self.pending.box.height,
+            @intCast(i32, dimensions.height) - (2 * border_width),
+        );
+
+        // A client may change its view constraints when it becomes aware that
+        // its view is displayed on a different output.
```
ifreund, line 305: It really shouldn't, at least not for xdg shell views. Have you observed a client doing this? Anyways, the client hasn't event received the enter event when this code is run so I don't think we need this call.

Otherwise this patch LGTM :)

ifreund (approved):

river/View.zig
Leon-Plickat: I haven't found any clients yet that do so, but I haven't actively looked for one.

My though was that if the other output has f.e. a different scaling factor the client may wish to chance it's constraints. Technically it is also valid for a client to arbitrarily (at least from our perspective) change them based on output.

However, if the client does not receive the enter event until later, it makes no difference either way. I'll remove this.

ifreund (approved):
LGTM. This is a very nice bit of polish, thanks!

PR #655, opened 2022-07-19
LayoutDemand: handle too many view dimensions correctly

tiosgz:
The condition was wrong for the first extra view (0 isn't < 0), plus
the counting stopped at that moment.


ifreund (approved):
LGTM, thanks!

PR #660, opened 2022-07-27
Add installation section

Drsheppard01:


wiki-me:
Repology has a widget that shows in what distro's a certain software is available (see the cataclysm dda [readme](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/README.md) for an example).

That could be a more concise and easily maintainable alternative.

Drsheppard01:
firstly, Repology does not hide Fedora Copr, and also I find river in Gentoo, which is not on repology

There are examples where both repology and software installation are used: 

https://github.com/peltoche/lsd
https://github.com/johnfactotum/foliate

wiki-me:
Are you sure it's on gentoo?, i don't see it on the gentoo search page ([here](https://packages.gentoo.org/packages/search?q=river)), looks like some kind of bug in the gentoo website.

Leon-Plickat:
We already have a repology badge in the readme.

---

I don't really think I get the point of this PR. All sane package managers have a search functionality, so we do not really need to list all available packages. Listing the exact commands is also weird to me: We should not need to explain people how to use the package managers of their distributions.

Drsheppard01:
I wanted to show other installation options that are not affected by repology (Fedora COPR, Gentoo)

PR #664, opened 2022-08-02
river: block SIGPIPE on start to avoid weird exits

tleydxdy:
When river or wlroots write to a closed socket it could generate SIGPIPE causing the whole desktop to seemingly "crash" with no error log of any kind. So ignore the SIGPIPE and let write fail with EPIPE.

### DIFF
--- a/river/main.zig
+++ b/river/main.zig
@@ -41,7 +41,18 @@ const usage: []const u8 =
 
 pub var server: Server = undefined;
 
+fn sa_handler(_: c_int) callconv(.C) void {}
+
 pub fn main() anyerror!void {
+    // ignore SIGPIPE so we don't get killed when socket unexpectedly closes (thanks xwayland)
+    // use our own handler instead of SIG_IGN so we don't leak this when execve()
+    const act = os.Sigaction{
+        .handler = .{ .handler = sa_handler },
+        .mask = os.empty_sigset,
+        .flags = 0,
+    };
+    os.sigaction(os.SIG.PIPE, &act, null);
+
     // This line is here because of https://github.com/ziglang/zig/issues/7807
     const argv: [][*:0]const u8 = os.argv;
     const result = flags.parse(argv[1..], &[_]flags.Flag{


marler8997:
Related: https://github.com/ziglang/zig/pull/11982

river/main.zig
```diff
@@ -42,6 +42,13 @@ const usage: []const u8 =
 pub var server: Server = undefined;
 
 pub fn main() anyerror!void {
+    const act = os.Sigaction{
+        .handler = .{ .sigaction = os.SIG.IGN },
+        .mask = os.empty_sigset,
+        .flags = os.SA.SIGINFO,
+    };
+    os.sigaction(os.SIG.PIPE, &act, null);
+
```
ifreund, line 51: If I understand the `signal(7)` man page and related man pages correctly, setting the signals disposition to ignore like this is different from using `sigprocmask` to block the signal. I think your code here is correct and we do indeed want to ignore this signal rather than block it, but that means we need to un-ignore it in a forked child process before calling `execve()`. See the two existing appearances of `sigprocmask()` in river's code.

ifreund (commented):
Thanks for the patch, this definitely seems like the right thing to do!

river/main.zig
tleydxdy: I just copied over the function from that zig issue actually, skimmed some docs and promptly missed this detail. now I did some more digging and I'm not sure if what the code does is correct any more (using siginfo and SIG_IGN at the same time) https://lists.gnu.org/archive/html/bug-gnulib/2008-06/msg00196.html

tleydxdy: So I suppose the better move as add our own handler that ignores the signal? this way it is also automatically reset when fork-exec right?

ifreund: > So I suppose the better move is add our own handler that ignores the signal? this way it is also automatically reset when fork-exec right?

Ah, yeah that seems to do what we want without requiring us to remember anything on fork.

river/main.zig
```diff
@@ -41,7 +41,16 @@ const usage: []const u8 =
...
+fn sa_handler(_: c_int) callconv(.C) void {}
+
 pub fn main() anyerror!void {
+    const act = os.Sigaction{
+        .handler = .{ .handler = sa_handler },
+        .mask = os.empty_sigset,
+        .flags = 0,
+    };
+    os.sigaction(os.SIG.PIPE, &act, null);
+
```
ifreund, line 53: It would be nice to have a short comment in the source code to explain why we ignore SIGPIPE and why we use an empty handler instead of `SIG_IGN`. Other than that this LGTM!

ifreund:
Thanks for this! I ended up digging a little deeper and pushed another commit to master as well: 1a9cba2aa91a1ab4085aa5670644ee9349d0595a

Issue #383, opened 2021-08-05
[feature request][river-layout] Tag parameter for user_command event


Leon-Plickat:
It is not guaranteed that the next layout_demand event after a user_command
event has the same active tags. As an example, a user could trigger a
user_command while no views are visible, then switch to a different tag set
which has active tags.

Obviously the active tags of the previous layout_demand may also be differently.

Therefore, it is currently impossible to correctly implement a layout generator
which has user commands apply only to the currently active tag set.

Simply adding the currently active tags as a parameter to the user_command event
would solve this. It would add effectively no complexity serverside and even
reduce complexity and improve correctnes client-side in the case of layouts with
per-tag configuration of any kind.

Leon-Plickat:
Layouts with per-tag values are IMO a valid use-case and river should support external clients wanting to do this in a correct, non-hacky fashion.

Leon-Plickat:
If I am not mistaken, this would only require a new version of the interface, not the entire protocol, so it is non-breaking.

novakne:
So this issue can be closed

PR #671, opened 2022-08-13
river-layout: add user_command_tags event

Leon-Plickat:
Implements #383

This patch increases the version of the layout protocol to v4

It is not guaranteed that the next layout_demand event after a user_command
event has the same active tags (for example when there are no views visible).
As an example, a user could trigger a user_command while no views are visible,
then switch to a different tag set which has active views. The active tags of
the previous layout_demand may also be different.

Therefore it is impossible to correctly implement a layout generator which has
user commands apply only to the currently active tag set, which is solved by
this patch.

### DIFF
--- a/river/command/layout.zig
+++ b/river/command/layout.zig
@@ -77,5 +77,5 @@ pub fn sendLayoutCmd(
     } else return;
 
     layout.layout.sendUserCommand(args[2]);
-    output.arrangeViews();
+    if (layout == output.current.layout) output.arrangeViews();
 }


ifreund (changes requested):
I'm not too excited about breaking changes to the layout protocol right now, could we not implement this with a new tags event sent immediately before `user_command` if the client has bound instead version 2 of the `river_layout_v3` interface instead?

Leon-Plickat:
Yeah that would work as well.

Although I'd argue that since the protocol has still a pretty limited
amount of users, we can do breaking changes without affecting too many
users. Since an additional event for tags is inconsistent with the
layout_demand event, we would likely change this later again.

ifreund:
> Although I'd argue that since the protocol has still a pretty limited amount of users, we can do breaking changes without affecting too many users. Since an additional event for tags is inconsistent with the layout_demand event, we would likely change this later again.

I'd still rather not break things we don't have to. Perhaps the new event should be called `user_command_tags` or similar to make it clear that it is only related to the `user_command` event not the `layout_demand`.

ifreund:
Also we have no real knowledge of how many users the layout protocol has. Yes there are only a few public layout generators but many more people may have their own custom forks or private layout generators. This is something we've explicitly encouraged people to do, and I don't want to break things for them unless necessary.

Leon-Plickat:
Well, we apparently have differing opinions on API stability :D

Anyway, I did it the way you suggested now.

river/command/layout.zig
```diff
@@ -76,6 +76,7 @@ pub fn sendLayoutCmd(
         if (mem.eql(u8, layout.namespace, target_namespace)) break layout;
     } else return;
 
+    layout.layout.sendUserCommandTags(output.pending.tags);
```
ifreund, line 79: We need to check the version of the layout object here before sending the event. Otherwise, we will send it to clients that have bound the interface at version 1 and cause them to be terminated.

protocol/river-layout-v3.xml
```diff
@@ -174,8 +174,23 @@
 
         A layout_demand will be sent after this event if the compositor is
         currently using this layout object to arrange the output.
+
+        If version 2 or higher of the river_layout_v3 object is bound, the
+        user_command_tags event is guaranteed to be send before the user_command
```
ifreund, line 179:
```suggestion
        user_command_tags event is guaranteed to be sent directly before the user_command
```

protocol/river-layout-v3.xml
```diff
@@ -174,8 +174,23 @@
...
+        event.
       </description>
       <arg name="command" type="string"/>
     </event>
+
+    <event name="user_command_tags" since="2">
+      <description summary="a command sent by the user">
+        If version 2 or higher of the river_layout_v3 object is bound, this
+        event will be send before every user_command event. This allows layout
```
ifreund, line 188:
```suggestion
        event will be sent directly before every user_command event. This allows layout
```

protocol/river-layout-v3.xml
```diff
@@ -174,8 +174,23 @@
...
+        generators to be aware of the active tags when a user command is send.
```
ifreund, line 189:
```suggestion
        generators to be aware of the active tags when a user command is sent.
```

rivertile/main.zig
```diff
@@ -298,6 +298,7 @@ const Output = struct {
                     .bottom => layout.commit("rivertile - bottom", ev.serial),
                 }
             },
+            else => {},
```
ifreund, line 301: I think it's a bit more readable to have this explicit.
```suggestion
            .user_command_tags => {},
```

ifreund (commented):
Looking good, just a few minor fixes/cleanups.

ifreund (approved):
LGTM, Thanks!

Issue #680, opened 2022-08-22
riverctl input does not work on startup


fedeizzo:
### Basic info

distro: NixOS 22.05 (Kernel 5.19.0)
river version: 0.1.3

### What is the problem?
River correctly calls the file under `./config/river/init` but `riverctl input <name> ...` commands are not executed (or are executed with no results) and the log does not show any error. The command `riverctl list-input-configs` does not return anything.
If I call the script after river initialization commands are executed without any problem and `riverctl list-input-configs` shows the settings.

### Configuration
I tried two different approaches. The first one with the input name hardcoded

```bash
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad accel-profile none
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad click-method clickfinger
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad drag enabled
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad drag-lock enabled
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad disable-while-typing enabled
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad natural-scroll enabled
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad tap enabled
riverctl input 1267:12572:DLL0945:00_04F3:311C_Touchpad scroll-method two-finger
```

and the second one with a for loop

```bash
for pad in $(riverctl list-inputs | grep -i 'touchpad\|mouse'); do
    riverctl input $pad accel-profile none
    riverctl input $pad click-method clickfinger
    riverctl input $pad drag enabled
    riverctl input $pad drag-lock enabled
    riverctl input $pad disable-while-typing enabled
    riverctl input $pad natural-scroll enabled
    riverctl input $pad tap enabled
    riverctl input $pad scroll-method two-finger
done
```

### Logs
```
info: initializing server
info(wlroots): [libseat] [libseat/libseat.c:73] Seat opened with backend 'seatd'
info(wlroots): [libseat] [libseat/backend/seatd.c:212] Enabling seat
info(wlroots): [backend/session/session.c:110] Successfully loaded libseat session
debug(wlroots): [backend/session/session.c:385] Ignoring '/dev/dri/card1': not a KMS device
info(wlroots): [backend/backend.c:220] Found 1 GPUs
[...]

Issue #665, opened 2022-08-06
Launching apps at start


TheMainMMan:
I tried many ways and different methods to get this right, I just don't understand where the problem is. I can do all that from the config file or point to the sh script which then launches startup apps, doesn't matter, the end result is the same (they don't execute), so when I launch river there is nothing, empty desktop, just the mouse cursor is shown, then I have to reload configuration file (Supet+Shitf+R) and then it works as it should have the first time, it's so bizarre and weird.
It's not a big problem, I just have to restart it, but it bugs me, any ideas where the problem is ?

sum-catnip:
Well, what did you try? Im doing riverctl spawn xyz without any issues in my init file.

TheMainMMan:
Yeah, `riverctl spawn "~/.config/river/autostart.sh"` , or one by one `riverctl spawn mako &` etc.
I think that maybe config file isn't read at all on launch, only on reload.
It's located at `~/.config/river/`
I'm yet to test that, would have to change reload binding for example, because that's the only one that works on launch (that I tried), not much else can be done anyway, if the config file isn't read.

Leon-Plickat:
Show us your config, otherwise we can just guess.

sum-catnip:
is the config called "init" and is it executable? Give us an ls -la inside your ~/.config/river directory

TheMainMMan:
Yes it's executable and it's called "init" , problem is something else most likely, which I realized after I made this post, like I said when I press reload config binding, it works. My guess is that it doesn't read the config file when it first starts, using Gentoo here and I start river from the terminal after the login with `dbus-run-session river`, I used several WM's on Wayland before River and I didn't had that problem, so I don't know. I can't test that now but I will pretty soon, to change the reload binding to something else, that way I can know for sure that it doesn't read the config file at all on launch, for some reason.

Leon-Plickat:
If you want others to help you, upload your config. All the information you have provided so far isn't particularly relevant.

If you do not show us your config, I'll have to close this issue as it won't go anywhere.

TheMainMMan:
Sure, here it is. Current one is with the separate sh file for startup apps, but I also tried just spawning apps from the init file (`riverctl spawn xy`), same thing happens.

init
```
#!/bin/sh

# This is the example configuration file for river.
#
# If you wish to edit this, you will probably want to copy it to
# $XDG_CONFIG_HOME/river/init or $HOME/.config/river/init first.
#
[...]

PR #681, opened 2022-08-22
add note about layout generator in init

adamcstephens:
Add a note to give people some guidance that configuration should be done above the layout generator.

See: 
#665
#680

Happy to tweak the wording, but this could use some more clarity in my opinion.

### DIFF
--- a/example/init
+++ b/example/init
@@ -157,4 +157,6 @@ riverctl csd-filter-add app-id "gedit"
 # Set the default layout generator to be rivertile and start it.
 # River will send the process group of the init executable SIGTERM on exit.
 riverctl default-layout rivertile
+
+# The layout generator runs until river exits. All configuration should be above here.
 rivertile -view-padding 6 -outer-padding 6


adamcstephens:
I still think this would help some users, but given the lack of interest I'm closing it.

ifreund:
I feel like the most fool-proof solution would be to start rivertile in the background with `&`.

Leon-Plickat:
That depends on whether we want to adress the cause or the effect.

The problem is that we apparently don't communicate well enough that
the (default) init is just a normal shell script, judged by the amount
of people who have had this and similar problems.

So do we want to explain (again) that all the usual shell control
flow, with which the average user no doubt is already familiar with,
applies, or do we just silently work around this? I am fine with
either approach.

ifreund:
> So do we want to explain (again) that all the usual shell control
flow, with which the average user no doubt is already familiar with,
applies, or do we just silently work around this?

I think the best solution is to do both. I think it's been demonstrated that no matter how often/clearly something is documented some subset of users won't read it and just copy paste/try things out.

ifreund:
I just pushed e2a27b3a4eaea4dc302dabd9e111bdc6e42489d9 to master which starts rivertile in the background in the example init.

I tried a bit to come up with something short and enlightening to put in a comment, but didn't think any of my attempts were an improvement over the new master branch status quo. It's impossible for us to explain all of POSIX shell in our documentation, and certainly out of scope.

Thank you for opening this PR and starting this discussion! I'm going to close this now as the comment added no longer makes sense since rivertile is now started in the background. Feel free to open another if you see further improvements though :)

adamcstephens:
Thank you!

PR #686, opened 2022-08-30
view: log app-id on map and unmap

alex-courtis:
There is no means to determine app-id, as there is no wayland equivalent of xprop.

Log app-id as well as title.

### DIFF
--- a/river/View.zig
+++ b/river/View.zig
@@ -480,7 +480,7 @@ pub fn shouldTrackConfigure(self: Self) bool {
 
 /// Called by the impl when the surface is ready to be displayed
 pub fn map(self: *Self) !void {
-    log.debug("view '{s}' mapped", .{self.getTitle()});
+    log.debug("view app-id='{s}' title='{s}' mapped", .{self.getAppId(), self.getTitle()});
 
     {
         assert(self.foreign_toplevel_handle == null);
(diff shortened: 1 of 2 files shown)


river/View.zig
```diff
@@ -480,7 +480,7 @@ pub fn shouldTrackConfigure(self: Self) bool {
 
 /// Called by the impl when the surface is ready to be displayed
 pub fn map(self: *Self) !void {
-    log.debug("view '{s}' mapped", .{self.getTitle()});
+    log.debug("view app-id='{s}' title='{s}' mapped", .{self.getAppId(), self.getTitle()});
```
alex-courtis, line 483: We could change this to info to be a bit more user friendly.

ifreund:
[lswt](https://git.sr.ht/~leon_plickat/lswt) can be used to determine app-id and other properties.

Leon-Plickat:
Also note that this is a bit flaky, as some applications unfortunately set their app-id *after* they map their toplevels.

alex-courtis:
> [lswt](https://git.sr.ht/~leon_plickat/lswt) can be used to determine app-id and other properties.

Fantastic! OK if I add that to the man page @Leon-Plickat ?

alex-courtis:
> Also note that this is a bit flaky, as some applications unfortunately set their app-id _after_ they map their toplevels.

That's unfortunate. At least we will see it on unmap.

alex-courtis:
lswt works beautifully.

For the case of badly behaved apps (like zoom) that quickly open and close "notification" windows the user can do something like:

```sh
while true; do lswt; echo; sleep 0.5; done
```

Thanks for the review, closing.

PR #687, opened 2022-08-30
river: implement keyboard groups

Leon-Plickat:
Usage:
```
riverctl keyboard-group-create <group-name>
riverctl keyboard-group-add-keyboard <group-name> <input-device-identifier>
```

Still a few rough spots, but it works.

### DIFF
--- a/river/InputDevice.zig
+++ b/river/InputDevice.zig
@@ -108,7 +108,7 @@ fn handleDestroy(listener: *wl.Listener(*wlr.InputDevice), _: *wlr.InputDevice) 
 
     switch (device.wlr_device.type) {
         .keyboard => {
-            const keyboard = @fieldParentPtr(Keyboard, "device", device);
+            const keyboard = @fieldParentPtr(Keyboard, "provider", @ptrCast(*Keyboard.Provider, device));
             keyboard.deinit();
             util.gpa.destroy(keyboard);
         },
--- a/river/Seat.zig
+++ b/river/Seat.zig
@@ -41,6 +41,7 @@ const Switch = @import("Switch.zig");
 const View = @import("View.zig");
 const ViewStack = @import("view_stack.zig").ViewStack;
 const XwaylandOverrideRedirect = @import("XwaylandOverrideRedirect.zig");
+const KeyboardGroup = @import("KeyboardGroup.zig");
 
 const log = std.log.scoped(.seat);
 const PointerConstraint = @import("PointerConstraint.zig");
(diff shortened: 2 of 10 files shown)


Leon-Plickat:
Problem: River still updates the keymap whenever it sends a key event from a different keyboard. That is pretty expensive, so we need to come up with a way to prevent that for grouped keyboards.

river/KeyboardHandler.zig
```diff
@@ -0,0 +1,113 @@
...
+
+const KeycodeSet = @import("KeycodeSet.zig");
+const Seat = @import("Seat.zig");
+
+const server = &@import("main.zig").server;
+
+/// Pressed keys for which a mapping was triggered on press
+eaten_keycodes: KeycodeSet = .{},
+
+pub fn handleKey(self: *Self, seat: *Seat, wlr_device: *wlr.InputDevice, event: *wlr.Keyboard.event.Key) void {
```
ifreund, line 35: I don't think we actually need to factor this logic out of `Keyboard.zig`. Instead we could pass the `wlr_keyboard_group`'s "virtual" input device to `Seat.addDevice()`, creating a river `Keyboard` for the `wlr_keyboard_group`'s aggregate keyboard and using the exact same code paths.

What do you think of that approach?

ifreund (changes requested):

river/KeyboardHandler.zig
Leon-Plickat: I was initially convinced this would not work, hence the  current implementation. But can not remember why, so I'll give it a try.

river/Seat.zig
```diff
@@ -276,8 +279,10 @@ fn keyboardNotifyEnter(self: *Self, wlr_surface: *wlr.Surface) void {
             .len = wlr_keyboard.num_keycodes,
         };
 
-        const keyboard = @intToPtr(*Keyboard, wlr_keyboard.data);
-        keycodes.subtract(keyboard.eaten_keycodes);
+        if (wlr.KeyboardGroup.fromKeyboard(wlr_keyboard) == null) {
+            const keyboard = @intToPtr(*Keyboard, wlr_keyboard.data);
+            keycodes.subtract(keyboard.handler.eaten_keycodes);
+        }
```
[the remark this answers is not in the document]
Leon-Plickat: Oh, I think this is a remainder from before I wired up listeners to the virtual keyboards of keyboard groups.

river/InputDevice.zig
```diff
@@ -108,7 +108,7 @@ fn handleDestroy(listener: *wl.Listener(*wlr.InputDevice), _: *wlr.InputDevice)
 
     switch (device.wlr_device.type) {
         .keyboard => {
-            const keyboard = @fieldParentPtr(Keyboard, "device", device);
+            const keyboard = @fieldParentPtr(Keyboard, "provider", @ptrCast(*Keyboard.Provider, device));
```
Leon-Plickat, line 111: TBH I habe no idea if this will work. Probably no, thanks to alignment.

See: https://github.com/ziglang/zig/issues/6611

ifreund:
Thanks for the PR!

PR #691, opened 2022-09-12
map-button

### LINKED ISSUE #685: Feature Request: map mouse button
### Use Case

mod-middle-click to toggle floating

### Prior Art

* dwm [config.def.h](https://git.suckless.org/dwm/file/config.def.h.html#l99)
* i3 `bindsym --whole-window $mod1+button2 floating toggle`

### Options

1. Overload `map-pointer` to allow all commands
2. Create `map-button` to allow mapping button to a command
3. Allow `map` to take input event code

1 is unacceptable as `map-pointer` is specifically for mouse motions.

3 would be simplest for the user. Implementation would also be simpler than 2.

alex-courtis:
resolves #685 

Adds `map-button` and `unmap-button`: executes a command similar to `map`, using an event code instead of a keysym.

Does not feature `[-release|-repeat|-layout index]`.

Silently overrides `map-pointer`. Please advised whether a warning or automatic `unmap-pointer` is required.

First time using zig; pointers and corrections would be gratefully appreciated.

### DIFF
--- a/river/ButtonMapping.zig
+++ b/river/ButtonMapping.zig
@@ -26,13 +26,9 @@ event_code: u32,
 modifiers: wlr.Keyboard.ModifierMask,
 command_args: []const [:0]const u8,
 
-/// When set to true the mapping will be executed on button release rather than on press
-release: bool,
-
 pub fn init(
     event_code: u32,
     modifiers: wlr.Keyboard.ModifierMask,
-    release: bool,
     command_args: []const []const u8,
 ) !Self {
     const owned_args = try util.gpa.alloc([:0]u8, command_args.len);
@@ -44,7 +40,6 @@ pub fn init(
     return Self{
         .event_code = event_code,
         .modifiers = modifiers,
-        .release = release,
         .command_args = owned_args,
     };
 }
@@ -54,12 +49,11 @@ pub fn deinit(self: Self) void {
     util.gpa.free(self.command_args);
 }
 
-/// Compare mapping with given event code, modifiers and keyboard state
+/// Compare mapping with given event code and modifiers
 pub fn match(
     self: Self,
     event_code: u32,
     modifiers: wlr.Keyboard.ModifierMask,
-    release: bool,
 ) bool {
-    return event_code == self.event_code and std.meta.eql(modifiers, self.modifiers) and release == self.release;
+    return event_code == self.event_code and std.meta.eql(modifiers, self.modifiers);
 }
--- a/river/ButtonMapping.zig
+++ b/river/ButtonMapping.zig
@@ -1,0 +1,65 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 The River Developers
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, version 3.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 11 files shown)


alex-courtis:
### Testing Performed

```
riverctl map-button normal Super BTN_MIDDLE toggle-float

riverctl map-button normal Super+Control BTN_MIDDLE toggle-fullscreen

riverctl unmap-button normal Super+Control BTN_MIDDLE

riverctl unmap-button normal Super+Control BTN_MIDDLE

riverctl map-button normal Super+Control BTN_MIDDLE
error: not enough arguments

riverctl map-button normal Super+Control BTN_BLEH toggle-fullscreen
error: unknown button BTN_BLEH

riverctl map-button normal Super+Bleh BTN_MIDDLE toggle-fullscreen
error: invalid modifier 'Bleh'

riverctl map-button bleh Super+Control BTN_MIDDLE toggle-fullscreen
error: cannot add/remove mapping to/from non-existant mode 'bleh'

riverctl unmap-button normal Super+Control BTN_MIDDLE toggle-fullscreen
error: too many arguments

riverctl unmap-button normal Super+Control
error: not enough arguments
```

Same mapping with different commands for different modes: correct commands executed


Same mapping for map-button and map-pointer: map-button overrides map-pointer


Ensured button not sent to client on click


zsh completions but not bash/fish

river/ButtonMapping.zig
```diff
@@ -0,0 +1,59 @@
+// This file is part of river, a dynamic tiling wayland compositor.
+//
+// Copyright 2020 The River Developers
```
Leon-Plickat, line 3:
```suggestion
// Copyright 2022 The River Developers
```

river/Seat.zig
```diff
@@ -405,6 +405,23 @@ pub fn handleMapping(
...
 
+/// Handle any user-defined mapping for passed button and modifiers.
+/// Returns true if there was a mapping.
+pub fn handleButtonMapping(
+    self: *Self,
+    event_code: u32,
+    modifiers: wlr.Keyboard.ModifierMask,
+) bool {
+    const modes = &server.config.modes;
+    for (modes.items[self.mode_id].button_mappings.items) |*mapping| {
```
Leon-Plickat, line 416:
```suggestion
    for (modes.items[self.mode_id].button_mappings.items) |mapping| {
```

You only need a pointer if you want to modify the item.

Leon-Plickat:
Seems a bit weird to me to have both `map-buttom` and `map-pointer` TBH. Maybe it would make sense to combine the two?

alex-courtis:
> Seems a bit weird to me to have both `map-buttom` and `map-pointer` TBH. Maybe it would make sense to combine the two?

Sounds good. keyboard vs mouse can be inferred from keycode vs event code.
Looks like that was my initial idea #685 ...

`-release|-repeat` don't really make sense for buttons. Suggestion:
* mapping always occurs on press
* return an error if the user attempts to set those options

`-layout index` should be implementable.

river/Seat.zig
alex-courtis: Thank you!

Leon-Plickat:
> keyboard vs mouse can be inferred from keycode vs event code.

What keyboard? Both `map-pointer` and `map-button` are for the mouse.

alex-courtis:
> > keyboard vs mouse can be inferred from keycode vs event code.
> 
> What keyboard? Both `map-pointer` and `map-button` are for the mouse.

Whoops... I got that backwards. I was thinking of overloading `map` for keyboard and mouse.

Proposal:
`map-pointer mode modifiers button action|command`

alex-courtis:
Superceeded by #692

PR #692, opened 2022-09-13
map-pointer command

### LINKED ISSUE #685: Feature Request: map mouse button

alex-courtis:
resolves #685

superceeds #691

`map-pointer mode modifiers button action|command`

View under cursor is always focused when mapped button pressed.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -531,6 +531,10 @@ fn handlePointerMapping(self: *Self, event: *wlr.Pointer.event.Button, view: *Vi
                 .command => |args| {
                     self.seat.focus(view);
                     self.seat.runCommand(args);
+                    // This is mildly inefficient as running the command may have already
+                    // started a transaction. However we need to start one after the Seat.focus()
+                    // call in the case where it didn't.
+                    server.root.startTransaction();
                 },
             }
             break true;
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -528,6 +528,10 @@ fn handlePointerMapping(self: *Self, event: *wlr.Pointer.event.Button, view: *Vi
             switch (mapping.action) {
                 .move => if (!fullscreen) self.enterMode(.move, view),
                 .resize => if (!fullscreen) self.enterMode(.resize, view),
+                .command => |args| {
+                    self.seat.focus(view);
+                    self.seat.runCommand(args);
+                },
             }
             break true;
         }
--- a/river/PointerMapping.zig
+++ b/river/PointerMapping.zig
@@ -17,17 +17,12 @@
 const Self = @This();
 
 const std = @import("std");
+const assert = std.debug.assert;
 const wlr = @import("wlroots");
 
 const util = @import("util.zig");
 
-pub const ActionType = enum {
-    move,
-    resize,
-    command,
-};
-
-pub const Action = union(ActionType) {
+pub const Action = union(enum) {
     move: void,
     resize: void,
     command: []const [:0]const u8,
@@ -36,31 +31,32 @@ pub const Action = union(ActionType) {
 event_code: u32,
 modifiers: wlr.Keyboard.ModifierMask,
 action: Action,
-arena: std.heap.ArenaAllocator,
+/// Owns the memory backing the arguments if action is a command.
+arena_state: std.heap.ArenaAllocator.State,
 
 pub fn init(
     event_code: u32,
     modifiers: wlr.Keyboard.ModifierMask,
-    action_type: ActionType,
+    action_type: std.meta.Tag(Action),
     command_args: []const [:0]const u8,
 ) !Self {
-    var arena: std.heap.ArenaAllocator = std.heap.ArenaAllocator.init(util.gpa);
+    assert(action_type == .command or command_args.len == 1);
+
+    var arena = std.heap.ArenaAllocator.init(util.gpa);
     errdefer arena.deinit();
 
     const action: Action = switch (action_type) {
-        ActionType.move => Action.move,
-        ActionType.resize => Action.resize,
-        ActionType.command => blk: {
-            const allocator: std.mem.Allocator = arena.allocator();
+        .move => .move,
+        .resize => .resize,
+        .command => blk: {
+            const arena_allocator = arena.allocator();
 
-            var owned_args = try std.ArrayListUnmanaged([:0]const u8).initCapacity(allocator, command_args.len);
-
-            for (command_args) |arg| {
-                const owned = try allocator.dupeZ(u8, arg);
-                owned_args.appendAssumeCapacity(owned);
+            const owned_args = try arena_allocator.alloc([:0]const u8, command_args.len);
+            for (command_args) |arg, i| {
+                owned_args[i] = try arena_allocator.dupeZ(u8, arg);
             }
 
-            break :blk Action{ .command = owned_args.toOwnedSlice(allocator) };
+            break :blk .{ .command = owned_args };
         },
     };
 
@@ -68,10 +64,11 @@ pub fn init(
         .event_code = event_code,
         .modifiers = modifiers,
         .action = action,
-        .arena = arena,
+        .arena_state = arena.state,
     };
 }
 
 pub fn deinit(self: *Self) void {
-    self.arena.deinit();
+    self.arena_state.promote(util.gpa).deinit();
+    self.* = undefined;
 }
--- a/river/PointerMapping.zig
+++ b/river/PointerMapping.zig
@@ -14,13 +14,64 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+const Self = @This();
+
+const std = @import("std");
 const wlr = @import("wlroots");
 
-pub const Action = enum {
+const util = @import("util.zig");
+
+pub const ActionType = enum {
     move,
     resize,
+    command,
+};
+
+pub const Action = union(ActionType) {
+    move: void,
+    resize: void,
+    command: []const [:0]const u8,
 };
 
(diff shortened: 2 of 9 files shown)


river/Seat.zig
```diff
@@ -419,7 +419,7 @@ pub fn handleSwitchMapping(
     }
 }
 
-fn runCommand(self: *Self, args: []const [:0]const u8) void {
+pub fn runCommand(self: *Self, args: []const [:0]const u8) void {
```
alex-courtis, line 422: Exposed for convenience.

alex-courtis:
### Tested

```
riverctl map-pointer normal Super BTN_LEFT move-view

riverctl map-pointer normal Super BTN_RIGHT resize-view

riverctl map-pointer normal Super BTN_MIDDLE toggle-float

riverctl map-pointer normal Super+Control BTN_MIDDLE toggle-fullscreen

riverctl map-pointer normal Super+Control BTN_MIDDLE swap next

riverctl unmap-pointer normal Super+Control BTN_MIDDLE

riverctl unmap-pointer normal Super+Control BTN_MIDDLE

riverctl map-pointer normal Super+Control BTN_MIDDLE
error: not enough arguments

riverctl map-pointer normal Super+Control BTN_BLEH toggle-fullscreen
error: unknown button BTN_BLEH

riverctl map-pointer normal Super+Control BTN_MIDDLE
error: not enough arguments

riverctl map-pointer normal Super BTN_RIGHT resize-view aoeu
error: too many arguments
```


Fullscreen view move-view resize-view and commands.


Ensured button not sent to client on click


Multiple outputs actions and commands

river/command/map.zig
```diff
@@ -140,20 +139,18 @@ pub fn mapPointer(
         PointerMapping.Action.move
     else if (mem.eql(u8, args[4], "resize-view"))
```
alex-courtis, line 140: Parsing move/resize could be moved to `PointerMapping` `init`.

river/Cursor.zig
```diff
@@ -847,8 +849,6 @@ fn surfaceAtFilter(view: *View, filter_tags: u32) bool {
 pub fn enterMode(self: *Self, mode: enum { move, resize }, view: *View) void {
     log.debug("enter {s} cursor mode", .{@tagName(mode)});
 
-    self.seat.focus(view);
```
Leon-Plickat, line 850: Hmm... I can not come up with an example right now, but I have the feeling that this might cause some issues somewhere. Button mappings aren't the only thing that can change the cursor mode.

Right now I would say that it's better to keep this `focus()` call here and move the one in `handlePointerMapping()` into the `.command` branch of the switch.

river/PointerMapping.zig
```diff
@@ -14,13 +14,47 @@
...
+command_args: []const [:0]const u8,
+
+pub fn init(
+    event_code: u32,
+    modifiers: wlr.Keyboard.ModifierMask,
+    action: Action,
+    command_args: []const [:0]const u8,
+) !Self {
+    const args_len = if (action == Action.command) command_args.len else 0;
+    const owned_args = try util.gpa.alloc([:0]u8, args_len);
```
Leon-Plickat, line 41: Even for static length arrays, it's probably a good idea to use `ArrayListUnmanaged`, since it's API is a bit more handy.

Roughly like this:
```zig
var owned_args: std.ArrayListUnmanaged([:0]u8) = .{};
try owned_args.ensureTotalCapacity(util.gpa, command_args.len);
errdefer {
    for (owned_args.items) |arg| util.gpa.free(arg);
    owned.deinit(util.gpa);
}
for (command_args) |arg| {
    const owned = try util.gpa.dupeZ(arg);
    owned_args.appendAssumeCapacity(owned);
}
```

It might even make sense to put these allocations into an arena for easier cleanup, maybe (Just make sure to not grow the array list after the initial setup).

river/PointerMapping.zig
```diff
@@ -14,13 +14,47 @@
...
 action: Action,
+command_args: []const [:0]const u8,
+
+pub fn init(
+    event_code: u32,
+    modifiers: wlr.Keyboard.ModifierMask,
+    action: Action,
+    command_args: []const [:0]const u8,
+) !Self {
+    const args_len = if (action == Action.command) command_args.len else 0;
```
Leon-Plickat, line 40: I'd prefer a cleaner separation of the command and non-command cases.

river/PointerMapping.zig
```diff
@@ -14,13 +14,47 @@
...
 pub const Action = enum {
     move,
     resize,
+    command,
 };
 
 event_code: u32,
 modifiers: wlr.Keyboard.ModifierMask,
 action: Action,
+command_args: []const [:0]const u8,
```
Leon-Plickat, line 32: You could turn `action` into a tagged union (using `Action` as tag-type). Then `command_args` could be the `command` case, with the two others being `void`.

Leon-Plickat (commented):
Took me some time to get to this, but here is a preliminary review. Just a few minor things.

alex-courtis:
> Took me some time to get to this, but here is a preliminary review. Just a few minor things.

No worries at all.

Thanks for the zig tips... I'll get to this one early next week.

river/Cursor.zig
alex-courtis: Safe. Done.

river/PointerMapping.zig
alex-courtis: Refactored as switch.

river/PointerMapping.zig
alex-courtis: That's a nice construct. Done.

river/PointerMapping.zig
```diff
@@ -14,13 +14,72 @@
...
+            errdefer {
+                for (owned_args.items) |arg| util.gpa.free(arg);
+                owned_args.deinit(util.gpa);
+            }
+            for (command_args) |arg| {
+                const owned = try util.gpa.dupeZ(u8, arg);
+                owned_args.appendAssumeCapacity(owned);
+            }
+
+            const action: Action = Action{ .command = owned_args.toOwnedSlice(util.gpa) };
```
alex-courtis, line 62: I think I have misunderstood your intent here. Did you mean to change the action/command type to an `ArrayList` or use the slice?

river/PointerMapping.zig
```diff
@@ -14,13 +14,72 @@
...
+
+            owned_args.deinit(util.gpa);
```
alex-courtis, line 64: I do not think this is correct. It seems that a double free may occur from the `errdefer` block.

river/PointerMapping.zig
alex-courtis: I'm struggling with the use of an arena. Attempt:

```zig
        ActionType.command => blk: {
            var arena = std.heap.ArenaAllocator.init(util.gpa);
            defer arena.deinit();

            var owned_args = try std.ArrayListUnmanaged([:0]const u8).initCapacity(arena.allocator(), command_args.len);

            for (command_args) |arg| {
                const owned = try util.gpa.dupeZ(u8, arg);
                owned_args.appendAssumeCapacity(owned);
            }

            break :blk Action{ .command = owned_args.toOwnedSlice(util.gpa) };
        },
```

Even though the strings and the slice are allocated using gpa the arena always frees everything.

alex-courtis:
This has been a fantastic learning experience.

Thank you for all your help and patience.

river/PointerMapping.zig
Leon-Plickat: The arena would be part of the `Action` struct, holding the strings as well as the slice. Then in `Action.deinit()` just call `Arena.deinit()` for cleanup.

river/PointerMapping.zig
Leon-Plickat: No, I think you got it. This is exactly what I recommended: Using an ArrayList instead of doing the slice allocation and population manually.

river/PointerMapping.zig
alex-courtis: That make a lot of sense, thank you.

I'll get to that next week.

river/PointerMapping.zig
alex-courtis: Done.

river/PointerMapping.zig
alex-courtis: Added the `ArenaAllocator` to `PointerMapping`.

This is a really clean pattern. It can applied to `Mapping` and friends.

alex-courtis:
This works. Repeated tests and dogfooding now.

I could not validate the freeing via valgrind. There's a lot of noise from wl_roots which I can [deal with](https://github.com/alex-courtis/way-displays/blob/3d2b4fab16d4c43eada42723fbe61670d096f259/way-displays.supp). Is there a zig tool?

When, say, commenting out the arena.deinit, I did not see anything referencing PointerMapping.

ifreund (approved):
Thanks for your contribution and sorry for the long wait. I haven't had as much time for river recently as I would like.

I've squashed your commits and added a few fixes/style tweaks in a separate commit myself rather than bothering you with another round of review and more delay.

Assuming the CI is green I'll merge this in a few minutes :)

alex-courtis:
> I've squashed your commits and added a few fixes/style tweaks in a separate commit myself rather than bothering you with another round of review and more delay.

Thank you. Much more efficient and my preferred workflow.

Issue #472, opened 2021-11-20
Give windows tags as they spawn


Magnitudexx:
I could not find a way to give specific applications certain tags when they spawn automatically? Is there a feature like that, is it not implemented yet or is it not intended to work like that?

Leon-Plickat:
This is currently not possible.

How and when this will be possible depend on which direction #447 and/or #357 take.

Magnitudexx:
Ok i will wait and see, river seems like a really interesting compositor but until there is something like this i personally cant use it as a daily driver. Looking forward to the fututre!

ifreund:
implemented in https://github.com/riverwm/river/pull/852

Issue #201, opened 2021-01-14
Spawning programs on a different tag


FoundOne:
I prefer to put my instant messengers/chats in a certain tag, which is not and just having the notifications form them when I'm doing other stuff. Every time I start river I have to spawn them manually, which gets tedious over time. Is there a way to automate that process?

ifreund:
Not yet, but it is planned to add a `-tags` argument to spawn that allows setting the tags to spawn the program with. This will work by keeping track of the pid of the spawned process and checking against the pid of new wayland clients that map a toplevel.

ifreund:
With xdg-activation there is now a better, non pid based way to implement this. See this wlroots PR https://github.com/swaywm/wlroots/pull/3305

firstrow:
https://github.com/riverwm/river/commit/0b142bd16baab2dba9e01444892032d05c179c2a - tag rule.

ifreund:
The tag rule is a different feature than the one discussed here, but it certainly makes the complication of implementing this feature request less attractive.

PR #696, opened 2022-09-17
river: predefine tags for app-id and title

choelzl:
Related issues: #201 #472 

This PR adds a command to `riverctl`. The objective being to pre-define tags for app-id's and titles (similarly to how it is done for floating).

Usage example: `riverctl spawn-tags-filter app-id "kitty" 4`

### DIFF
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -179,18 +179,19 @@ pub fn getSpawnTagsFilter(self: Self, view: *View) u32 {
     const tags_curr = view.output.current.tags;
     const tags_mask = view.output.spawn_tagmask;
     const tags = tags_curr & tags_mask;
-    
+
     if (view.getAppId()) |app_id| {
-        if(self.spawn_tags_app_ids.contains(std.mem.span(app_id))) {
+        if (self.spawn_tags_app_ids.contains(std.mem.span(app_id))) {
             return self.spawn_tags_app_ids.get(std.mem.span(app_id)) orelse tags;
         }
     }
 
     if (view.getTitle()) |title| {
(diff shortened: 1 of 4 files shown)


river/Config.zig
```diff
@@ -171,6 +175,30 @@ pub fn shouldFloat(self: Self, view: *View) bool {
     return false;
 }
 
+pub fn getSpawnTagsFilter(self: Self, view: *View) u32 {
```
Leon-Plickat, line 178:
```suggestion
pub fn getSpawnTagsFilter(self: Self, view: *View) ?u32 {
```

Return tags if found, `null` if not.

river/View.zig
```diff
@@ -494,6 +494,9 @@ pub fn map(self: *Self) !void {
         if (self.getTitle()) |s| handle.setTitle(s);
         if (self.getAppId()) |s| handle.setAppId(s);
 
+        self.pending.tags = server.config.getSpawnTagsFilter(self);
```
Leon-Plickat, line 497:
```suggestion
        if (server.config.getSpawnTagsFilter(self)) |tags| self.pending.tags = tags;
```

Initial tags are set in `.init()`

sfrique:
@choelzl can you updated it?

@Leon-Plickat besides this changes, what else need to be done? 

I am really interested in this feature

HenkVanMaanen:
Interested too in this feature!

ifreund:
This patch has grown significant conflicts due to the introduction of the rules system.

Now that the rules system has been fleshed out I would be willing to accept a patch adding an action type to set initial tags.

PR #708, opened 2022-10-16
add command to set keybaord layout

Leon-Plickat:


TAforever:
This is a very good idea. Could you create a waybar module in the future to display the keyboard layout?

pmkap:
Nice. Works well for me too. I think it would be nice if trailing `_` could be left out. Lots of people probably only use the first argument.

Also, at this point, couldn't we make this keyboard specific very easily?





> This is a very good idea. Could you create a waybar module in the future to display the keyboard layout?

No. It would need to be added to a protocol like river-status (and implemented).

Leon-Plickat:
> Could you create a waybar module in the future to display the keyboard layout?

"You" as in *me* specifically? No. I don't use waybar and I generally am not interested in patching projects I don't use myself.

"You" in a general, "some random person" sense? Yes. I don't think we even need a custom protocol for that: Waybar should be able to just bind the wl_keyboard.

Leon-Plickat:
> Also, at this point, couldn't we make this keyboard specific very easily?

No. There is nothing easy - or even sane - about per-keyboard layouts. It's basically a hack: It requires the compositor switching the layout live based on the last pressed key. It's overly complicated for very little return.

Unless I can see a real world, non-hypothetical use case that applies to more than just a single person, I see no reason to spend any time on that.

TAforever:
Yes, by "you" I mean the developers of the river.
I like the river and I would like the river to develop and get better. Of the wayland compositors, we actually only have sway for now, but how can  use manual tiling :) The fact is that waybar is the best bar for TWC and already supports tag and window title modules and I don’t think that these modules were made by the waybar developer. I also think that users need other layouts than English, and setting them using an environment variable is not the best solution.

Leon-Plickat:
Did anyone who tested this also experience the micro-stutter when loading a new layout? Wondering whether this is just a quirk of my system. I am not sure if anything can be done about it, since it seems to happen in xkbcommon.

---

> That's not only about this PR since this the way commands works but finding wich value is wrong with an InvalidValue error is a bit annoying with a command with so many args in it, would be nicer if the error would print the wrong value.

I don't think xkbcommon is nice enough to tell us what wrong. Either way, I think proper error messages make sense for riverctl. Before we do that though, we probably want to change our protocol to send explicit requests instead of a string parsed by river, so that we can do all the verification and error messages client side.

---

>  I think it would be nice if trailing _ could be left out. Lots of people probably only use the first argument.

I generally prefer if things aren't *too* magic. Maybe we can use a configuration syntax that makes this more explicit, like maybe `layout=de options=caps:swapescape variant=nodeadkeys`. But that would be a pretty different to how other options work. Either way, that's bikeshedding and can always be done later.

---

> Yes, by "you" I mean the developers of the river.

Well, I don't feel responsible for the general ecosystem around (our) Wayland compositor(s). The river modules available in waybar have been implemented by people who explicitly wanted to have them. It will be the same for a hypothetical keyboard layout module.

TAforever:
I looked at the list of those who created the waybar modules and so
the river/tags module was created by Isaac Freund, and the hyprland modules were created by Vaxry, the main developer of hyprland. Therefore, if my request was excessive or impudent, then I ask your forgiveness.

novakne:
> Did anyone who tested this also experience the micro-stutter when loading a new layout?

Haven't experience something like this

MaxVerevkin:
> I don't think we even need a custom protocol for that: Waybar should be able to just bind the wl_keyboard.

Are you sure that this is possible? I mean can a wayland client determine the active layout without receiving keyboard input? (I haven't tried this PR, but I don't think it is possible right now).

doc/riverctl.1.scd
```diff
@@ -330,6 +330,10 @@ A complete list may be found in _/usr/include/linux/input-event-codes.h_
 *list-input-configs*
 	List all input configurations.
 
+*keyboard-layout* _layout_ _variant_ _model_ _options_ _rules_
+	Set the XKB layout for all keyboards. Use \_ for values you do not
+	want to set.
+
```
ifreund, line 336: This feature looks good, but I'm not a huge fan of this CLI interface. It seems to me that most users will only want to specify only the layout. A few more will want to specify options, but I suspect very few users will set the model or rules for example.

Consider also the readability of 
```
riverctl keyboard-layout de _ foo caps:swapesc _
```
vs
```
riverctl keyboard-layout -options caps:swapesc -model foo de
```

ifreund (commented):
Hey, thanks for the patch! 

The implementation looks nice and clean, my only gripe is with the CLI design.

doc/riverctl.1.scd
Leon-Plickat: I was a bit hesitant to do something like this initially, considering command parsing is done server side and river feels like the wrong place for making the riverctl UX nice.

Anyway, I should have this done sometime this week.

ifreund:
To avoid any confusion, this PR was closed by accident and github has thwarted our attempts to re-open it.

PR #712, opened 2022-10-26
Add Q&A for how this layout is used

ideasman42:
It took me a while to figure out how to activate this layout, adding a note to the Q&A.

Also correct typo.


ifreund (approved):
Thanks!

PR #730, opened 2022-12-02
Global spawn-tagmask

tleydxdy:
Not quite complete yet, but wanted to post it so people can comment about their use-case and see what makes the most sense.

Currently per-output spawn-tagmask have a few problems. Setting it for an particular output is difficult since it involves focus. The tagmask will be lost if the output is gone (it went to sleep, etc).

Having just an global tagmask would solve these problems, it should not affect any use-cases afaik, but it certainly might break some people's config/use-case that I'm not aware of.
Having both global and per-output tagmask won't break existing setups but the problems will still exist and can be confusing for users to have two tagmasks.

### DIFF
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -85,6 +85,9 @@ warp_cursor: WarpCursorMode = .disabled,
 /// Output.layout_namespace is null.
 default_layout_namespace: []const u8 = &[0]u8{},
 
+/// Bitmask that whitelists tags for newly spawned views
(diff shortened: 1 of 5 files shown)


ifreund:
Hey, thanks for the patch. I decided to make this command apply globally, I doubt anyone will prefer the old behavior in practice. Just pushed 39104ae9e3db55e5c3e8428bfdad67153b58737d to master.

PR #746, opened 2022-12-29
river: fix resize command

Leon-Plickat:
In 489a49735 the view.move() call, which is used to keep the view centered after a resize, was accidentally removed.


ifreund (approved):
LGTM, thanks for the fix!

PR #747, opened 2022-12-30
touch: support drag and drop

### LINKED ISSUE #634: Implement touch drags
River now has basic touch support as of #632, but touch drags are not yet implemented.

I took an incomplete stab at this here but it's pretty tricky and I don't have hardware to test on: https://github.com/riverwm/river/tree/touch-drags

Related: https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/3452

ifreund:
Closes #634

Any help testing this PR by people with touchscreens would be much appreciated and is require before this can be merged as I don't have a test system with a touch screen currently.

To test, the `weston-dnd` client might be useful. Alternatively, dragging firefox tabs by touch should work now in theory and show a drag icon just as dragging them with the pointer currently does.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -85,6 +85,11 @@ const Image = enum {
 
 const default_size = 24;
 
+const LayoutPoint = struct {
+    lx: f64,
+    ly: f64,
+};
+
 const log = std.log.scoped(.cursor);
 
 /// Current cursor mode as well as any state needed to implement that mode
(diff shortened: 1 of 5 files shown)


ifreund:
I've gotten confirmation that this works as expected for weston-dnd on IRC for single touch point drags at least. It turns out that weston-dnd [ignores all touch points unless they have id 0](https://gitlab.freedesktop.org/wayland/weston/-/blob/5b7561b25d7db05839def4d69e2ca1f3a1e18e56/clients/dnd.c#L649) though.

I can't seem to find a better test client and am not planning on writing one myself without touchscreen hardware.

This PR seems to not break anything and be at least as correct as weston and sway are if not more so when it comes to multitouch where we handle drags similarly to wayfire. Lets merge it and see if anyone complains :)

PR #750, opened 2022-12-31
Rework focus-follows-cursor to work with warp

tleydxdy:
When focus-follows-cursor is used with cursor-warp, some windows will get focus before the cursor properly "enters" the window since they have a larger input-region than their window geometry, this causes the cursor to be yanked to the middle unexpectedly.

This fix makes it so the focus is only given when the cursor enters the window geometry.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -112,6 +112,8 @@ hide_cursor_timer: *wl.EventSource,
 hidden: bool = false,
 may_need_warp: bool = false,
 
+last_focus_follows_cursor_target: ?*View = null,
+
 /// Keeps track of the last known location of all touch points in layout coordinates.
 /// This information is necessary for proper touch dnd support if there are multiple touch points.
 touch_points: std.AutoHashMapUnmanaged(i32, LayoutPoint) = .{},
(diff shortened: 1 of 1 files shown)


river/Cursor.zig
```diff
@@ -1049,21 +1050,33 @@ pub fn checkFocusFollowsCursor(self: *Self) void {
...
-                    }
-                },
-                .layer_surface, .lock_surface => {},
-                .xwayland_override_redirect => assert(build_options.xwayland),
-            }
+        switch (result.parent) {
+            .view => |view| {
+                // Don't re-focus the last focused view when the mode is .normal
+                if (server.config.focus_follows_cursor == .normal and
+                    view == self.last_followed_view) return;
```
ifreund, line 1057: Could you explain why we need this `last_followed_view` field instead of the current `self.seat.wlr_seat.pointer_state.focused_surface != result.surface` check?

river/Cursor.zig
```diff
@@ -1049,21 +1050,33 @@ pub fn checkFocusFollowsCursor(self: *Self) void {
...
+                // Some windows have a input region bigger than their window
+                // geometry, we only want to move focus when the cursor
+                // properly enters the window (the box that we draw borders around)
+                var output_layout_box: wlr.Box = undefined;
+                server.root.output_layout.getBox(view.output.wlr_output, &output_layout_box);
+                const cursor_lx = self.wlr_cursor.x - @intToFloat(f64, output_layout_box.x);
```
ifreund, line 1063: The result of this is not in layout coordinates but in output-relative coordinates.
```suggestion
                const cursor_ox = self.wlr_cursor.x - @intToFloat(f64, output_layout_box.x);
```

river/Cursor.zig
tleydxdy: because the focused_surface is set when the cursor is over the input_region, which we don't want

ifreund (approved):
LGTM, thanks!

Just force pushed to rebase and tweak the variable name to last_focus_follows_cursor_target.

PR #751, opened 2022-12-31
river_control: mark callback events as destructors

MaxVerevkin:
I suppose these kind of changes require a version bump, but since callback has no requests, this change should not break anything. Please let me know if my assumption is wrong.

### DIFF
--- a/protocol/river-control-unstable-v1.xml
+++ b/protocol/river-control-unstable-v1.xml
@@ -63,7 +63,7 @@
       by the compositor after one of the events is sent.
     </description>
 
-    <event name="success">
+    <event name="success" type="destructor">
       <description summary="command successful">
         Sent when the command has been successfully received and executed by
         the compositor. Some commands may produce output, in which case the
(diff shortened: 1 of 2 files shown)


river/Control.zig
```diff
@@ -117,6 +117,7 @@ fn handleRequest(control: *zriver.ControlV1, request: zriver.ControlV1.Request,
                 };
                 defer if (err == command.Error.Other) util.gpa.free(failure_message);
                 callback.sendFailure(failure_message);
+                callback.destroy();
```
ifreund, line 120: Hmm, I think I'd rather have zig-wayland generate a `destroySendFailure()` that both sends the failure event and destroys the callback so that we can't forget to call destroy and potentially leak memory on the server side.

protocol/river-control-unstable-v1.xml
```diff
@@ -63,7 +63,7 @@
       by the compositor after one of the events is sent.
     </description>
 
-    <event name="success">
+    <event name="success" type="destructor">
```
ifreund, line 66: This change is fine to make without a version bump, the only reason this annotation wasn't here from the start is that it wasn't supported for events yet when I wrote this protocol: https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/82.

ifreund (commented):
Thanks for the patch, I'll just make that zig-wayland change I mentioned before merging it.

ifreund (approved):
Thanks for the patch!

PR #752, opened 2023-01-01
Update riverctl.1.scd

Lemanr-ops:
The proposed note I think is very much needed. I tried changing the keyboard layout using so many methods but .bash_profile was the only one that had actually changed the layout. for more info, I have the xwayland flag on during river install. This Arch install is very recent and the hardware is from the thinkpad E15 12th gen. Just encase any of that info useful. Feel free to change the wording, I just want .bash_profile mentioned as an alternative for others who have found changing keyboard layout to be seemingly impossible.


ifreund (commented):
The `XKB_DEFAULT_*` environment variables are no longer the recommended way to configure the keyboard layout. The `riverctl keyboard-layout` command should be used instead.

Lemanr-ops:
Thanks for the quick reply

Issue #71, opened 2020-07-24
Make input device properties configurable


dnkl:
Wasn't really sure how to phrase the title, but basically, make it possible to configure things like left/right-hand mouse, trackpad settings such as tap-to-click, drag and drag-lock etc.

Pretty sure this is on someones mental TODO list, but couldn't find a tracking issue. Adding one since I'm very interested in this.

ifreund:
This is certainly something we need eventually, though other features I personally find more important have taken priority so far.

Leon-Plickat:
It actually is on my mental TODO list.

I personally think the best way to approach this is with an external program to configure inputs (like kanshi does for outputs) that uses a wayland extension to talk to the compositor. If such a protocol extension gains traction, there would be a compositor agnostic way to configure input devices, which I think would be quite nice.

ifreund:
An input equivalent to `wlr-output-management` sounds desirable to me, and implementing such a protocol instead of river-specific configuration options would be preferable.

The steps forward for this would be:
- Writing such a protocol and submitting it to `wayland-protocols` for review
- Implementing this protocol in wlroots
- Implementing whatever can't reasonably be done in wlroots in river

ifreund:
This feature is also the only major blocker to supporting multiple seats as most (all?) of river's code already behaves as if there are multiple seats. The only real missing piece for multi-seat support is having a way to assign different input devices to different seats, which is part of input configuration and should probably be covered by the proposed protocol as well.

Leon-Plickat:
As I see no chance of such a protocol happening in the nearish future, I decided to experiment with just implementing basic input configuration in river. See #303, although it does not work yet.

ifreund:
#303 has been merged and river now supports basic input configuration. This issue however will stay open to track design/implementation of the proposed protocol described above.

PR #753, opened 2023-01-01
river-status: add keyboard_layout event

MaxVerevkin:
I'm new to zig and server-side wayland, so please let me know if if there are better ways to do certain things.

`sway` was used as a reference.

- Simple test client: https://github.com/MaxVerevkin/river-kbd-layout-watcher
- Status bar integration: https://github.com/MaxVerevkin/i3status-rust/tree/river-kbd

### DIFF
--- a/river/SeatStatus.zig
+++ b/river/SeatStatus.zig
@@ -27,6 +27,8 @@ const util = @import("util.zig");
 const Seat = @import("Seat.zig");
 const Output = @import("Output.zig");
 const View = @import("View.zig");
+const Keyboard = @import("Keyboard.zig");
+const InputDevice = @import("InputDevice.zig");
 
 seat: *Seat,
 seat_status_v1: *zriver.SeatStatusV1,
@@ -40,6 +42,13 @@ pub fn init(seat_status: *SeatStatus, seat: *Seat, seat_status_v1: *zriver.SeatS
     seat_status.sendMode(server.config.modes.items[seat.mode_id].name);
     if (seat.focused_output) |output| seat_status.sendOutput(output, .focused);
     seat_status.sendFocusedView();
+
+    var it = server.input_manager.devices.iterator(.forward);
+    while (it.next()) |device| {
+        if (device.wlr_device.type != .keyboard) continue;
+        const wlr_keyboard = device.wlr_device.toKeyboard();
+        seat_status.sendKeyboardLayout(device, Keyboard.getActiveLayoutName(wlr_keyboard));
+    }
 }
 
 fn handleRequest(seat_status_v1: *zriver.SeatStatusV1, request: zriver.SeatStatusV1.Request, _: *SeatStatus) void {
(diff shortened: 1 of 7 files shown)


MaxVerevkin:
Test client: https://github.com/MaxVerevkin/rust-river-kbd-layout-watcher

ifreund (commented):
I haven't looked into this in great depth yet, but I don't see how this handles two keyboards on the same seat with different active layouts...

MaxVerevkin:
Hmm, I guess the event should also contain some kind of input identifier?

river/SeatStatus.zig
```diff
@@ -78,3 +87,22 @@ pub fn sendMode(self: Self, mode: [*:0]const u8) void {
...
+    if (self.seat_status.getVersion() < 5) return;
+
+    const identifier = std.fmt.allocPrint(
+        util.gpa,
+        "{s}\x00",
+        .{device.identifier},
+    ) catch return;
+    defer util.gpa.free(identifier);
+
+    const c_ident = @ptrCast([*:0]const u8, identifier.ptr);
```
MaxVerevkin, line 101: Should I change `InputDevice.identifier` to be a null terminated sting to avoid this?

MaxVerevkin:
Should this event be sent when keyboard is created / destroyed?

MaxVerevkin:
An alternative to this can be a separate protocol extension that exposes input information similar to how wl_output/xdg_output describes outputs. Are there any such extensions already (I couldn't find any)? Any opinions on this?

river/SeatStatus.zig
ifreund: Sure, that would be more idiomatic. Also note that allocPrintZ() can be used instead of this pointer cast.

Leon-Plickat:
> Edit: I suppose this is covered by https://github.com/riverwm/river/issues/71

I don't think so. My idea was to limit that protocol to devices handled by libinput, because unifying multiple API surfaces into a single protocol seems a bit unreasonable.

MaxVerevkin:
Right now the event is sent when on binding, on keyboard creation and when layout changes. I think it's better so send it when keyboard is destroyed too. Should I add `keyboard_destroyed` event or just send `keyboard_layout_clear`?

ifreund:
I don't think I'm going to accept this feature as part of a river-specific protocol. Other compositors have the exact same issue and it would be better for this to be solved in a compositor-agnostic way if possible.

See for example https://github.com/labwc/labwc/issues/1317

Alternatively, with my plans for a complete window management protocol this information may end up being exposed to the window manager process by river which could then relay it to status bars and the like.

Sorry to keep you hanging on this for so long without a clear decision one way or the other.

MaxVerevkin:
> Other compositors have the exact same issue and it would be better for this to be solved in a compositor-agnostic way if possible.

I absolutely agree.

In the mean time, for anyone interested, I will continue maintaining this branch, until a better solution is designed.

karampok:
Hi, I am interested, as I do miss a such feature (just be able to print in the bar the current layout).
Do you know if there there anyother  way (standalone tool) that I can query the layout in wayland/river? 
xdb-switch status or setxkbmap -query or localectl do not work. I only have seen sway-lock to be able to do that.

Thanks!

PR #756, opened 2023-01-02
river_seat_status 3 -> 4

MaxVerevkin:
Perhaps a check could be added to zig-wayland scanner to prevent this kind of errors?

### DIFF
--- a/protocol/river-status-unstable-v1.xml
+++ b/protocol/river-status-unstable-v1.xml
@@ -100,7 +100,7 @@
     </event>
   </interface>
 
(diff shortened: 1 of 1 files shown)


MaxVerevkin:
Or.. is is this actually correct? I thought that when an object created by another object, it inherits the version of parent object. But this doesn't quite work for wl_surface.frame (wl_surface has version 5, wl_callback has version 1).

ifreund:
Yeah, this isn't an error and youll see this often in upstream protocols. It would only be an error if river_seat_status had a version 4 event/request.

PR #759, opened 2023-01-05
river: fix bug in snap down / right

lx7:
Found a bug in snap down / snap right. The calculation of view.pending.box is inverted:
- snap right should use output_width (not output_height)
- snap down should use output_height (not output_width)


Leon-Plickat (approved):

ifreund (approved):
Thanks for the fix!

PR #761, opened 2023-01-06
fix border color by premultiplying alpha

### LINKED ISSUE #760: Border alpha doesn't work as expected
According to riverctl manpage, color format can be:
0xRRGGBBAA

This border is represented as ```0xff79c6ff```

image

This one is: ```0xff79c600```

image

The interesting thing is, first color has 255 alpha and you can see the color fully. It's expected, but in the second one, the alpha is 0, I shouldn't be seeing any color. It should be fully transparent.

Am I missing something or is this a bug?

MaxVerevkin:
Closes #760

### DIFF
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -46,19 +46,19 @@ pub const HideCursorWhenTypingMode = enum {
     enabled,
 };
 
-/// Color of background in RGBA (alpha should only affect nested sessions)
+/// Color of background in RGBA with premultiplied alpha (alpha should only affect nested sessions)
 background_color: [4]f32 = [_]f32{ 0.0, 0.16862745, 0.21176471, 1.0 }, // Solarized base03
 
 /// Width of borders in pixels
 border_width: u31 = 2,
 
-/// Color of border of focused window in RGBA
+/// Color of border of focused window in RGBA with premultiplied alpha
 border_color_focused: [4]f32 = [_]f32{ 0.57647059, 0.63137255, 0.63137255, 1.0 }, // Solarized base1
 
-/// Color of border of unfocused window in RGBA
+/// Color of border of unfocused window in RGBA with premultiplied alpha
 border_color_unfocused: [4]f32 = [_]f32{ 0.34509804, 0.43137255, 0.45882353, 1.0 }, // Solarized base01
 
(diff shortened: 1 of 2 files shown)


novakne:
It seems to fixed it, still curious to understand  why we had a difference between light and dark background though

MaxVerevkin:
@novakne, I haven't looked into the actual implementation. but I _guess_ that color blending (for example A + B, A on top of B) happens as follows: `result[i] = A[i]*A[alpha] + B[i]*(1-A[alpha])`.

Whichever part of the graphics stack does it, it assumes that `A[i]*A[alpha]` is already done before blending. So it becomes `result[i] = A[i] + B[i]*(1-A[alpha])`.

Then for white background we have `result[i] = A[i] + 1 - A[alpha]`. If alpha is zero, result >= 1, i.e. white -- same as background.

For black background we have `result[i] = A[i]`, i.e. the color itself.

Leon-Plickat:
This indeed makes alpha behave as expected, however I am still not totally convinced this is actually our bug.

We use wlroots `wlr_render_rect()` function to draw borders, whichs documentation does not say anything about pre-multiplying alpha.

However I also found [this issue](https://gitlab.freedesktop.org/wlroots/wlroots/-/issues/1703), which has this comment:

image

May or may not be related.

---

Also, if this is our bug, I wonder if alpha should instead be multiplied in `renderRect()` in `river/render.zig` and not in the parser.

ifreund (commented):
Hey, thanks for the investigation and patch!

As far as I understand things, pretty much everything in the compositing stack assumes premultiplied alpha by default. See for example this commit and the linked ML thread: https://gitlab.freedesktop.org/wayland/wayland/-/commit/c6c1bbe4abf9e5ce159442be22c1278d39844c08

I'm just waiting to get confirmation from people more familiar with wlroots rendering code than I am before I merge this. I'll probably send a patch clarifying their documentation as well.

As for where the best place to implement this is in river, I think doing it here in parsing is fine though it would probably be good to update the doc comments in Config.zig to mention that alpha is premultiplied.

ifreund (approved):
Just rebased and force pushed to improve comments in Config.zig and expand on the commit message.

Thanks for the patch!

PR #762, opened 2023-01-06
InputConfig: Implement disable while trackpointing

novakne:
New option in libinput https://wayland.freedesktop.org/libinput/doc/latest/configuration.html#disable-while-trackpointing

I don't have the hardware to test it though so if someone with a trackpoint can try it that would be nice.

### DIFF
--- a/river/InputConfig.zig
+++ b/river/InputConfig.zig
@@ -137,6 +137,23 @@ pub const DwtState = enum {
     }
 };
 
+pub const DwtpState = enum {
+    disabled,
+    enabled,
+
+    pub fn apply(dwtp_state: DwtpState, device: *c.libinput_device) void {
+        const want = @as(c_uint, switch (dwtp_state) {
+            .disabled => c.LIBINPUT_CONFIG_DWTP_DISABLED,
+            .enabled => c.LIBINPUT_CONFIG_DWTP_ENABLED,
+        });
+        if (c.libinput_device_config_dwtp_is_available(device) == 0) return;
+        const current = c.libinput_device_config_dwtp_get_enabled(device);
+        if (want != current) {
+            _ = c.libinput_device_config_dwtp_set_enabled(device, want);
+        }
+    }
+};
+
 pub const MiddleEmulation = enum {
     disabled,
     enabled,
@@ -271,6 +288,7 @@ events: ?EventState = null,
 drag: ?DragState = null,
 @"drag-lock": ?DragLock = null,
 @"disable-while-typing": ?DwtState = null,
+@"disable-while-trackpointing": ?DwtpState = null,
 @"middle-emulation": ?MiddleEmulation = null,
 @"natural-scroll": ?NaturalScroll = null,
 @"left-handed": ?LeftHanded = null,
--- a/completions/bash/riverctl
+++ b/completions/bash/riverctl
@@ -88,6 +88,7 @@ function __riverctl_completion ()
 				drag \
 				drag-lock \
 				disable-while-typing \
+				disable-while-trackpointing \
 				middle-emulation \
 				natural-scroll \
 				left-handed \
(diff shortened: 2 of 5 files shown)


adamcstephens:
I tested this on my X1 Carbon Gen 7 and I can see no difference between enabled and disabled. In both cases the trackpoint has priority and touchpad movement is ignored while the trackpoint has pressure. 

According to the libinput docs it is enabled by default, so I would expect disabled to allow both to fight for control. As far as I can tell this does not happen and the trackpoint always "wins".

novakne:
Hmm that's weird, I can't see why the code wouldn't works, unless I forgot to make a change somewhere. Does `riverctl list-input-configs` list your input with the right config?

novakne:
Also just thought about that but I assumed this was a config option to use with the touchpad input name but maybe it's with the trackpoint name, which one did you tried?

Leon-Plickat:
Note that is sometimes is also a BIOS setting.

adamcstephens:
> Also just thought about that but I assumed this was a config option to use with the touchpad input name but maybe it's with the trackpoint name, which one did you tried?

I tried this on the touchpad. Given that the other options on the libinput page are for the touchpad, this seems logical to me.

```
─❯ riverctl input pointer-1739-52619-SYNA8004:00_06CB:CD8B_Touchpad disable-while-trackpointing enabled

─❯ riverctl list-input-configs
pointer-1739-52619-SYNA8004:00_06CB:CD8B_Touchpad
	events: enabled
	disable-while-trackpointing: enabled
	natural-scroll: enabled
	tap-button-map: left-right-middle

pointer-1739:52619:SYNA8004:00_06CB:CD8B_Touchpad
	middle-emulation: disabled
```

I'm not sure why this returns two separate entries, but all the settings apply to the sole touchpad.

> Note that is sometimes is also a BIOS setting.

This machine only lets me enable/disable the touchpad and trackpad in the BIOS, so nothing about priority.

Ordoviz:
On my ThinkPad X1 Yoga, running `riverctl input pointer-2-7-SynPS/2_Synaptics_TouchPad disable-while-trackpointing disabled` does turn off DWTP as expected.

novakne:
So I tried the function with more logging info and it appears that `c.libinput_device_config_dwtp_is_available(device)` return `1` and `c.libinput_device_config_dwtp_set_enabled(device, want)` return `c.LIBINPUT_CONFIG_STATUS_SUCCESS` on my system even if I don't have a trackpoint.

So my guess would be that this PR works as expected and for some reasons unrelated to river @adamcstephens trackpoint doesn't supports this config options.

novakne:
Resolved conflicts

novakne:
Resolved conflicts one last time before 0.3.0 hopefully, if not feel free to ping me before reviewing/merging

river/InputConfig.zig
```diff
@@ -137,6 +137,23 @@ pub const DwtState = enum {
...
+    pub fn apply(dwtp_state: DwtpState, device: *c.libinput_device) void {
+        const want = @as(c_uint, switch (dwtp_state) {
+            .disabled => c.LIBINPUT_CONFIG_DWTP_DISABLED,
+            .enabled => c.LIBINPUT_CONFIG_DWTP_ENABLED,
+        });
+        if (c.libinput_device_config_dwtp_is_available(device) == 0) return;
+        const current = c.libinput_device_config_dwtp_get_enabled(device);
+        if (want != current) {
+            _ = c.libinput_device_config_dwtp_set_enabled(device, want);
+        }
```
ifreund, line 153: From skimming the libinput source code, it doesn't seem necessary to call any libinput functions other than `libinput_device_config_dwtp_set_enabled()` here. If it's not supported or there is no state change libinput will do nothing.

ifreund: Hmm, looks like we do a lot of redundant libinput getter calls in this file...

ifreund (approved):
LGTM, I'll clean up the redundant code in this file in a separate commit tomorrow.

river/InputConfig.zig
Leon-Plickat: IIRC there was a reason why I checked the current state of all libinput settings before changing them, but i can't remember it.

PR #763, opened 2023-01-06
ci: Use meson setup command

novakne:
Using `meson build` without `setup` have been deprecated, fix warning message:
  'WARNING: Running the setup command as `meson [options]` instead of
  `meson setup [options]` is ambiguous and deprecated.


ifreund (approved):
LGTM, thanks!

PR #767, opened 2023-01-07
borders: fix visual glitch with fractional scaling

### LINKED ISSUE #496: Visual glitch on border's bottom left corner, with a scaling factor of 1.75
I was testing how foot behaved under various scaling factors, and noticed what looks like a mistake in river's rendering of the SSD's bottom left corner.

Note how the left border continues down below the bottom border:
river-bottom-right-border-scale-175

I didn't notice any issues with 1.5 or 2x scaling, only at 1.75x.

Version: 0.2.0-dev-d93ee2c - built december 11:th (yup, getting a bit outdated...)

MaxVerevkin:
Fixes #496

### DIFF
--- a/river/render.zig
+++ b/river/render.zig
@@ -381,39 +381,56 @@ fn renderBorders(output: *const Output, view: *View) void {
     };
     const actual_box = if (view.saved_buffers.items.len != 0) view.saved_surface_box else view.surface_box;
 
+    var view_box: wlr.Box = .{
+        .x = view.current.box.x,
+        .y = view.current.box.y,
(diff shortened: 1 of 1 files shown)


ifreund (commented):
Hello, thanks for looking into this.

I think however that the root cause was in the `scaleBox()` function and I just pushed f370202b685eb9e9ddeda50acfb06179f9fc82c4 to fix it.

If you can still reproduce the issue after that commit please let me know, I cannot.

MaxVerevkin:
Yep, it's fixed.

PR #772, opened 2023-01-12
View: Fix self pointer in notifyTitle()

### LINKED ISSUE #771: Switching between Firefox tabs no longer updates the window name module in Waybar
Hi!

I'm running river `0.3.0-dev.12+2c0c606` and noticed that the Waybar module that displays the window name/title no longer updates when switching between tabs in Firefox.

By some testing I found that the offending commit seems to be this [one](https://github.com/riverwm/river/commit/09f3f141ae4ade0ef66acb889c1a3db371c1b23f), with the update to zig 0.10. 

Here's a before:

https://user-images.githubusercontent.com/19192649/211902954-51513b43-8e60-4167-a607-5936fd1f3474.mp4

And after:

https://user-images.githubusercontent.com/19192649/211903135-30ab4a1c-a570-48e8-90f9-ccb30d5bcf6f.mp4

Switching between windows still updates the Waybar window module as expected though (which can be seen in the after video). Therefore I'm assuming this is related to river rather than Waybar, but let me know otherwise or if more info is needed.

novakne:
This cause `seat_node.data.focused.view == &self` to be false and so to never send the focused view title to client

Fixed #771

### DIFF
--- a/river/View.zig
+++ b/river/View.zig
@@ -551,14 +551,14 @@ pub fn unmap(self: *Self) void {
     server.root.startTransaction();
 }
 
-pub fn notifyTitle(self: Self) void {
+pub fn notifyTitle(self: *const Self) void {
     if (self.foreign_toplevel_handle) |handle| {
         if (self.getTitle()) |s| handle.setTitle(s);
     }
     // Send title to all status listeners attached to a seat which focuses this view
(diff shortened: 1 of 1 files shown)


ifreund (approved):
LGTM, thanks for investigating this!

I just force pushed to tweak the commit message and rebase onto master.

PR #773, opened 2023-01-12
Cursor: fix incorrect lock manager state assertion

zakvf:
This caused river to panic upon focusing an OR surface via the cursor (e.g. clicking on an OR menu).


ifreund (approved):
Whoops, my bad. Thanks for the fix!

PR #778, opened 2023-01-20
View: Remove uneeded if statement

novakne:
Just come accros this TODO and it has been fixed in zig 0.10.1


ifreund:
Turns out that it's easiest if we carry this workaround until zig 0.11 since there apparently isn't a FreeBSD 0.10.1 tarball on https://ziglang.org/download/ for the CI.

Issue #703, opened 2022-10-06
[Feature Request] Regex for title and app-id filters


aivarsV:
While writing init file and browsing through river source, I noticed, that `app-id` and `title` for float-filter and csd-filter are compared one-to-one exactly. Has noone suggested adding at least minimal regex/globbing (dot for any single caracter and asterisk for any number of any caracter) support for filters? How much work it would require to implement something like this and would it be in line with river project goals?

As an example on use-case, I had issue with firefox sharing indicator back, when I was using sway. I set up floating filter for that window, providing exact window title. After firefox update they changed hyphen in window title and my setup broke. This a few times until I replaced exact title with regex.

Leon-Plickat:
We have talked about this on IRC a few weeks ago. Not sure about full regex, but at least minor globbing is something we'd like to have. How much work an implementation requires strongly depends on how advanced you want to get: If all you want is multi-char globbing strictly at the end and/or start of the match string, it's probably no more than twenty lines. Anything more advanced will require more code though.

aivarsV:
I guess, that wasn't me in IRC.
Now, that you mention, multi-char globbing at start/end of the match string might do the trick in most cases.

sfrique:
Also it would be nice to be able to match `title`, `class` and `app-id`  for combinations

PR #787, opened 2023-02-12
Config: Asterisk globing for csd and float filters

aivarsV:
Filter pattern starting with asterisk will match all titles/app-ids, that ends with filter pattern, while filter with astersik at the end will match beginning of test string.

This implements #703, however I'm not sure about a few points:

* Currently, river crashes if there is filter containig single asterisk and new window spawns. Is it possible, that I'm missing something obvious about zig? (Discalimer: I only have two day experince in zig) Or, maybe riverctl should error out, when passing single asterisk as filter?

* There are few lines of code in `river/command/filter.zig` in `csdFilterUpdateViews`. I guess it will need some adjustments.

### DIFF
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -178,15 +178,24 @@ pub fn deinit(self: *Self) void {
     self.xkb_context.unref();
 }
 
+const FilterKind = enum {
+    @"app-id",
+    title,
+};
+const FilterEffect = enum {
+    float,
+    csd,
+};
+
 pub fn shouldFloat(self: Self, view: *View) bool {
     if (view.getAppId()) |app_id| {
-        if (self.float_filter_app_ids.contains(std.mem.span(app_id))) {
+        if (self.matchesFilterPattern(FilterKind.@"app-id", FilterEffect.float, std.mem.span(app_id))) {
             return true;
         }
     }
 
     if (view.getTitle()) |title| {
-        if (self.float_filter_titles.contains(std.mem.span(title))) {
+        if (self.matchesFilterPattern(FilterKind.title, FilterEffect.float, std.mem.span(title))) {
             return true;
         }
     }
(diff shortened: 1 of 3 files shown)


river/Config.zig
```diff
@@ -178,15 +178,22 @@ pub fn deinit(self: *Self) void {
     self.xkb_context.unref();
 }
 
+const FilterKind = enum {
```
Leon-Plickat, line 181: This enum combines two things that shouldn't be: The effect of the filter (floating, csd) and the string it is matched against (app-id, title). I think it would be a bit cleaner to keep that separate.

aivarsV:
I've addressed @Leon-Plickat suggestion and updated docs, so it should be ok now.

river/Config.zig
```diff
@@ -178,15 +178,24 @@ pub fn deinit(self: *Self) void {
...
+};
+const FilterEffect = enum {
+    float,
+    csd,
+};
+
 pub fn shouldFloat(self: Self, view: *View) bool {
     if (view.getAppId()) |app_id| {
-        if (self.float_filter_app_ids.contains(std.mem.span(app_id))) {
+        if (self.matchesFilterPattern(FilterKind.@"app-id", FilterEffect.float, std.mem.span(app_id))) {
```
Leon-Plickat, line 192:
```suggestion
        if (self.matchesFilterPattern(.@"app-id", .float, std.mem.span(app_id))) {
```

Zig can infer the type here.

river/Config.zig
```diff
@@ -196,16 +205,48 @@ pub fn shouldFloat(self: Self, view: *View) bool {
...
+}
+
+pub fn itemMatchesPattern(fltr: []const u8, item: [:0]const u8) bool {
+    if (fltr[0] == '*') {
+        if (fltr.len - 1 > item.len) return false;
+        if (fltr.len == 1) return true;
+        if (std.mem.eql(u8, fltr[1 .. fltr.len - 1], item[item.len - (fltr.len - 1) .. item.len - 1])) return true;
+    } else if (fltr[fltr.len - 1] == '*') {
+        if (fltr.len - 1 > item.len) return false;
+        if (std.mem.eql(u8, fltr[0 .. fltr.len - 1], item[0 .. fltr.len - 1])) return true;
```
Leon-Plickat, line 247:
```suggestion
pub fn itemMatchesPattern(fltr: []const u8, item: [:0]const u8) bool {
    debug.assert(fltr.len != 0);
    debug.assert(item.len != 0);
    
    if (fltr[0] == '*') {
        if (fltr.len == 1) return true;
        return mem.endsWith(u8, item, flter[1..]);
    } else if (fltr[fltr.len - 1] == '*') {
        return mem.startsWith(u8, item, flter[0..fltr.len - 1]]);
   } else {
       return mem.eql(u8, fltr, item);
   }
```

This if course assumes that the filter `*string*` is illegal. So you should either modify the snippet above to also handle that case or modify the filter command to disallow such filters. The command will have to be modified anyway, to disallow globbing-stars inside strings, which this PR leaves out.

river/Config.zig
```diff
@@ -196,16 +205,48 @@ pub fn shouldFloat(self: Self, view: *View) bool {
...
+        .float => switch (kind) {
+            .title => self.float_filter_titles,
+            .@"app-id" => self.float_filter_app_ids,
+        },
+        .csd => switch (kind) {
+            .title => self.csd_filter_titles,
+            .@"app-id" => self.csd_filter_app_ids,
+        },
+    };
+    var it = filters.keyIterator();
```
Leon-Plickat, line 233: Originally, the lists were kept separate to that `filters.contains()` could be used. This is no longer the case, so can join them all into a single list, which should make the code a bit more straight-forward. Just create a simple struct which points to a string, and also has booth `FilterKind` and `FilterEffect` field. I'd personally also add an arena which owns the strings for simplicity. Then add an ArrayList of that struct (not owned by the arena, growables don't go well with those).

Leon-Plickat (changes requested):

river/Config.zig
aivarsV: Yes, I can see how this makes sense, however I didn't understand "arena" concept. If you could point me to resource or example of it's use, I would consider implementing it.

Leon-Plickat: You can ignore the arena bit. I forgot that filters can also be removed, so the strings need to be free-able, meaning they need to be owned by gpa.

aivarsV:
I think asterisks winthin filter string can be handled as any other character. At least for now.
However I feel as in dead-end with another issue - how to preoperly handle empty titles/app-ids and empty filters. Since it's trivial to create window with empty title or empy app-id, I can imagine how someone might want to make such windows floating or use client side decorations. Of, course, I might be overthinking.

Leon-Plickat:
> empty filters

Just don't allow them. Make it an error if a user tries to set an empty filter.

> empty titles/app-ids

While technically possible, well-behaved clients will not do this. They should match against a `*` filter, since that is supposed to affect everything, but otherwise can be ignored.

PR #793, opened 2023-02-28
Switch to the wlroots scene graph API

### LINKED ISSUE #370: The Pointer Constraints implementation has limitations
In particular, the current implementation only works for the root surface in an xdg surface tree or xwayland surfaces (which have no surface tree). This is highlighted by the usage of `View.fromWlrSurface()` which I tried to remove in 7428519a932ffd97c5127cda78e744ba95186dc7 but discovered that the pointer constraints implementation uses it. However, to implement the protocol properly *any* surface should be a valid target for a pointer constraint, including layer surfaces.

This isn't strictly speaking a bug, at least to my knowledge, as the compositor is free to ignore constraints for pretty much any reason it wants. Making this work for all surfaces would be much better though.

### LINKED ISSUE #519: Floating window moving broken with windows that don't ack configures
Run a client that doesn't respond to configures, or weston-simple-shm with this patch:
```diff
diff --git a/clients/simple-shm.c b/clients/simple-shm.c
index 1c2bea9e..13023a5e 100644
--- a/clients/simple-shm.c
+++ b/clients/simple-shm.c
@@ -130,9 +130,8 @@ handle_xdg_surface_configure(void *data, struct xdg_surface *surface,
 {
        struct window *window = data;
 
-       xdg_surface_ack_configure(surface, serial);
-
        if (window->wait_for_configure) {
+               xdg_surface_ack_configure(surface, serial);
                redraw(window, NULL, 0);
                window->wait_for_configure = false;
        }
```

After the window has been:
1. made floating
2. been tried to be resized

, the window can't be moved any more and the cursor moves very weirdly when trying to move the window. This video showcases the issue:

https://user-images.githubusercontent.com/16741932/151663418-053d8dfd-a250-4e3d-81eb-46c6bd9e3a64.mp4

### LINKED ISSUE #594: Issues with pointer constraints
I don't know whether to make this two separate issues or just one since they're both pointer constraint issues, but I assume its easy to split them up later if they should be split up.

- [ ] 1: When a view starts with pointer constraints (such a gamescope or the wlroots pointer constraints example), the cursor doesn't get moved to the right place on the screen. This is more noticeable with `attach-mode bottom`. I think this might be because the views position and size isn't set by the layout manager immediately when the view is created (I guess this is a reason why Sway waits for the surface to commit before moving the cursor to the constraint area?).
- [ ] 2: There is a very weird camera snapping issue in Minecraft that can be reproduced by: pause > unpause > move mouse > click. I've been using a small patch (I think patch is the correct term) which seems to fix it, but it doesn't seem like a "correct" fix. I tried this in Minetest as well but it doesn't seem to have this issue. Minecraft is the only game that I know has this problem, but then again Minecraft is the only game that I play that has a first person camera. I also don't know for sure if this one is necessarily a "pointer constraints" issue.

<details>
<summary>Patch I use to fix Minecraft camera snapping</summary>

```
diff --git a/river/Cursor.zig b/river/Cursor.zig
index da1b8d6..e8ea9b8 100644
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -962,7 +962,7 @@ fn passthrough(self: *Self, time: u32) void {
         // If input is allowed on the surface, send pointer enter and motion
         // events. Note that wlroots won't actually send an enter event if
         // the surface has already been entered.
-        if (server.input_manager.inputAllowed(result.surface)) {
+        if (self.constraint == null and server.input_manager.inputAllowed(result.surface)) {
             self.seat.wlr_seat.pointerNotifyEnter(result.surface, result.sx, result.sy);
             self.seat.wlr_seat.pointerNotifyMotion(time, result.sx, result.sy);
         }
```
</details>

ifreund:
This is a pretty massive patchset so any testing of this branch would be appreciated to avoid disturbing the master branch's stability too much. It's now at the point where river can be run and seems to function correctly save a few missing features.

Please report any bugs or crashes you find with a comment here :)

Fixes #519
Closes #370
Fixes #594

TODO:
- [x] reimplement foreign-toplevel-management
- [x] reimplement pointer constraints
- [x] Audit scene tree creation/destruction to identify leaks
- [x] Grep for TODOs in the code
- [x] Fix any bugs reported here


Leon-Plickat:
Some issues I encountered:

* Floating views can be rendered behind tiled views.
* No frame perfection on view map / unmap: The updated layout is visually applied before a new view is visible. And if a view is unmapped, there is a frame where it is no longer visible but the layout has not updated yet.
* It is possible resize certain windows too small via pointer, for example foot (the border gets smaller than the surface). For other views (f.e. gnome-weather) minimum view size is respected. Might be a client error.
* Views on the CSD-filter still have borders rendered. (just a visual bug, the non-bordered size is used as expected for moving and resizing)
* fullscreen does not work. Instead, just a black screen (with cursor) is shown. Based on focus-follow cursor not working (moving the cursor off the screen with the fullscreened view to another and back does not result in the fullscreened view being focused again) I guess this is a scene issue and not a rendering issue. This applied both to fullscreen  requested by clients and for fullscreen set with a keymap.

Leon-Plickat:
* Floating CSD views can no longer request move / resize

ifreund:
> - Views on the CSD-filter still have borders rendered. (just a visual bug, the non-bordered size is used as expected for moving and resizing)
> - fullscreen does not work. Instead, just a black screen (with cursor) is shown. Based on focus-follow cursor not working (moving the cursor off the screen with the fullscreened view to another and back does not result in the fullscreened view being focused again) I guess this is a scene issue and not a rendering issue. This applied both to fullscreen requested by clients and for fullscreen set with a keymap.

These two issues and a few others should now be fixed, thanks for testing!

Leon-Plickat:
* a fullscreened floating view that is moved to another output can go outside the output bounds when leaving fullscreen (if the outputs are not congruent)

ifreund:
> a fullscreened floating view that is moved to another output can go outside the output bounds when leaving fullscreen (if the outputs are not congruent)

Should be fixed along with a similar bug affecting `View.float_box`.

ifreund:
> It is possible resize certain windows too small via pointer, for example foot (the border gets smaller than the surface). For other views (f.e. gnome-weather) minimum view size is respected. Might be a client error.

Has anyone managed to reproduce this with anything other than foot?

Leon-Plickat:
> Has anyone managed to reproduce this with anything other than foot?

Weston-terminal, although that one is arguably not a good client.

ifreund:
Ok I see the issue. It looks like foot doesn't actually use the `xdg_toplevel.set_min_size` to tell river about its minimum size.

The change in behavior in this branch is that river draws borders based on the size river requested of the clients rather than the size of the buffer actually committed by the client as done on master branch.

I think I'm going to keep this new behavior as is for now and focus on the other bugs, well-behaved clients should really use the set_min_size request.

Leon-Plickat:
Ok, this is weird:


https://user-images.githubusercontent.com/49593314/222436643-21f2c75f-4865-4f2a-a58c-b479adf2c0bf.mp4

So far only reproducable with imv, but reliably. Sometimes wiggling can be stopped by hiding the view, either by focusing another tag or by hiding it behind a larger view. A session that had a wiggle-window will apparently inevitably crash, sometimes immediately, other times a few seconds late.

Leon-Plickat:
Now I can't get it to crash anymore...

But I discovered the wiggling only happens when tiled views exist. If all views are floating, there is no wiggling.

ifreund:
This behavior looks to be a clear imv bug: https://todo.sr.ht/~exec64/imv/47

I'm not sure yet if it's feasible for river to avoid this kind of loop in practice. I hope this kind of bug isn't too common in clients.

I don't think this bug could lead directly to a crash though and I haven't hit any crash despite being able to reproduce it several times with various imv versions.

vainiovano:
Moving a slow client like firefox after resizing sometimes smoothly warps the cursor:

https://user-images.githubusercontent.com/16741932/222887435-40c85d07-0bec-458f-813e-06fd16fe79f1.mp4

<br>
Also, defocusing a client that doesn't commit after acking the deactivation like paused mpv is slow.

ifreund:
> > Also, defocusing a client that doesn't commit after acking the deactivation like paused mpv is slow.
> 
> Expected. For frame perfection, the transaction system waits for the ack of all affected clients before rendering. This eventually times out, after 200ms. Maybe we could decrease the timeout a bit, but this case will always feel a bit slow.
> 
> While it should be possible to modify the transaction system so that in certain cases it does not wait for acks, I belive that would add a lot of complexity. Clients that don't ack are basically broken anyway, meaning this is not our bug.

Perhaps we should only wait for a ack/commit from clients we request a size change of before committing the transaction. I was considering doing this anyways actually as I think it could make things a bit more responsive without sacrificing frame perfection.

ifreund:
>  Also, defocusing a client that doesn't commit after acking the deactivation like paused mpv is slow. 

Fixed in by the latest commit, I think I like the new code better too.

ifreund:
> Moving a slow client like firefox after resizing sometimes smoothly warps the cursor:

@vainiovano I believe this should be fixed by 4d2f5bb32ddeb7edb0d3720dcbc49c82cc3cc3dd but I haven't been able to reproduce that behavior reliably. If you can still reproduce the bug please let me know.

vainiovano:
Now the cursor is dragging behind windows when moving them

https://user-images.githubusercontent.com/16741932/222949439-884321a1-4541-44c6-ae2a-23cb65324736.mp4

ifreund:
> Now the cursor is dragging behind windows when moving them

@vainiovano thanks, that should theoretically be fixed as well with the latest commit.

This one I cannot reproduce at all though on my beefy desktop PC even while running river under valgrind to artificially slow everything down. What kind of hardware are you running on out of curiosity?

vainiovano:
Yeah, that fixed it!

> This one I cannot reproduce at all though on my beefy desktop PC even while running river under valgrind to artificially slow everything down. What kind of hardware are you running on out of curiosity?

Thinkpad Yoga S1 with Intel HD Graphics 4400. I tested it with both hardware and software cursors, so it can't be about that.

ifreund:
I think it's time to merge this to master, there aren't any regressions I know of and there are many improvements :)

PR #811, opened 2023-03-23
Implement -current-tags for send-to-output

pinpox:
As proposed on IRC, this implements the option to pass `-current-tags` to the `send-to-output` command. The flag will cause views to be tagged with the views of the destination when send to a different output. 

Visually, this results in the view always staying visible on output move.

### DIFF
--- a/river/command/output.zig
+++ b/river/command/output.zig
@@ -17,8 +17,8 @@
 const std = @import("std");
 const assert = std.debug.assert;
 const mem = std.mem;
-
 const wlr = @import("wlroots");
+const flags = @import("flags");
 
 const server = &@import("../main.zig").server;
 
@@ -52,7 +52,13 @@ pub fn sendToOutput(
     _: *?[]const u8,
 ) Error!void {
     if (args.len < 2) return Error.NotEnoughArguments;
-    if (args.len > 2) return Error.TooManyArguments;
+    const result = flags.parser([:0]const u8, &.{
+        .{ .name = "current-tags", .kind = .boolean },
+    }).parse(args[1..]) catch {
+        return error.InvalidOption;
+    };
+    if (result.args.len < 1) return Error.NotEnoughArguments;
+    if (result.args.len > 1) return Error.TooManyArguments;
 
     // If the noop output is focused, there is nowhere to send the view
     if (seat.focused_output == null) {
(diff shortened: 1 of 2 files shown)


river/command/output.zig
```diff
@@ -52,7 +52,20 @@ pub fn sendToOutput(
...
+    };
+    if (result.args.len < 1) return Error.NotEnoughArguments;
+    if (result.args.len > 1) return Error.TooManyArguments;
+
+    var dir_arg = args[1];
+
+    // If -current-tags is set, the direction parameter will be shifted by one
+    if (args.len == 3) {
+        dir_arg = args[2];
+    }
```
Leon-Plickat, line 68: all of this is no longer needed now that you use the args parser.

river/command/output.zig
```diff
@@ -52,7 +52,20 @@ pub fn sendToOutput(
     _: *?[]const u8,
 ) Error!void {
     if (args.len < 2) return Error.NotEnoughArguments;
-    if (args.len > 2) return Error.TooManyArguments;
+    const result = flags.parser([:0]const u8, &.{
```
Leon-Plickat, line 55: You also need to import `flags` before you use it.

Leon-Plickat (changes requested):

.gitignore
```diff
@@ -1,3 +1,4 @@
 zig-cache/
 zig-out/
 doc/*.[1-9]
+result
```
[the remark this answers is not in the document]
pinpox: Just an artifact from me developing on NixOS. removed.

river/command/output.zig
pinpox: removed

river/command/output.zig
pinpox: Done.

river/command/output.zig
```diff
@@ -20,6 +20,7 @@ const mem = std.mem;
 
 const wlr = @import("wlroots");
 
+const flags = @import("flags");
```
Leon-Plickat, line 23: minor nitpick: flags here is a package and should be grouped with the other package import. Right now it is grouped with server, which is a global imported from a local sourcce file.

Leon-Plickat (approved):

river/command/output.zig
pinpox: Hope I understood correctly, I copied the grouping from `map.zig`, should be fixed.

ifreund (approved):
The implementation looks good to me, this just needs to be documented in the riverctl man page before it can be merged.

ifreund (approved):
Thanks!

rickalex21:
Hello,

This is no longer working, not sure why? Was `-current-tags` removed??

```sh
riverctl map normal $mod+Control Semicolon send-to-output -current-tags next
```

```sh
-❯ riverctl  send-to-output -current-tags next
error: too many arguments
```

Leon-Plickat:
@rickalex21 this feature has not landed in a release yet, so you need to build river from source to make use of it. It will not be available yet in any river version packaged in any repository.

rickalex21:
@Leon-Plickat Thanks, perhaps I had a custom build and forgot. I am
having problem building. I opened issue #916 .

PR #815, opened 2023-03-30
[wip] add '-current-tags' option to spawn command

Leon-Plickat:
Not super happy with the implementation, not sure if this is even the right approach. Seems to work though.

Also missing is a way for the tokens to expire.


ifreund:
This seems unfinished and would likely need to be reworked completely for the river window management protocol anyways. Also, I don't immediately see why this would be an option rather than just the default behavior.

Closing for now to clean up the PR queue before codeberg migration.

PR #819, opened 2023-04-07
command/input: match device names against glob pattern

### LINKED ISSUE #741: [Feature Request] Configure inputs by class
I currently use this script to configure my touchpa

```sh
configure_touchpads() {
    riverctl list-inputs \
        | grep 'type: pointer' -B 1 \
        | grep -vE 'type: pointer|^--$' \
        | xargs -I '{}' riverctl input '{}' $@
}

# Touchpad config
configure_touchpads natural-scroll enabled

configure_touchpads tap enabled
configure_touchpads tap-button-map left-right-middle
```

I don't like using the device names, because then I need to maintain different configurations for all my machines.

Ideally there would be a thing analogous to `xorg.conf`'s [Input-Class Selection](https://man.archlinux.org/man/xorg.conf.5#INPUTCLASS_SECTION):

- MatchIsKeyboard
- MatchIsPointer
- MatchIsJoystick
- MatchIsTablet
- MatchIsTabletPad
- MatchIsTouchpad
- MatchIsTouchscreen

Then you could do something like `riverctl input match:touchpad tap enabled`.

jameschensmith:
Integrates the basic glob matching implementation introduced by the rules system into the input command to match device names from a glob pattern.

Closes #741.

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -407,46 +407,54 @@ The _input_ command can be used to create a configuration rule for an input
 device identified by its _name_.
 The _name_ of an input device consists of its type, its numerical vendor id,
 its numerical product id and finally its self-advertised name, separated by -.
+A _glob_ pattern can be used in place of _name_ to match input devices with less
+specificity. A _glob_ is a string that may optionally have an _\*_ at the
+beginning and/or end. An _\*_ in a _glob_ matches zero or more arbitrary
+characters in the app-id or title.
+
+For example, _abc_ is matched by _a\*_, _\*a\*_, _\*b\*_, _\*c_, _abc_, and
+_\*_ but not matched by _\*a_, _b\*_, _\*b_, _c\*_, or _ab_. Note that _\*_
+matches everything while _\*\*_ and the empty string are invalid.
 
 A list of all device properties that can be configured may be found below.
 However note that not every input device supports every property.
 
-*input* _name_ *events* *enabled*|*disabled*|*disabled-on-external-mouse*
+*input* _glob_ *events* *enabled*|*disabled*|*disabled-on-external-mouse*
 	Configure whether the input devices events will be used by river.
 
-*input* _name_ *accel-profile* *none*|*flat*|*adaptive*
+*input* _glob_ *accel-profile* *none*|*flat*|*adaptive*
 	Set the pointer acceleration profile of the input device.
 
-*input* _name_ *pointer-accel* _factor_
+*input* _glob_ *pointer-accel* _factor_
 	Set the pointer acceleration factor of the input device. Needs a float
 	between -1.0 and 1.0.
 
-*input* _name_ *click-method* *none*|*button-areas*|*clickfinger*
+*input* _glob_ *click-method* *none*|*button-areas*|*clickfinger*
 	Set the click method of the input device.
 
-*input* _name_ *drag* *enabled*|*disabled*
+*input* _glob_ *drag* *enabled*|*disabled*
 	Enable or disable the tap-and-drag functionality of the input device.
 
-*input* _name_ *drag-lock* *enabled*|*disabled*
+*input* _glob_ *drag-lock* *enabled*|*disabled*
 	Enable or disable the drag lock functionality of the input device.
 
-*input* _name_ *disable-while-typing* *enabled*|*disabled*
+*input* _glob_ *disable-while-typing* *enabled*|*disabled*
 	Enable or disable the disable-while-typing functionality of the input device.
 
-*input* _name_ *middle-emulation* *enabled*|*disabled*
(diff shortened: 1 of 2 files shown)


doc/riverctl.1.scd
```diff
@@ -407,62 +407,70 @@ The _input_ command can be used to create a configuration rule for an input
...
+
+For example, _abc_ is matched by _a\*_, _\*a\*_, _\*b\*_, _\*c_, _abc_, and
+_\*_ but not matched by _\*a_, _b\*_, _\*b_, _c\*_, or _ab_. Note that _\*_
+matches everything while _\*\*_ and the empty string are invalid.
 
 A list of all device properties that can be configured may be found below.
 However note that not every input device supports every property.
 
-*input* _name_ *events* *enabled*|*disabled*|*disabled-on-external-mouse*
+*input* _name_|_glob_ *events* *enabled*|*disabled*|*disabled-on-external-mouse*
```
ifreund, line 422: From my point of view, an name is just a glob pattern that matches exactly one input.
```suggestion
*input* _glob_ *events* *enabled*|*disabled*|*disabled-on-external-mouse*
```

ifreund (changes requested):
This currently doesn't handle multiple globs matching the same input identifier well.

For the rules system we give the most specific glob the highest priority when matching.

For input configuration however we also need to consider the case where e.g. the second most specific glob explicitly sets configuration that the most specific/highest priority match left default. I think that in this case the configuration from second most specific glob should override the values left default.

Perhaps this simplest way to handle this would be a separate list of globs for each configuration option.

We also need to consider how best to display this information to the user, which likely means adding a command similar to the `list-rules` command.

jameschensmith:
> This currently doesn't handle multiple globs matching the same input identifier well.

Yeah, I see what you mean.

> For the rules system we give the most specific glob the highest priority when matching.
> 
> For input configuration however we also need to consider the case where e.g. the second most specific glob explicitly sets configuration that the most specific/highest priority match left default. I think that in this case the configuration from second most specific glob should override the values left default.

I like it. This change is a little more complicated than I had thought, but I also agree in how it should work, and so I'm down to figure this out.

> Perhaps this simplest way to handle this would be a separate list of globs for each configuration option.
> 
> We also need to consider how best to display this information to the user, which likely means adding a command similar to the `list-rules` command.

I'm going to work through this and come back with some thoughts. Currently, my thought is that `list-input-configs` could be reused for this, where it can display your globs-to-configuration-options set. Then, `list-inputs` could actually have a more verbose option which shows what options are configured for each input device. Still working through this thought, though.

doc/riverctl.1.scd
```diff
@@ -407,62 +407,70 @@ The _input_ command can be used to create a configuration rule for an input
 device identified by its _name_.
 The _name_ of an input device consists of its type, its numerical vendor id,
 its numerical product id and finally its self-advertised name, separated by -.
+A _glob_ pattern can be used in place of _name_ to match input devices with less
```
jameschensmith, line 410: `_name_` with be replaced with `_glob_` in this section once everything is worked out in code.

jameschensmith:
I apologize that I haven't provided an update on this in a while. I've been away from this pull request for a while, but I'm getting back to it with much eagerness 👍

jameschensmith:
Laying out some thoughts here. Would love to hear what you think.

Regarding the input config priority, I had an alternative solution to what you had suggested (from what I understand, you had suggested that each config option had it's own array of globs). My solution would be that whenever a new input config is added, the input managers `configs` array list is sorted by glob specificity (similar to the new rules list that you had brought up). Then, the configuration is applied by looping through each config & device (O(n<sup>2</sup>), trying to think of a more efficient method) to ensure the config priority is applied correctly.

Regarding how best to display the input configurations to the user, I think with the above change made the current output would not be the worst. Exploring the columnar layout that is used in rules list, I don't think it will output well for the input configs (especially if your configuration identifiers are using the full device identifier). I would still suggest the option of having `list-inputs` have a more verbose option which shows what options are configured for each input device. Open to other suggestions, though.

ifreund:
> I apologize that I haven't provided an update on this in a while. I've been away from this pull request for a while, but I'm getting back to it with much eagerness +1

No need to apologize, this is a FOSS project and we are all volunteering our time :)

> My solution would be that whenever a new input config is added, the input managers `configs` array list is sorted by glob specificity (similar to the new rules list that you had brought up). Then, the configuration is applied by looping through each config & device (O(n2), trying to think of a more efficient method) to ensure the config priority is applied correctly.

Having a single list seems reasonable, I don't think performance is a huge concern here due to the small numbers of config options and devices. Let's focus on making the code as simple and robust as possible.

ifreund:
Thanks for working on this, I've ended up landing this feature in https://github.com/riverwm/river/pull/959

PR #828, opened 2023-04-30
Rework move/resize handling

### LINKED ISSUE #812: River ignores cursor movements when resizing slow windows
Because of how river currently ties cursor movement to floating window resizes, the speed of resizing a window depends on which window is being resized and what the application is doing. mpv's resizing speed changes quite a lot depending on whether the video is paused or not:

https://user-images.githubusercontent.com/16741932/227445727-ffec8f13-5808-4f67-b20a-2eaca86a45be.mp4

The same thing happens with Firefox too which resizes faster when the window is smaller.

ifreund:
This branch replaces the absolute x/y/width/height in the pending state of views with accumulated deltas.

It also changes the behavior of resizing to always allow growing a window even if against the edge of the output. This behavior or something like it is, as far as I can tell, necessary to handle clients (such as mpv) responding to the resize with a size larger than what we asked for in one dimension. 

For example resizing mpv along only the x-axis causes it to grow along the y-axis as well to maintain aspect ratio. The xdg-shell protocol states:

> The window geometry specified in the configure event is a maximum; the client cannot resize beyond it. Clients that have aspect ratio or cell sizing configuration can use a smaller size, however.

I'm not sure how else we should expect clients that wish to maintain an aspect ratio to grow in size however. The behavior of mpv in this case seems like the only reasonable thing clients can do and I don't know of any compositor that sends an error in this case.

The current state of this branch is pretty much a proof of concept. The code is really ugly and needs cleanup before it can be merged.

The only known bug is when mpv is paused and I try to resize it vertically, which it refuses. It seems that a transaction timeout in this case (or perhaps in general?) makes it possible for the window to get moved vertically instead of nothing happening. This needs more investigation as well as more careful consideration of how this new approach interacts with transaction timeouts in general.

Any testing would be appreciated, I have yet to test this heavily on slower machines.

Fixes #812

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -79,6 +79,9 @@ const Mode = union(enum) {
         delta_x: f64 = 0,
         delta_y: f64 = 0,
 
+        fixup_x: i32 = 0,
+        fixup_y: i32 = 0,
+
         /// Resize edges, maximum of 2 are set and they may not be opposing edges.
         edges: wlr.Edges,
         /// Offset from the left or right edge
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -711,8 +711,8 @@ pub fn startMove(cursor: *Self, view: *View) void {
 
     const new_mode: Mode = .{ .move = .{
         .view = view,
-        .offset_x = @floatToInt(i32, cursor.wlr_cursor.x) - view.current.box.x,
-        .offset_y = @floatToInt(i32, cursor.wlr_cursor.y) - view.current.box.y,
+        .offset_x = @floatToInt(i32, cursor.wlr_cursor.x) - view.current_box.x,
+        .offset_y = @floatToInt(i32, cursor.wlr_cursor.y) - view.current_box.y,
     } };
     cursor.enterMode(new_mode, view, .move);
 }
@@ -731,7 +731,7 @@ pub fn startResize(cursor: *Self, view: *View, proposed_edges: ?wlr.Edges) void 
         break :blk cursor.computeEdges(view);
     };
 
-    const box = &view.current.box;
+    const box = &view.current_box;
     const lx = @floatToInt(i32, cursor.wlr_cursor.x);
     const ly = @floatToInt(i32, cursor.wlr_cursor.y);
     const offset_x = if (edges.left) lx - box.x else box.x + box.width - lx;
@@ -750,7 +750,7 @@ pub fn startResize(cursor: *Self, view: *View, proposed_edges: ?wlr.Edges) void 
 
 fn computeEdges(cursor: *const Self, view: *const View) wlr.Edges {
     const min_handle_size = 20;
-    const box = &view.current.box;
+    const box = &view.current_box;
 
     var output_box: wlr.Box = undefined;
     server.root.output_layout.getBox(view.current.output.?.wlr_output, &output_box);
@@ -796,7 +796,7 @@ fn enterMode(cursor: *Self, mode: Mode, view: *View, image: Image) void {
     cursor.seat.focus(view);
 
     if (view.current.output.?.layout != null) {
-        view.float_box = view.current.box;
+        view.float_box = view.current_box;
         view.pending.float = true;
     }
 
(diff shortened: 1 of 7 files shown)


ifreund:
Upstream issue: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/33

vainiovano:
- It's a bit unexpected that resizing a window against an output edge resizes it from the other size
- River crashes when resizing mpv too much against an output edge:
```#0  0x00007fc62236e26c in  () at /lib64/libc.so.6
#1  0x00007fc62231ea08 in raise () at /lib64/libc.so.6
#2  0x00007fc622307538 in abort () at /lib64/libc.so.6
#3  0x00000000002a1749 in os.abort () at /usr/lib/zig/std/os.zig:560
#4  0x00000000002608c8 in debug.panicImpl (trace=0x0, first_trace_addr=..., msg=...) at /usr/lib/zig/std/debug.zig:384
#5  0x000000000024d14a in builtin.default_panic (msg=..., error_return_trace=0x0, ret_addr=...) at /usr/lib/zig/std/builtin.zig:839
        first_trace_addr = 2421975
#6  0x000000000024f4d7 in debug.assert (ok=false) at /usr/lib/zig/std/debug.zig:278
#7  0x00000000002a604d in math.clamp__anon_24016 (val=452, lower=2, upper=0) at /usr/lib/zig/std/math.zig:454
#8  0x000000000026c072 in View.resizeUpdatePosition (view=0x1dde9c0, width=1916, height=1078) at View.zig:256
        data = 0x1d925f0
        box = 0x1ddec00
        output_width = 1920
        output_height = 1080
        unconstrained_y = 452
        border_width = 2
        unconstrained_x = 827
#9  0x000000000026c5fd in XdgToplevel.handleCommit () at XdgToplevel.zig:347
        self = 0x1dde9c0
        view = 0x1dde9c0
        old_geometry = {x = 0, y = 0, width = 1897, height = 1068}
        listener = 0x1ddeaa0
#10 wayland_server_core.Listener(*types.compositor.Surface).setNotify__struct_10841.wrapper (listener=0x1ddeaa0, data=0x1d8ea90)
    at /home/naakka/river/zig-cache/zig-wayland/wayland_server_core.zig:559
#11 0x00007fc6226d101e in wl_signal_emit_mutable () at /lib64/libwayland-server.so.0
#12 0x00007fc6221774f6 in  () at /usr/lib/libffi.so.8
#13 0x00007fc622173f5e in  () at /usr/lib/libffi.so.8
#14 0x00007fc622176b73 in ffi_call () at /usr/lib/libffi.so.8
#15 0x00007fc6226cfada in  () at /lib64/libwayland-server.so.0
#16 0x00007fc6226d4180 in  () at /lib64/libwayland-server.so.0
#17 0x00007fc6226d2ae2 in wl_event_loop_dispatch () at /lib64/libwayland-server.so.0
#18 0x00007fc6226d32d7 in wl_display_run () at /lib64/libwayland-server.so.0
#19 0x000000000025e6cc in main.main () at main.zig:132
        startup_command = {ptr = 0x14632e0 "\020\371\353\001", len = 31}
        result = {args = {ptr = 0x7ffd4caff360, len = 0}, flags = {h = false, version = false, c = {ptr = 0x0, len = 0}, log-level = {ptr = 0x0, len = 0}}}
        child_pgid = {data = 24651, some = 1 '\001'}
```
- River crashes when switching tags while a window is being moved or resized
- River crashes when tiling a window that is being moved or resized
- Resizing mpv in a nested river is a lot more broken than previously (the window size just jumps all over the place)
- An unpaused mpv sometimes jumps too when resizing it from above, without a transaction timeout

ifreund:
I'm not terribly happy with the path taken by this and the original #805. A lot of complexity and bugs have been introduced and resizing mpv still doesn't work reliably.

Before continuing work on polishing this branch I plan to explore other options, perhaps leveraging the xdg-toplevel configure_bounds event.

Leon-Plickat:
Is there any chance we could convince enough widely used compositors to implement the xdg-shell to the letter and just kill mpv when it does its weird resizing?

ifreund:
> Is there any chance we could convince enough widely used compositors to implement the xdg-shell to the letter and just kill mpv when it does its weird resizing?

The problem with this is that I agree with the mpv authors that mpv's current behavior is the most intuitive for the user. I'm disappointed that xdg-shell uses mechanism over policy here instead of a additional request to inform the compositor of a fixed aspect ratio or similar (I've been told adding such a request has been shot down before when I brought it up in the #wlroots IRC channel).

PR #829, opened 2023-05-02
editorconfig: fix typo

sagehane:
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#charset seems to suggest that the value should be named `utf-8` and Neovim with `editorconfig` enabled complained for me as-is.


ifreund (approved):
Thanks!

PR #832, opened 2023-05-08
Respect resizes coming from clients

vainiovano:
When a floating window resizes itself, the new size won't automatically end up in wlroots' XdgToplevel.scheduled, so river needs to always set the size when configuring a toplevel.

This makes `weston-fullscreen` no longer jump back to its old size when resized after pressing `z`.


ifreund (approved):
Nice find! I just force pushed to add a comment in the code as well so I don't potentially forget this when refactoring in the future.

Thanks for the patch!

PR #839, opened 2023-06-04
riverctl: Add error messages for failed connections

Leon-Plickat:
Adds error messages for when the connection to the server fails. The case of people trying to use riverctl in a TTY without knowing that it needs `WAYLAND_DISPLAY` seems common enough and I think proper error messages will help a bit here.


ifreund (commented):
Looks like a great improvement, thanks!

ifreund (approved):
Thanks!

PR #852, opened 2023-06-25
Add tag rule

Leon-Plickat:
Testing welcome.


10leej:
Can I get a quick doc on how to make use of these? I dunno the first thing about zip.

Leon-Plickat:
@10leej I already documented them in the manpage. Look for `doc/riverctl.1.scd`.

justinesmithies:
> Testing welcome.

Works a treat thank you !

Leon-Plickat:
I split it into three commits which hopefully makes reviewing a tiny bit nicer.

I decided to change the order of the columns in the `list-rules` command to put "action" last. This prevents us from having to write code to determine the maximum length of the string representation of rules for each new rule.

justinesmithies:
Tested this morning and adding, listing and removing tag rules works as expected.

river/command/rule.zig
```diff
@@ -123,41 +143,51 @@ pub fn listRules(_: *Seat, args: []const [:0]const u8, out: *?[]const u8) Error!
...
+                    else => unreachable,
+                }});
+            }
+        },
+        .tag => {
+            const rules = server.config.tag_rules.rules.items;
+            for (rules) |rule| {
+                try fmt.formatBuf(rule.title_glob, .{ .width = title_column_max, .alignment = .Left }, writer);
+                try fmt.formatBuf(rule.app_id_glob, .{ .width = app_id_column_max, .alignment = .Left }, writer);
+                try writer.print("{}\n", .{rule.value});
```
ifreund, line 188: Would it be nicer to print this in binary with `{b}`?

It would admittedly be prettier to right-align the numbers in the tag column though that would require determining their max width when printed. If printing in binary the printed width is `std.math.log2(value)`.

river/command/rule.zig
```diff
@@ -107,6 +120,13 @@ pub fn ruleDel(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
             apply_ssd_rules();
             server.root.applyPending();
         },
+        .tag => {
+            server.config.tag_rules.del(.{
+                .app_id_glob = app_id_glob,
+                .title_glob = title_glob,
+                .value = undefined, // Will not be accessed.
```
ifreund, line 127: I think this use of `undefined` is a strong indication that the `del()` function's API should be changed to take a `struct { app_id_glob: []const u8, title_glob: []const u8 }` instead.

river/rule_list.zig
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+
+            return null;
+        }
+
+        /// Returns the length of the longest globs.
+        pub fn getMaxGlobLen(self: *const Self) MaxGlobLen {
+            var app_id_len: usize = "*".len;
+            var title_len: usize = "*".len;
```
ifreund, line 118: These initial values aren't necessary for correctness so I think it is less confusing to use 0 here. Slices of length 0 are rejected by the glob validation function.
```suggestion
            var app_id_len: usize = 0;
            var title_len: usize = 0;
```

doc/riverctl.1.scd
```diff
@@ -272,6 +273,8 @@ matches everything while _\*\*_ and the empty string are invalid.
 	  and existing views.
 	- *csd*: Use client-side decorations for the view. Applies to new
 	  and existing views.
+	- *tag*: Set the initial tags of the view. Requires the tags as
+	  an argument.
```
ifreund, line 277:
```suggestion
	  an argument. Applies only to new views.
```

Also, there's an example table from `list-rules` further down that should probably have its column order changed to match the new output.

ifreund (commented):
This is already looking great, I'm mostly just nitpicking style here.

Thanks for implementing this much-requested feature! It's nice to see that the `RuleList` abstraction made this so straightforward :)

river/command/rule.zig
Leon-Plickat: Not sure if binary would be better. Right now all other commands use the integer representation, so for consistency I think it makes sense here as well.

ifreund: I think the base 10 representation makes sense for taking a tag value as input as it allows users to use the standard bitshifting techniques to construct a set of tags.

I really don't think viewing a set of tags in base 10 is that useful though. It seems to me that a user seeing a `5` in the output of this list could be quite confused when the view appears on the tags commonly 1 and 3.

Instead displaying the base 2 representation of `101` seems less likely to be confusing IMO. I suppose there's still the chance that they might expect it to appear on tag one hundred and one, we could consider adding leading zeros to offset that but I'm not convinced that would be worthwhile.

ifreund:
As tiosgz mentioned on IRC, it would be nice to update the shell completions before merging. Otherwise this looks good to go!

Leon-Plickat:
Added the completions

ifreund (approved):
LGTM, Thanks!

PR #855, opened 2023-06-30
river: fix crash when trying to add tag rule without tag argument

### LINKED ISSUE #854: Crash when trying to add empty tag rule
I tested the tag rule and found the following errors: if I don't specify a tag as an argument, it causes the river to crash. If I assign tags 1 2 3 to programs A B C, it works, but if I assign tags 1 3 5 for example, I get errors. Instead of opening on tags 1 3 5 the programs open on tags 1 2 3.  That is, as I understand it is impossible to have intermediate empty tags. In addition, programs are also marked with tag 1, that is, all 3 programs are also visible on the first tag. Just in case, I give a [link](https://pastebin.com/ZEAmY2ch) to my init script, if I'm doing something wrong.

Leon-Plickat:
Fixes #854


ifreund (approved):
Thanks for the fix!

Issue #869, opened 2023-08-10
Loading keyboard layout from xkb files


Isti115:
I have noticed that the support for loading custom keyboard layout from files is currently missing, so I looked into the source and found that you have marked it as a `TODO`.

I managed to get it working with the following patch:
```patch
Submodule deps/zig-xkbcommon contains modified content
diff --git a/deps/zig-xkbcommon/src/xkbcommon.zig b/deps/zig-xkbcommon/src/xkbcommon.zig
index cb81158..ae1ad3b 100644
--- a/deps/zig-xkbcommon/src/xkbcommon.zig
+++ b/deps/zig-xkbcommon/src/xkbcommon.zig
@@ -112,9 +112,11 @@ pub const Keymap = opaque {
     extern fn xkb_keymap_new_from_names(context: *Context, names: ?*const RuleNames, flags: CompileFlags) ?*Keymap;
     pub const newFromNames = xkb_keymap_new_from_names;
 
-    // TODO
-    //extern fn xkb_keymap_new_from_file(context: *Context, file: *FILE, format: Format, flags: CompileFlags) ?*Keymap;
-    //pub const newFromFile = xkb_keymap_new_from_file;
+    pub const FILE = opaque {};
+
+    extern fn xkb_keymap_new_from_file(context: *Context, file: *FILE, format: Format, flags: CompileFlags) ?*Keymap;
+    pub const newFromFile = xkb_keymap_new_from_file;
 
     extern fn xkb_keymap_new_from_string(context: *Context, string: [*:0]const u8, format: Format, flags: CompileFlags) ?*Keymap;
     pub const newFromString = xkb_keymap_new_from_string;
diff --git a/river/command/keyboard.zig b/river/command/keyboard.zig
index 93f5cb9..c089ea7 100644
--- a/river/command/keyboard.zig
+++ b/river/command/keyboard.zig
@@ -26,12 +26,15 @@ const util = @import("../util.zig");
 const Error = @import("../command.zig").Error;
 const Seat = @import("../Seat.zig");
 
+extern fn fopen(__filename: [*c]const u8, __modes: [*c]const u8) *xkb.Keymap.FILE;
+
 pub fn keyboardLayout(
     _: *Seat,
     args: []const [:0]const u8,
     _: *?[]const u8,
 ) Error!void {
     const result = flags.parser([:0]const u8, &.{
+        .{ .name = "file", .kind = .arg },
         .{ .name = "rules", .kind = .arg },
[...]

PR #870, opened 2023-08-12
keyboard: add the ability to load layout from file

Isti115:
As per the discussion on IRC that followed my issue (#869) I have added a separate `riverctl keyboard-layout-file` command with the file path as its single parameter.

The loading mechanism reads the file and passes the buffer to `xkbcommon` instead of the descriptor / handle to avoid dealing with the problem mentioned in #869.


Leon-Plickat:
I am still curious what issue this solves, since custom xkbcommon layouts are already possible, provided you use the canonical path to put your layout file (`$HOME/.xkb/symbols/`).

Isti115:
@Leon-Plickat Oh, I just replied to you in the issue.

I couldn't get that to work, so there's that, but also, even if I could've, I prefer keeping my home directory as clean as possible and store these kinds of things somewhere else. (In this case, under `$XDG_CONFIG_HOME/xkb`, which is supposedly [also a valid location](https://xkbcommon.org/doc/current/md_doc_user_configuration.html) [with even higher precedence], but doesn't work with `keyboard-layout` either.)

Anyway, having the flexibility to load a file from anywhere doesn't hurt in my opinion.

Leon-Plickat:
The reason you could not get it to work is because your custom layout does not have a name.

no-name

If you add your layout file to one of the canonical locations, xkb-common adds it to its database of layouts. Yes, even if you have a local config file, all global configuration is still loaded. Since it has no name, you can't access it.

The name is a quoted string between `xkb_keymap` and `{`.

Isti115:
I just tried adding a name there, but I'm still getting `error: invalid value`.

Leon-Plickat:
Probably. I suspect xkbcommon compiles its layout list on library init.

Isti115:
Sorry for the delay, I have been away from home last week, but so far I still haven't managed to get my layout loaded without my modifications.
@Leon-Plickat are you strongly opposed to merging this? I think that it would be useful even if we figure out the issue with the layout file on my end.
@ifreund What is your opinion on the matter?

Leon-Plickat:
I am not opposed to this, I merely remain unconvinced this is necessary.
My custom layout works find as is and I don't see why yours shouldn't.

Leon-Plickat:
I have converted your layout file into the canonical form. [You can find it here](https://git.sr.ht/~leon_plickat/vortex). Place those directories into `~/.xkb` and make sure all the files you `include` are also in the right place (I was too lazy to hunt those down, but I am sure you have them somewhere). Use `--rules vortex`.

You can check if it compiles with
```sh
xkbcli compile-keymap --layout vortex --rules vortex
```

For `riverctl keyboard-layout` you'll also need to provide the rules.

river/command.zig
```diff
@@ -62,6 +62,7 @@ const command_impls = std.ComptimeStringMap(
         .{ "keyboard-group-destroy",    @import("command/keyboard_group.zig").keyboardGroupDestroy },
         .{ "keyboard-group-remove",     @import("command/keyboard_group.zig").keyboardGroupRemove },
         .{ "keyboard-layout",           @import("command/keyboard.zig").keyboardLayout },
+        .{ "keyboard-layout-file",      @import("command/keyboard.zig").keyboardLayoutFile },
```
Leon-Plickat, line 65: I would prefer either to name this `keyboard-layout-from-file`, since that's more descriptive, or instead to add a `-from-file` option to the current `keyboard-layout` command.

river/command/keyboard.zig
```diff
@@ -58,15 +72,30 @@ pub fn keyboardLayout(
...
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
+    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+    defer _ = gpa.deinit();
+
+    var file = std.fs.cwd().openFile(args[1], .{}) catch return error.CannotOpenFile;
+    defer file.close();
+
+    var file_bytes = file.readToEndAlloc(gpa.allocator(), 512 * 4096) catch return error.CannotReadFile;
```
Leon-Plickat, line 89: This operation has to relevant failures: Either the file isn't readable, or we run out of memory.
```suggestion
    var file_bytes = file.readToEndAlloc(gpa.allocator(), 512 * 4096) catch |err| {
        switch(err) {
             error.OutOfMemory => return err,
            else => return error.CannotReadFile,
        }
    };
```

river/command/keyboard.zig
```diff
@@ -58,15 +72,30 @@ pub fn keyboardLayout(
...
-    }
+pub fn keyboardLayoutFile(
+    _: *Seat,
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
+    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+    defer _ = gpa.deinit();
+
+    var file = std.fs.cwd().openFile(args[1], .{}) catch return error.CannotOpenFile;
```
Leon-Plickat, line 86: Does it really make sense to have two different errors for being unable to open the file and being unable to read the file? I'd combine them.

river/command/keyboard.zig
Leon-Plickat, line 89: Also I think you are leaking this buffer.

Leon-Plickat (changes requested):

Leon-Plickat:
> I would like to be convinced that this feature is necessary before reviewing/merging it.

I don't think it's strictly necessary, however it is arguably more convenient than splitting combined keymap files into separate files to make xkbcommon happy.

So I think it might be interesting to know how common it is to have these combined keymap files. Are they typical for weird / unusual keyboards? If yes, this feature might be nice sugar. Otherwise, if this is not common and just used by people DIYing their own layouts, I think it's fine to ask them to do it the Right Way™ instead.

Isti115:
@Leon-Plickat RE: https://github.com/riverwm/river/pull/870#issuecomment-1686626908

Thank you once again for taking your time with this!
I have followed your instructions by cloning the contents of the linked repository to `~/.xkb`, but I still get `error: invalid value` when I try to load the layout.

`xkbcli` runs successfully (as it also does on my own layouts in `.config/xkb/`) and `xkbcomp` fails as expected (due to the custom keycode mapping names, as it also does during my previous experiments), which I am able to fix by converting all the identifiers to be similar to those used in the commonly distributed layouts. I have uploaded a version here that has this modification and all the includes inlined as well.
[Vortex-Core-merged-compat.txt](https://github.com/riverwm/river/files/12400775/Vortex-Core-merged-compat.txt)

I would think that this issue is orthogonal to the current one though, since as far as I know `xkbcomp` only plays a role if I wanted to use `xwayland`.

Leon-Plickat:
On Mon Aug 21, 2023 at 8:54 PM CEST, István Donkó wrote:
> Thank you once again for taking your time with this!
> I have followed your instructions by cloning the contents of the
> linked repository to `~/.xkb`, but I still get `error: invalid value`
> when I try to load the layout.

I believe xkbcommon caches layouts, so river needs to be restarted.

If you can compile the vortex layout with xkbcli, it should work in
river. Otherwise track down the errors xkcli shows you.

ifreund:
> I believe xkbcommon caches layouts, so river needs to be restarted.

Would the dedicated keymap-file command work around this caching? That would be another point in favor of adding it.

Restarting river to tweak configuration isn's fun.

Isti115:
Ah, I forgot about the caching since I brought it up last time (https://github.com/riverwm/river/pull/870#issuecomment-1676103402), and voilà, after restarting `river` and having the layout definitions in their _"canonical form"_ as provided by @Leon-Plickat I finally managed to load the layout without getting an `invalid value` error, but strangely only some of my keys work. :open_mouth:

For example none of the alphanumeric keys got properly assigned, they just show up as `NoSymbol` when checked through `wev`. At first I thought that this was due to those having my custom five level type, but I couldn't manage to get them working in any other way either, and I feel like I cannot reasonably allocate any more time to debugging this issue after already having an alternative solution that works reliably.

Also, iterating on this was pretty painful due to having to restart `river` each time (which didn't seem to reliably clear the cache either :raised_eyebrow:), so I also wanted to mention the faster (and less drastic) feedback loop as an advantage to the ability to directly load from the contents of a file. 

Another point could be the fact that this feature is present in other major window managers (such as Sway and Hyprland), so if it's missing from `river`, it could be an annoying issue for those who try to migrate. (The number of people using completely custom layouts being pretty low is a fair counterpoint though.)

Leon-Plickat:
> Would the dedicated keymap-file command work around this caching?

I think it would.

IIRC the part of xkbcommon that applies a keymap is separate from the part that compiles a keymap from the various files and caches them.

Isti115:
I guess that this wasn't entirely clear from my previous message, but as far as I understand, there could technically be no caching issue with my approach, since `xkbcommon` doesn't even get the file path, I just pass the contents through a buffer, which it could cache as much as it wanted, that couldn't cause any issues, since the same contents should compile to the same layouts. :grinning: 

Anyway, I did test it before I wrote my previous reply, and `keyboard-layout-file` does properly apply the changes each time. I'll get to the requested code changes now.

river/command/keyboard.zig
```diff
@@ -58,15 +72,37 @@ pub fn keyboardLayout(
...
-    }
+pub fn keyboardLayoutFile(
+    _: *Seat,
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
+    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+    const allocator = gpa.allocator();
+    defer _ = gpa.deinit();
+
```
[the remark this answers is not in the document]
Isti115: The answer if very simple, because I didn't know that it existed. :smiley: 
Thanks for the review, I'll include this into the `fixup`!

river/command/keyboard.zig
```diff
@@ -58,15 +72,33 @@ pub fn keyboardLayout(
...
-    }
+pub fn keyboardLayoutFile(
+    _: *Seat,
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
+    var file = std.fs.cwd().openFile(args[1], .{}) catch return error.CannotReadFile;
+    defer file.close();
+
+    var file_bytes = file.readToEndAlloc(util.gpa, 512 * 4096) catch |err| {
```
Leon-Plickat, line 86: Not a fan of uncommented magic numbers in code, TBH.

river/command.zig
```diff
@@ -112,6 +113,8 @@ pub const Error = error{
     InvalidOrientation,
     InvalidRgba,
     InvalidValue,
+    CannotReadFile,
+    CannotParseFile,
```
Leon-Plickat, line 117:
```suggestion
```

river/command.zig
```diff
@@ -155,6 +158,8 @@ pub fn errToMsg(err: Error) [:0]const u8 {
         Error.InvalidOrientation => "invalid orientation. Must be 'horizontal', or 'vertical'",
         Error.InvalidRgba => "invalid color format, must be hexadecimal 0xRRGGBB or 0xRRGGBBAA",
         Error.InvalidValue => "invalid value",
+        Error.CannotReadFile => "cannot read file",
+        Error.CannotParseFile => "cannot parse file",
```
Leon-Plickat, line 162:
```suggestion
```

river/command/keyboard.zig
```diff
@@ -26,6 +26,20 @@ const util = @import("../util.zig");
 const Error = @import("../command.zig").Error;
 const Seat = @import("../Seat.zig");
 
+fn applyLayout(keymap: *xkb.Keymap) Error!void {
```
Leon-Plickat, line 29: This function never returns an error.
```suggestion
fn applyLayout(keymap: *xkb.Keymap) void {
```

Leon-Plickat (changes requested):

river/command/keyboard.zig
Isti115: Well, I stole that one from here:
https://github.com/ifreund/zig-wayland/blob/7f528883e057b282b1cb51db2899f9183b2e5a3b/src/scanner.zig#L191

It's a huge overkill for this purpose, but I don't know what's the best way to come up with a reasonable estimated upper limit for the size of XKB files. :thinking:

river/command/keyboard.zig
Isti115: Oh, wow, I've seen Zig (rightfully) complain about so many things during static code checking, how come it didn't catch this one? :open_mouth:

Leon-Plickat: Because this can not be an error. Imagine a scenario where there is an interface of some sort with predefined function pointers and you are the developer of an implementation. While the interface function pointers may allow for an error to occur, your implementation might not actually be capable of failure. Still the return type of your function needs to be an error union to satisfy the interface.

Isti115: Hmm, but why can't the function without error act as a subtype to a function with an error?

river/command/keyboard.zig
Leon-Plickat: The size isn't the issue, just that it is arbitrary. `readToEndAlloc()` will only allocate as much as is needed to store the contents of the file. As such, I'd use `math.maxInt(usize)`.

And one more thing I forgot on the first review round: `readToEndAlloc()` can return `error.FileTooBig`, which should make your function return `error.OutOfMemory`.

river/command/keyboard.zig
Leon-Plickat: I wasn't discussing programming language theory, just pointing out the current state of the language

river/command.zig
Isti115: I think that providing a read error and a parse error has meaningful difference, for example, providing a wrong path leads to a read error, while providing a path to a file with contents that XKB cannot compile leads to a parse error.

Isti115:
I think that with https://github.com/riverwm/river/commit/b41f1f8d8d3a0e944916306a71a53ade34611b86 I have now addressed all the reviews except for removing `CannotParseFile`, which I don't agree with currently. What are other's opinion on that? https://github.com/riverwm/river/pull/870#discussion_r1308538658

river/command/keyboard.zig
```diff
@@ -58,15 +72,34 @@ pub fn keyboardLayout(
...
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
+    var file = std.fs.cwd().openFile(args[1], .{}) catch return error.CannotReadFile;
+    defer file.close();
+
+    var file_bytes = file.readToEndAlloc(util.gpa, std.math.maxInt(usize)) catch |err| {
+        switch (err) {
+            error.FileTooBig => return error.OutOfMemory,
+            error.OutOfMemory => return error.OutOfMemory,
```
Leon-Plickat, line 89: Nit pick:

```suggestion
            error.OutOfMemory, error.FileTooBig => return error.OutOfMemory,
```

Isti115:
@Leon-Plickat @ifreund Sorry for the ping, but it has now been a week since the last response. Do you think that there is anything else to be done before this can be merged?

river/command/keyboard.zig
```diff
@@ -58,15 +72,33 @@ pub fn keyboardLayout(
...
-    }
+pub fn keyboardLayoutFile(
+    _: *Seat,
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
+    var file = std.fs.cwd().openFile(args[1], .{}) catch return error.CannotReadFile;
+    defer file.close();
+
+    var file_bytes = file.readToEndAlloc(util.gpa, std.math.maxInt(usize)) catch |err| {
```
ifreund, line 86:
```suggestion
    const file_bytes = file.readToEndAlloc(util.gpa, std.math.maxInt(usize)) catch |err| {
```

river/command/keyboard.zig
```diff
@@ -58,15 +72,33 @@ pub fn keyboardLayout(
...
-        // wlroots will log an error if this fails and there's unfortunately
-        // nothing we can really do in the case of failure.
-        _ = wlr_keyboard.setKeymap(new_keymap);
-    }
+pub fn keyboardLayoutFile(
+    _: *Seat,
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
+    var file = std.fs.cwd().openFile(args[1], .{}) catch return error.CannotReadFile;
```
ifreund, line 83:
```suggestion
    const file = std.fs.cwd().openFile(args[1], .{}) catch return error.CannotReadFile;
```

river/command/keyboard.zig
ifreund, line 86: Let's cap this at something exceedingly large but not ludicrous like the current value (millions of terrabytes). I suggest one GiB or (`1024 * 1024 * 1024).

It will probably never be relevant, but in the case that some poor user accidentally mistakes their 100 GiB 16k@240 fps video for a keymap file they won't cause their system (likely running with overcommit enabled) to grind to a halt or cause river to get OOM killed.

river/command/keyboard.zig
```diff
@@ -58,15 +72,33 @@ pub fn keyboardLayout(
...
-        const wlr_keyboard = device.wlr_device.toKeyboard();
-        // wlroots will log an error if this fails and there's unfortunately
-        // nothing we can really do in the case of failure.
-        _ = wlr_keyboard.setKeymap(new_keymap);
-    }
+pub fn keyboardLayoutFile(
+    _: *Seat,
+    args: []const [:0]const u8,
+    _: *?[]const u8,
+) Error!void {
```
ifreund, line 82: This function must check that there are no extra arguments passed and that enough arguments are passed, something like:
```zig
    if (args.len > 2) return Error.TooManyArguments;
```

doc/riverctl.1.scd
```diff
@@ -389,6 +389,10 @@ matches everything while _\*\*_ and the empty string are invalid.
 	"grp:ctrl_space_toggle"). See *xkeyboard-config*(7) for possible values and
 	more information.
 
+*keyboard-layout-file* _path_
+	Sets the XKB layout for all keyboards from a definition file at the provided
```
ifreund, line 393: Consistent tense.
```suggestion
	Set the XKB layout for all keyboards from a definition file at the provided
```
It would also be nice to reference some documentation on the format used by xkbcommon if such documentation exists (ideally as a man page). Is this the same `*xkeyboard-config*(7)` we mention above?

ifreund (changes requested):
Hello, thanks for your patience!

Here's my proper in-depth review. When these comments are addressed this should be good to go.

Isti115:
@ifreund Thanks for the review, I'm a bit overwhelmed by work right now, but I'll make sure to finish this properly when I have the capacity!

ifreund:
> @ifreund Thanks for the review, I'm a bit overwhelmed by work right now, but I'll make sure to finish this properly when I have the capacity to do so!

No worries at all, take your time!

Isti115:
Finally getting around to cleaning this up! I have rebased my previous changes and updated the fixup commit with the changes from the suggestion. Thanks again for the great piece of software, thorough review and patience! I have had some more ideas in the last few weeks of using River, so I will try to find more time for contributions either here, or in the form of smaller utilities or layout generators! :)

ifreund (approved):
Thanks for the patch! I did a bit more cleanup myself and added a link to the best documentation I could find on the XKB keymap file format.

PR #871, opened 2023-08-18
Add output rule

Doclic:
This adds the ability to create rules that change the initial output of a view.
(Sorry if my code is not that great, I haven't used Zig a lot, feedback is welcome)

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -276,6 +276,12 @@ matches everything while _\*\*_ and the empty string are invalid.
 	  and existing views.
 	- *tag*: Set the initial tags of the view. Requires the tags as
 	  an argument. Applies only to new views.
+	- *output*: Set the initial output of the view. Requires the output
+	  as an argument. Applies only to new views. The output can be specified
+	  either by connector name (such as _HDMI-A-1_, or _DP-2_), or by
+	  identifier in the form of _MAKE MODEL SERIAL_, for example for an output
+	  with make: _HP Inc._, model: _HP 22w_, and serial: _CNC93720WF_, the
+	  identifier would be: _HP Inc. HP 22w CNC93720WF_.
 
 	Both *float* and *no-float* rules are added to the same list,
 	which means that adding a *no-float* rule with the same arguments
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -17,15 +17,19 @@
 const Self = @This();
 
 const std = @import("std");
+const fmt = std.fmt;
 const mem = std.mem;
 const globber = @import("globber");
 const xkb = @import("xkbcommon");
 
+const server = &@import("main.zig").server;
 const util = @import("util.zig");
 
 const Server = @import("Server.zig");
+const Output = @import("Output.zig");
 const Mode = @import("Mode.zig");
 const RuleList = @import("rule_list.zig").RuleList;
+const View = @import("View.zig");
 
 pub const AttachMode = enum {
     top,
@@ -76,6 +80,7 @@ modes: std.ArrayListUnmanaged(Mode),
 float_rules: RuleList(bool) = .{},
 ssd_rules: RuleList(bool) = .{},
 tag_rules: RuleList(u32) = .{},
+output_rules: RuleList([]const u8) = .{},
 
 /// The selected focus_follows_cursor mode
 focus_follows_cursor: FocusFollowsCursorMode = .disabled,
@@ -152,9 +157,33 @@ pub fn deinit(self: *Self) void {
     self.float_rules.deinit();
     self.ssd_rules.deinit();
     self.tag_rules.deinit();
+    for (self.output_rules.rules.items) |rule| {
+        util.gpa.free(rule.value);
+    }
+    self.output_rules.deinit();
 
     util.gpa.free(self.default_layout_namespace);
 
     self.keymap.unref();
     self.xkb_context.unref();
+}
+
+pub fn outputRuleMatch(self: *Self, view: *View) !?*Output {
+    const output_name = self.output_rules.match(view) orelse return null;
+    var it = server.root.active_outputs.iterator(.forward);
+    while (it.next()) |output| {
+        const wlr_output = output.wlr_output;
+        if (mem.eql(u8, output_name, mem.span(wlr_output.name))) return output;
+
+        // This allows matching with "Maker Model Serial" instead of "Connector"
+        const maker = wlr_output.make orelse "Unknown";
+        const model = wlr_output.model orelse "Unknown";
+        const serial = wlr_output.serial orelse "Unknown";
+        const identifier = try fmt.allocPrint(util.gpa, "{s} {s} {s}", .{ maker, model, serial });
+        defer util.gpa.free(identifier);
+
+        if (mem.eql(u8, output_name, identifier)) return output;
+    }
+
+    return null;
 }
--- a/river/View.zig
+++ b/river/View.zig
@@ -492,7 +492,8 @@ pub fn map(view: *Self) !void {
         view.pending.ssd = ssd;
     }
 
-    if (server.input_manager.defaultSeat().focused_output) |output| {
+    const focused_output = server.input_manager.defaultSeat().focused_output;
+    if (try server.config.outputRuleMatch(view) orelse focused_output) |output| {
         // Center the initial pending box on the output
         view.pending.box.x = @divTrunc(@max(0, output.usable_box.width - view.pending.box.width), 2);
         view.pending.box.y = @divTrunc(@max(0, output.usable_box.height - view.pending.box.height), 2);
--- a/river/command/rule.zig
+++ b/river/command/rule.zig
@@ -105,13 +105,13 @@ pub fn ruleDel(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
 
     switch (action) {
         .float, .@"no-float" => {
-            server.config.float_rules.del(.{
+            _ = server.config.float_rules.del(.{
                 .app_id_glob = app_id_glob,
                 .title_glob = title_glob,
             });
         },
         .ssd, .csd => {
-            server.config.ssd_rules.del(.{
+            _ = server.config.ssd_rules.del(.{
                 .app_id_glob = app_id_glob,
                 .title_glob = title_glob,
             });
@@ -119,7 +119,7 @@ pub fn ruleDel(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
             server.root.applyPending();
         },
         .tag => {
-            server.config.tag_rules.del(.{
+            _ = server.config.tag_rules.del(.{
                 .app_id_glob = app_id_glob,
                 .title_glob = title_glob,
             });
--- a/river/command/rule.zig
+++ b/river/command/rule.zig
@@ -33,6 +33,7 @@ const Action = enum {
     ssd,
     csd,
     tag,
+    output,
 };
 
 pub fn ruleAdd(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void {
@@ -49,7 +50,7 @@ pub fn ruleAdd(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
 
     const positional_arguments_count: u8 = switch (action) {
         .float, .@"no-float", .ssd, .csd => 1,
-        .tag => 2,
+        .tag, .output => 2,
     };
     if (result.args.len > positional_arguments_count) return Error.TooManyArguments;
     if (result.args.len < positional_arguments_count) return Error.NotEnoughArguments;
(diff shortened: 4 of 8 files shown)


river/Config.zig
```diff
@@ -76,6 +77,7 @@ modes: std.ArrayListUnmanaged(Mode),
 float_rules: RuleList(bool) = .{},
 ssd_rules: RuleList(bool) = .{},
 tag_rules: RuleList(u32) = .{},
+output_rules: RuleList(*Output) = .{},
```
Leon-Plickat, line 80: Using `*Output` here is problematic, because this way it's only possible to create rules for currently existing outputs. Also the pointer and therefore the rule will become invalid when the output disconnects (should the output reconnect, the pointer is still invalid).

It's probably better to use output name / identifier here instead.

river/View.zig
```diff
@@ -502,7 +502,7 @@ pub fn map(view: *Self) !void {
         view.pending.ssd = ssd;
     }
 
-    if (server.input_manager.defaultSeat().focused_output) |output| {
+    if (if (server.config.output_rules.match(view)) |output| output else server.input_manager.defaultSeat().focused_output) |output| {
```
Leon-Plickat, line 505: This will cause a crash should the output be disconnected. Proper fix is to use output name /identifier in the rule instead of `*Output`. Also the line is a it to long.

Leon-Plickat (changes requested):
Instead of `*Output`, use `[]const u8` with the output name / identifier and match against outputs when the rule is applied instead of when it's created.

That will also:
* make it possible to define rules for outputs not currently connected
* make output rules survive output reconnect

river/View.zig
```diff
@@ -502,18 +503,30 @@ pub fn map(view: *Self) !void {
         view.pending.ssd = ssd;
     }
 
-    if (server.input_manager.defaultSeat().focused_output) |output| {
+    var output: ?*Output = null;
+    if (server.config.output_rules.match(view)) |output_name| {
+        // Check if an output matches by name
```
Leon-Plickat, line 508: Would be neat if this checked against output name and connector name.

river/View.zig
```diff
@@ -502,18 +503,30 @@ pub fn map(view: *Self) !void {
         view.pending.ssd = ssd;
     }
 
-    if (server.input_manager.defaultSeat().focused_output) |output| {
+    var output: ?*Output = null;
+    if (server.config.output_rules.match(view)) |output_name| {
```
Leon-Plickat, line 507:
```suggestion
    const output: *Output = blk: {
        if (server.config.output_rules.match(view)) |output_name| {
        var it = server.root.outputs.first;
        while (it) |node| : (it = node.next) {
            if (mem.eql(u8, mem.span(node.data.wlr_output.name), output_name)) {
                break :blk &node.data;
            }
        }
        break :blk sever.input_manager.defaultSeat().focused_output;
    };

```

river/command/rule.zig
```diff
@@ -125,6 +134,12 @@ pub fn ruleDel(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
                 .title_glob = title_glob,
             });
         },
+        .output => {
+            server.config.output_rules.del(.{
```
Leon-Plickat, line 138: You need to free the output name.

Leon-Plickat (changes requested):

river/View.zig
Doclic: Sorry, if I sound stupid, but what's a connector?

Leon-Plickat: Outputs can be addressed either by connector (for example `HDMI-A-1`) or by description (like "LG Electronics LG ULTRAWIDE 0x00001F2D").

Check against both `wlr_output.name` and `wlr_output.description`. Note that the latter doesn't always exist, so check for `null` first.

Doclic: > Outputs can be addressed either by connector (for example `HDMI-A-1`) or by description (like "LG Electronics LG ULTRAWIDE 0x00001F2D").
> 
> Check against both `wlr_output.name` and `wlr_output.description`. Note that the latter doesn't always exist, so check for `null` first.

I've checked the `wlr_output.description` for my monitors, and they both include the connector; for example: `Philips Consumer Electronics Company Philips 220E AU51014013941 (DP-1)`, so you might as well just use the connector name directly.
Furthermore, `command/output.zig`'s `getOutput` function (used by command such as `riverctl focus-output`) only checks for `wlr_output.name`, so using the monitor's name doesn't work.

I've also checked `wlr/types/wlr_output.h` for alternatives, and I've found `struct wlr_output`'s `make`, `model`, and `serial` fields, but for some reason using these if they're not null in Zig crashes the compositor

Doclic:
Should I squash the commits?

river/View.zig
Leon-Plickat: > I've checked the wlr_output.description for my monitors, and they both include the connector

That's unfortunate, however comparing against output description is still a valid use case, not everybody has a stationary setup. I think it makes sense to concatenate make, model and serial and compare against that.

river/command/rule.zig
```diff
@@ -85,6 +87,13 @@ pub fn ruleAdd(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
                 .value = tag,
             });
         },
+        .output => {
+            try server.config.output_rules.add(.{
+                .app_id_glob = app_id_glob,
+                .title_glob = title_glob,
+                .value = try util.gpa.dupe(u8, result.args[0]),
```
Leon-Plickat, line 94: The duped string leaks when adding the rule fails. Allocate the string first and use `errdefer` to free it on error.

Leon-Plickat (changes requested):

river/View.zig
Doclic: > That's unfortunate, however comparing against output description is still a valid use case, not everybody has a stationary setup. I think it makes sense to concatenate make, model and serial and compare against that.

I have no idea why, but reading these fields crashes river, at least with [this code](https://github.com/riverwm/river/files/12463232/list-outputs-cmd.patch.gz) (patch had to be gzipped because GitHub is weird)

Leon-Plickat: Have you tried not dereferencing the pointers?

Doclic:
@Leon-Plickat any updates on this?

Leon-Plickat:
The output identifier is still problematic IMO. In the wlroots ecosystem of tools one can usually refer to an output via adapter name (`HDMI-A-1`, `eDP-1`, `VGA-1`, etc.) and via a string unique to the device (for example `LG Electronics LG ULTRAWIDE 0x00001F2D`). It's unfortunate that wlroots has changed and now adds the adapter name to the output name, but this has valid use cases and we should support it.

Doclic:
I fixed merge conflicts and made it use the new syntax for `riverctl rule-add`

river/command/rule.zig
```diff
@@ -100,29 +111,28 @@ pub fn ruleDel(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
...
-                .app_id_glob = app_id_glob,
-                .title_glob = title_glob,
-            });
+            server.config.tag_rules.del(rule);
+        },
+        .output => {
+            if (server.config.output_rules.find(rule)) |output_rule| {
+                util.gpa.free(output_rule.value);
+            }
+            server.config.output_rules.del(rule);
```
ifreund, line 135: Instead of adding a separate `find()` function and searching twice, why don't we make `del()` return the deleted rule's value as an optional (`?T`)?

doc/riverctl.1.scd
```diff
@@ -276,6 +276,8 @@ matches everything while _\*\*_ and the empty string are invalid.
 	  and existing views.
 	- *tag*: Set the initial tags of the view. Requires the tags as
 	  an argument. Applies only to new views.
+	- *output*: Set the initial output of the view. Requires the output
+	  as an argument. Applies only to new views.
```
ifreund, line 280: This should document how output names/identifiers are matched.

river/View.zig
```diff
@@ -492,18 +493,35 @@ pub fn map(view: *Self) !void {
...
+                    mem.eql(u8, mem.span(output.wlr_output.description.?), output_name)))
+                {
+                    break :blk output;
+                }
+            }
+        }
+        break :blk server.input_manager.defaultSeat().focused_output;
+    };
+
+    if (output != null) {
```
ifreund, line 513: I think this could be cleaned up a bit.

The part checking for an output rule match is a bit verbose and the details of the search are not relevant to the rest of the View.map() function. I think it belongs in a different function, `Config.zig` would be a fine place for such a function.

Then, one could write something like
```zig
if (Config.outputRuleMatch(view) orelse focused_output) |output| {
...
}
```
In any case, the `.?` unwrapping in the block below is unnecessary and should be eliminated with an unwrapping `if (foo) |bar|` if statement instead of a `!= null` check.

river/command/rule.zig
```diff
@@ -100,29 +111,28 @@ pub fn ruleDel(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
...
+        .app_id_glob = result.flags.@"app-id" orelse "*",
+        .title_glob = result.flags.title orelse "*",
+    };
     switch (action) {
         .float, .@"no-float" => {
-            server.config.float_rules.del(.{
-                .app_id_glob = app_id_glob,
-                .title_glob = title_glob,
-            });
+            server.config.float_rules.del(rule);
```
ifreund, line 121: Nice cleanup :)

ifreund (changes requested):
Hey, thanks for working on this!

As for the question of what exactly to use in order to identify outputs, perhaps it's useful to look at what kanshi does as that seems to be sufficient in practice for dynamic setups:
https://git.sr.ht/~emersion/kanshi/tree/f6e0c97dd77f0db9bec631810a44175e0836db06/item/main.c#L26-40

TLDR; It matches against both the connector name and an "identifier" created by combining the make, model, and serial number. The description is not used as far as I see.

This seems like a practical approach as the description is not intended to identify ouput devices uniquely, it is intended as some user-facing snippet.

I think we should follow kanshi's approach here unless anyone sees an issue with it.

Doclic:
> As for the question of what exactly to use in order to identify outputs, perhaps it's useful to look at what kanshi does as that seems to be sufficient in practice for dynamic setups: https://git.sr.ht/~emersion/kanshi/tree/f6e0c97dd77f0db9bec631810a44175e0836db06/item/main.c#L26-40

I implemented that, but the `make`, `model`, and `serial` fields were wrongly typed in the `zig-wlroots` bindings, so I fixed them and made an MR for it at https://github.com/swaywm/zig-wlroots/pull/66.
For now, I made this use my fork of the bindings to allow it to compile.

I also implemented every other change you suggested.

And I think that we should implement a `riverctl list-outputs` command to let people know what outputs they have connected, maybe it could look like this, similar to `riverctl list-inputs`:
```
Philips Consumer Electronics Company Philips 220E AU51014013941
        connector: DP-2

HP Inc. HP 22w CNC93720WF
        connector: HDMI-A-1
```
or it could just print `wlr_output.description`:
```
Philips Consumer Electronics Company Philips 220E AU51014013941 (DP-2 via VGA)
HP Inc. HP 22w CNC93720WF (HDMI-A-1)
```

Leon-Plickat:
> And I think that we should implement a riverctl list-outputs

There is already [wlr-randr](https://sr.ht/~emersion/wlr-randr/).

```sh
$  wlr-randr 
eDP-1 "SDC 0x4C48 (eDP-1)"
  Make: SDC
  Model: 0x4C48
  Serial: (null)
  Physical size: 290x160 mm
  Enabled: no
  Modes:
    1920x1080 px, 47.988998 Hz
    1920x1080 px, 59.995998 Hz (preferred, current)
  Position: 0,0
  Transform: normal
  Scale: 1.000000
```

Doclic:
> There is already [wlr-randr](https://sr.ht/~emersion/wlr-randr/).

Ah, my bad.

ifreund:
A `riverctl list-outputs` command may be worthwhile in order to make it clear exactly what unique identifier river uses for rules. It would however be pretty useless without also printing all the other information already provided by wlr-randr though.

I propose that we don't add a `riverctl list-outputs` command in this PR but instead explain that the identifier has the form `$MAKE $MODEL $SERIAL` and that those values may be determined using other tools such as wlr-randr

river/rule_list.zig
```diff
@@ -82,17 +82,28 @@ pub fn RuleList(comptime T: type) type {
...
+
+        pub fn find(list: *Self, rule: struct { app_id_glob: []const u8, title_glob: []const u8 }) ?Rule {
+            for (list.rules.items) |existing| {
+                if (mem.eql(u8, rule.app_id_glob, existing.app_id_glob) and
+                    mem.eql(u8, rule.title_glob, existing.title_glob))
+                {
+                    return existing;
                 }
             }
+            return null;
```
ifreund, line 106: This can be deleted as it's now unused.

river/command/rule.zig
```diff
@@ -24,6 +24,7 @@ const server = &@import("../main.zig").server;
 const util = @import("../util.zig");
 
 const Error = @import("../command.zig").Error;
+const Output = @import("../Output.zig");
```
ifreund, line 27:
```suggestion
```

river/View.zig
```diff
@@ -20,6 +20,7 @@ const build_options = @import("build_options");
 const std = @import("std");
 const assert = std.debug.assert;
 const math = std.math;
+const mem = std.mem;
```
ifreund, line 23:
```suggestion
```

ifreund (approved):
With this bit of minor cleanup and the submodule switch back over to upstream zig-wlroots this should be good to go!

ifreund (approved):
LGTM, Thanks!

Issue #865, opened 2023-07-28
Feature Request: rules for: (initial) window-size, fullscreen, (initial) position, snap


simonhughxyz:
It would be nice to have some more rule options such as setting a default window X,Y size, position and maybe snap top/bottom/left/right.
The main reason I would find this useful is because I spawn terminals with fzf to perform various functions that I would like to have a pre-configured certain size and position. 

Something like:
`riverctl rule-add size --app-id "APP" 400x300`

And for position:
`riverctl rule-add position --app-id "APP" 300x500`

And for snap:
`riverctl rule-add snap --app-id "APP" top`

Leon-Plickat:
These should be fairly trivial to implement. Just change the initial parameters of the views, the rest should work as expected. However care must be taken to avoid placing the view outside of the output bounds.

Also I think they should not auto-float views.

justinesmithies:
> It would be nice to have some more rule options such as setting a default window X,Y size, position and maybe snap top/bottom/left/right. The main reason I would find this useful is because I spawn terminals with fzf to perform various functions that I would like to have a pre-configured certain size and position.
> 
> Something like: `riverctl rule-add size --app-id "APP" 400x300`
> 
> And for position: `riverctl rule-add position --app-id "APP" 300x500`
> 
> And for snap: `riverctl rule-add snap --app-id "APP" top`

I'd really like to see this implemented as described by @simonhughxyz as it would be immensely useful.

JustinLovinger:
Your `sticky` behavior can be implemented by giving a window all tags.

ralphptorres:
> Your `sticky` behavior can be implemented by giving a window all tags.

Indeed, although this has a potentially unwanted side effect: all tags will be occupied. So, say, if you set your bar to conditionally display tag numbers depending on window occupancy, then all tags are flagged and displayed.

In my case, I don't really mind that. Also, the sticky behavior will add unnecessary complexity to the simple tag system river implements (which is nice tbh).

polykernel:
I implemented rules for window position and dimension in https://github.com/riverwm/river/pull/875. Any feedback is appreciated.

PR #875, opened 2023-08-29
river: add position, dimension and fullscreen rules

polykernel:
This commit adds position and dimensions rules for configuring
the initial position and dimensions of views.

When a view is not matched by any position rules, it is centered
in the avaliable output space matching the current behavior. If
the provided position rule places the view outside of the output,
the view's position is clamped to the output bounds (with respect
to borders).

When a view is not matched by any dimensions rules, no default
dimensions is set by the server. If the provided dimensions rule
exceeds the minimum or maximum width/height constraints of the view,
the view's width/height is clamped to the constraints.

partially implements https://github.com/riverwm/river/issues/865

I am relatively new to Zig and Wayland, please let me know if I am doing anything incorrectly. I was also going to implement the snap rule proposed in the linked issue but was not sure how it should interact with the position rule.

### DIFF
--- a/completions/fish/riverctl.fish
+++ b/completions/fish/riverctl.fish
@@ -88,10 +88,10 @@ complete -c riverctl -n '__fish_seen_subcommand_from unmap'                     
 complete -c riverctl -n '__fish_seen_subcommand_from attach-mode'                 -n '__fish_riverctl_complete_arg 2' -a 'top bottom'
 complete -c riverctl -n '__fish_seen_subcommand_from focus-follows-cursor'        -n '__fish_riverctl_complete_arg 2' -a 'disabled normal always'
 complete -c riverctl -n '__fish_seen_subcommand_from set-cursor-warp'             -n '__fish_riverctl_complete_arg 2' -a 'disabled on-output-change on-focus-change'
-complete -c riverctl -n '__fish_seen_subcommand_from list-rules'                  -n '__fish_riverctl_complete_arg 2' -a 'float ssd tag output position dimensions'
+complete -c riverctl -n '__fish_seen_subcommand_from list-rules'                  -n '__fish_riverctl_complete_arg 2' -a 'float ssd tag output position dimensions fullscreen'
 
 # Options and subcommands for 'rule-add' and 'rule-del'
-set -l rule_actions float no-float ssd csd tag output position dimensions
+set -l rule_actions float no-float ssd csd tag output position dimensions fullscreen no-fullscreen
 complete -c riverctl -n '__fish_seen_subcommand_from rule-add rule-del' -n "not __fish_seen_subcommand_from $rule_actions" -n 'not __fish_seen_argument -o app-id' -o 'app-id' -r
 complete -c riverctl -n '__fish_seen_subcommand_from rule-add rule-del' -n "not __fish_seen_subcommand_from $rule_actions" -n 'not __fish_seen_argument -o title'  -o 'title' -r
 complete -c riverctl -n '__fish_seen_subcommand_from rule-add rule-del' -n "not __fish_seen_subcommand_from $rule_actions" -n 'test (math (count (commandline -opc)) % 2) -eq 0' -a "$rule_actions"
--- a/completions/fish/riverctl.fish
+++ b/completions/fish/riverctl.fish
@@ -88,10 +88,10 @@ complete -c riverctl -n '__fish_seen_subcommand_from unmap'                     
 complete -c riverctl -n '__fish_seen_subcommand_from attach-mode'                 -n '__fish_riverctl_complete_arg 2' -a 'top bottom'
 complete -c riverctl -n '__fish_seen_subcommand_from focus-follows-cursor'        -n '__fish_riverctl_complete_arg 2' -a 'disabled normal always'
 complete -c riverctl -n '__fish_seen_subcommand_from set-cursor-warp'             -n '__fish_riverctl_complete_arg 2' -a 'disabled on-output-change on-focus-change'
-complete -c riverctl -n '__fish_seen_subcommand_from list-rules'                  -n '__fish_riverctl_complete_arg 2' -a 'float ssd tag output'
+complete -c riverctl -n '__fish_seen_subcommand_from list-rules'                  -n '__fish_riverctl_complete_arg 2' -a 'float ssd tag output position dimensions'
 
 # Options and subcommands for 'rule-add' and 'rule-del'
-set -l rule_actions float no-float ssd csd tag output
+set -l rule_actions float no-float ssd csd tag output position dimensions
 complete -c riverctl -n '__fish_seen_subcommand_from rule-add rule-del' -n "not __fish_seen_subcommand_from $rule_actions" -n 'not __fish_seen_argument -o app-id' -o 'app-id' -r
 complete -c riverctl -n '__fish_seen_subcommand_from rule-add rule-del' -n "not __fish_seen_subcommand_from $rule_actions" -n 'not __fish_seen_argument -o title'  -o 'title' -r
 complete -c riverctl -n '__fish_seen_subcommand_from rule-add rule-del' -n "not __fish_seen_subcommand_from $rule_actions" -n 'test (math (count (commandline -opc)) % 2) -eq 0' -a "$rule_actions"
--- a/completions/zsh/_riverctl
+++ b/completions/zsh/_riverctl
@@ -183,9 +183,9 @@ _riverctl()
                     # In case of a new rule added in river, we just need
                     # to add it to the third option between '()',
                     # i.e (float no-float <new-option>)
-                    _arguments '1: :(-app-id -title)' '2: : ' ':: :(float no-float ssd csd tag output position dimensions)'
+                    _arguments '1: :(-app-id -title)' '2: : ' ':: :(float no-float ssd csd tag output position dimensions fullscreen no-fullscreen)'
                 ;;
-                list-rules) _alternative 'arguments:args:(float ssd tag output position dimensions)' ;;
+                list-rules) _alternative 'arguments:args:(float ssd tag output position dimensions fullscreen)' ;;
                 *) return 0 ;;
             esac
         ;;
--- a/completions/zsh/_riverctl
+++ b/completions/zsh/_riverctl
@@ -183,9 +183,9 @@ _riverctl()
                     # In case of a new rule added in river, we just need
                     # to add it to the third option between '()',
                     # i.e (float no-float <new-option>)
-                    _arguments '1: :(-app-id -title)' '2: : ' ':: :(float no-float ssd csd tag output)'
+                    _arguments '1: :(-app-id -title)' '2: : ' ':: :(float no-float ssd csd tag output position dimensions)'
                 ;;
-                list-rules) _alternative 'arguments:args:(float ssd tag output)' ;;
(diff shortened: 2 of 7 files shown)


Leon-Plickat:
> If the provided position rule places the view outside of the output, the view's position is clamped to the output bounds (with respect to borders) and a warning is logged to indicate the supplied position is invalid.

This logic should happen when the rule is applied, not when it is created. Output sizes can change.

polykernel:
I took the liberty to also add fullscreen view rules as it seems to be low hanging fruit. In addition, I renamed `WindowPosition`, `WindowDimensionI took the liberty to also add fullscreen view rules as it seems to be fairly trivial to implement.

river/View.zig
```diff
@@ -488,14 +488,65 @@ pub fn map(view: *Self) !void {
...
+            view.pending.box.y = @max(view.pending.box.y, border_width);
+            view.pending.box.y = @min(view.pending.box.y, max_y);
+            view.pending.box.y = @max(view.pending.box.y, 0);
+
+            if (out_of_bounds) {
+                log.warn("view '{?s}' position {d},{d} out of bounds, fallback to default", .{
+                    view.getTitle(),
+                    position.x,
+                    position.y,
+                });
```
ifreund, line 524: Instead of duplicating all this bounds checking logic here I suggest the following:
1. Set the pending position of the view to a matching position rule or center it.
2. Set the pending dimensions of the view to a matching dimensions rule if present.
3. That's it, you're done. The view.setPendingOutput() call below will clamp the view's position/dimensions to the constraints and output dimensions.

There's no need to do any logging about the requested dimensions being out of bounds either IMO.

ifreund (changes requested):
Hello, thanks for the patch! Here are some review comments:

river/Config.zig
```diff
@@ -51,6 +51,16 @@ pub const HideCursorWhenTypingMode = enum {
     enabled,
 };
 
+pub const ViewPosition = struct {
+    x: c_int,
+    y: c_int,
+};
+
+pub const ViewDimension = struct {
```
ifreund, line 59: Oh, and I'd just call these `Position` and `Dimensions`. There's nothing about them that is view-specific.

river/View.zig
polykernel: The reason this whole block exists is solely because of the need to check the out of bounds condition before setting the position/dimensions of the view. My original thought process was that since the behavior of the position and dimensions rule are dependent on the output dimensions, I should log when the behavior changes. If this is deemed unnecessary, I can remove the block and use the suggested `view.setPendingOutput()` approach.

doc/riverctl.1.scd
```diff
@@ -263,7 +263,7 @@ matches everything while _\*\*_ and the empty string are invalid.
 	Add a rule that applies an _action_ to views with *app-id* and *title*
 	matched by the respective _glob_. Omitting *-app-id* or *-title*
 	is equivalent to passing *-app-id* _\*_ or *-title* _\*_.
-	Some actions require an _argument_.
+	Some actions require one of more _argument_.
```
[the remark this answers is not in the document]
polykernel: Fixed.

doc/riverctl.1.scd
```diff
@@ -276,11 +276,20 @@ matches everything while _\*\*_ and the empty string are invalid.
 	  and existing views.
 	- *tag*: Set the initial tags of the view. Requires the tags as
 	  an argument. Applies only to new views.
+	- *position*: Set the initial position of the view, clamping to the
+	  bounds of the output. Requires x and y coordinates of the view as
+	  arguments. Applies only to new views.
+	- *dimension*: Set the initial dimension of the view, clamping to the
```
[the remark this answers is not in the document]
polykernel: Fixed, thanks.

doc/riverctl.1.scd
```diff
@@ -276,11 +276,20 @@ matches everything while _\*\*_ and the empty string are invalid.
...
+	  constraints of the view. Requires width and height of the view as
+	  arguments. Applies only to new views.
+	- *fullscreen*: Make the view fullscreen. Applies only to new views.
+	- *windowed*: Don't make the view fullscreen. Applies only to
```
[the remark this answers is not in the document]
polykernel: Noted.

river/View.zig
ifreund: As I said in my original comment, I don't think there is a need to do any logging in the case that the position/dimensions place the view out of bounds. Thanks for cleaning things up!

polykernel: Alright, done.

river/command/rule.zig
```diff
@@ -95,6 +100,37 @@ pub fn ruleAdd(_: *Seat, args: []const [:0]const u8, _: *?[]const u8) Error!void
...
+                .value = .{
+                    .x = x,
+                    .y = y,
+                },
+            });
+        },
+        .dimensions => {
+            const width = try fmt.parseInt(c_int, result.args[1], 10);
+            const height = try fmt.parseInt(c_int, result.args[2], 10);
+            try server.config.dimensions_rules.add(.{
```
ifreund, line 118: I think we need to error on negative x/y and definitely on negative width/height. A simple way to do this is to pass the type `u31` to `parseInt()` instead of `c_int`.

river/View.zig
```diff
@@ -427,7 +427,7 @@ pub fn setPendingOutput(view: *Self, output: *Output) void {
     if (view.pending.fullscreen) {
         view.pending.box = .{ .x = 0, .y = 0, .width = undefined, .height = undefined };
         output.wlr_output.effectiveResolution(&view.pending.box.width, &view.pending.box.height);
-    } else if (view.pending.float) {
+    } else {
```
ifreund, line 430: This change seems wrong to me. I think setting a position or dimension rule should force the view to start floating. We do need to document that this overrides `no-float` rules for example.

doc/riverctl.1.scd
```diff
@@ -274,11 +274,11 @@ For example, _abc_ is matched by _a\*_, _\*a\*_, _\*b\*_, _\*c_, _abc_, and
 _\*_ but not matched by _\*a_, _b\*_, _\*b_, _c\*_, or _ab_. Note that _\*_
 matches everything while _\*\*_ and the empty string are invalid.
 
-*rule-add* [*-app-id* _glob_|*-title* _glob_] _action_ [_argument_]
+*rule-add* [*-app-id* _glob_|*-title* _glob_] _action_ [_arguments_]
 	Add a rule that applies an _action_ to views with *app-id* and *title*
 	matched by the respective _glob_. Omitting *-app-id* or *-title*
 	is equivalent to passing *-app-id* _\*_ or *-title* _\*_.
-	Some actions require an _argument_.
+	Some actions require one of more _arguments_.
```
ifreund, line 281:
```suggestion
	Some actions require one or more _arguments_.
```

completions/zsh/_riverctl
```diff
@@ -183,9 +183,9 @@ _riverctl()
                     # In case of a new rule added in river, we just need
                     # to add it to the third option between '()',
                     # i.e (float no-float <new-option>)
-                    _arguments '1: :(-app-id -title)' '2: : ' ':: :(float no-float ssd csd tag output)'
+                    _arguments '1: :(-app-id -title)' '2: : ' ':: :(float no-float ssd csd tag output position dimensions)'
```
ifreund, line 186:
```suggestion
                    _arguments '1: :(-app-id -title)' '2: : ' ':: :(float no-float ssd csd tag output position dimensions fullscreen no-fullscreen)'
```

completions/fish/riverctl.fish
```diff
@@ -88,10 +88,10 @@ complete -c riverctl -n '__fish_seen_subcommand_from unmap'
 complete -c riverctl -n '__fish_seen_subcommand_from attach-mode'                 -n '__fish_riverctl_complete_arg 2' -a 'top bottom'
 complete -c riverctl -n '__fish_seen_subcommand_from focus-follows-cursor'        -n '__fish_riverctl_complete_arg 2' -a 'disabled normal always'
 complete -c riverctl -n '__fish_seen_subcommand_from set-cursor-warp'             -n '__fish_riverctl_complete_arg 2' -a 'disabled on-output-change on-focus-change'
-complete -c riverctl -n '__fish_seen_subcommand_from list-rules'                  -n '__fish_riverctl_complete_arg 2' -a 'float ssd tag output'
+complete -c riverctl -n '__fish_seen_subcommand_from list-rules'                  -n '__fish_riverctl_complete_arg 2' -a 'float ssd tag output position dimensions'
```
ifreund, line 91:
```suggestion
complete -c riverctl -n '__fish_seen_subcommand_from list-rules'                  -n '__fish_riverctl_complete_arg 2' -a 'float ssd tag output position dimensions fullscreen'
```

ifreund (commented):
Here's a hopefully final review, just a couple more minor things I noticed before this can be merged.

completions/fish/riverctl.fish
polykernel: Fixed, thanks.

completions/zsh/_riverctl
polykernel: It appears I did a sloppy merge, thanks for catching this.

doc/riverctl.1.scd
polykernel: Fixed, thanks.

river/command/rule.zig
polykernel: I opted to check if x/y and width/height are negative explicitly in order to assign a more appropriate error `Error.InvalidValue`. I also added the non-negative requirement to the riverctl documentation.

river/View.zig
polykernel: I followed the recommendation given by @Leon-Plickat in https://github.com/riverwm/river/issues/865#issuecomment-1655499603. I thought that it would be a good idea to keep the float, position and dimensions rule independent for composibility. My interpretation of the position/dimensions rule is that it sets the position/dimensions of the view. It is up to the "mode" (not sure what the correct term is) to decide how the value gets used. However, I do see your point that really the only case where a position or dimensions rule would be used is for floating views.

ifreund: Hmm, yeah I guess we can go that way as well. In that case position and dimensions rules would have no effect unless the view is also matched by a float rule. They also have no effect if a view is started fullscreen, which seems reasonable. It would probably be a good idea to document this explicitly for clarity.

And, unless I'm missing something, it's still fine to revert this part of the patch. If views are not started floating then their position/dimensions will be overridden by the layout generator before they are sent to the client.

river/command/rule.zig
ifreund: I don't think I agree, "out of bounds" is a bit more descriptive than "invalid value" IMO and is what we use elsewhere in river for exactly this situation.

river/View.zig
polykernel: > If views are not started floating then their position/dimensions will be overridden by the layout generator before they are sent to the client.

Ah, I didn't know that was the case. I reverted the change and documented the interaction of position and dimensions rules with float and fullscreen modes.

river/command/rule.zig
polykernel: I see, I should have looked for precedence in the source. I have updated the block to pass the type `u31` to `parseInt()` as well as the field types in the `Position` and `Dimensions` struct.

ifreund (approved):
LGTM, Thanks!

PR #894, opened 2023-10-16
init example: use float-filter-add and csd-filter-add

tgy:
The `init` example file is using outdated rivectl commands.

This PR fixes this by using the new commands.

### DIFF
--- a/example/init
+++ b/example/init
@@ -150,11 +150,11 @@ riverctl border-color-unfocused 0x586e75
 # Set keyboard repeat rate
 riverctl set-repeat 50 300
 
-# Make all views with an app-id that starts with "float" and title "foo" start floating.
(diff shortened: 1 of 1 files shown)


example/init
```diff
@@ -150,11 +150,11 @@ riverctl border-color-unfocused 0x586e75
 # Set keyboard repeat rate
 riverctl set-repeat 50 300
 
-# Make all views with an app-id that starts with "float" and title "foo" start floating.
-riverctl rule-add float -app-id 'float*' -title 'foo'
+# Make all views with an app-id that starts with "float"
+riverctl float-filter-add app-id 'float*'
```
tgy, line 154: Note: filtering on both `app-id` and `title` simultaneously does not seem to be possible anymore

tgy:
my bad! i had an outdated version

PR #895, opened 2023-10-17
completions: Rewrite zsh

novakne:
- Fix some completions that never really worked correctly, e.g `riverctl input` didn't take the input name into account. Same with rule that didn't take into account glob.
- Add a lot of documentation to help people adding new commands.
- Add new rule-[add|del] order.
- `riverctl input` now autocomplete input name from `riverctl list-inputs`

This is still not perfect but it"s an improvement and I hope it can remove a bit of confusion when there is a need to add a new commands.

### DIFF
--- a/completions/zsh/_riverctl
+++ b/completions/zsh/_riverctl
@@ -2,11 +2,15 @@
 #
 # Completion script for riverctl, part of river <https://github.com/riverwm/river>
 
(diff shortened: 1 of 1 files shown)


ifreund (approved):
Looks like an improvement to my un-zsh-ified eyes, thanks!

PR #903, opened 2023-11-05
add '-oneshot' modifer to 'enter-mode'

### LINKED ISSUE #899: Feature Request: Oneshot mode
It would be good to be able to declare a `oneshot mode` which I can switch to and no matter what is executed in it,
it will return to either the previous mode or a specific mode I declared. 

For example:
Super+L triggers a mode change to "launch". I can then press p to launch my password manager, f to launch my feeds, etc.

I could imagine this could be a flag on `declare-mode` like `-o|--once` or even `-os|--one-shot`.

Leon-Plickat:
Closes #899

### DIFF
--- a/river/LockManager.zig
+++ b/river/LockManager.zig
@@ -113,9 +113,12 @@ fn handleLock(listener: *wl.Listener(*wlr.SessionLockV1), lock: *wlr.SessionLock
                 seat.setFocusRaw(.none);
                 seat.cursor.updateState();
 
-                // Enter locked mode
-                seat.prev_mode_id = seat.mode_id;
-                seat.enterMode(1);
+                // Enter locked mode. If the currently active mode has a
+                // onehsot lifetime, ignore it and return to normal mode on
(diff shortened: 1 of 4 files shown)


Anachron:
That was fast!

Will probably test on next build from master.

Thank you <3

river/LockManager.zig
```diff
@@ -223,7 +226,9 @@ fn handleUnlock(listener: *wl.Listener(void)) void {
             seat.setFocusRaw(.none);
 
             // Exit locked mode
-            seat.enterMode(seat.prev_mode_id);
+            assert(seat.mode_id == 1);
+            assert(seat.mode_lifetime == .continuous);
```
Leon-Plickat, line 230: Whoops, don't merge yet, these asserts are invalid; I forgot that enter-mode works in locked mode too.

I'll remove them once I am in front of a computer again.

ifreund:
Honestly, I think I'm happier leaving this complexity in the hands of the users for now. It will be pushed outside of river with the window management protocol anyways.

Issue #666, opened 2022-08-07
Restore view's output on output reconnect


WizardUli:
Sway has a nice functionality where it *for each workspace remembers the last output you have **explicitly** and successfully moved a workspace to* and on an output's disappearance and reappearance it tries to restore the output for workspaces which originally used it.

This have following use-cases:
- You can disconnect a laptop from a docking station and on a reconnect have all outputs restored for all views you haven't meanwhile explicitly reassign output yourself.
- You want to temporarily use your displays with a different device (like KVM switch). On quite a few displays switching between HW inputs intermittently disconnects them even when they are continuously physically connected.
- Some displays are buggy and intermittently disconnects when going to sleep.

Even when intermittent display connections are not common, I think everybody can understand the first use-case.

Implementation could probably be as simple as having a new property _explicit output_ + some restoring logic on outputs change. I can try to come up with a prototype but it is going to some times since I haven't even seen Zig source code yet :)

Leon-Plickat:
> Implementation could probably be as simple as having a new property explicit output + some restoring logic on outputs change.

I feel like this should be a list of outputs instead of a single one. Maybe the re-appearing output isn't the last one the window was on but the one it was on before that one. In that case it would still be nice if the window was auto-moved to it.

WizardUli:
> I feel like this should be a list of outputs instead of a single one.

Yep. That sound logical.

> You can identify outputs either by connector name (like "HDMI-A-1") or by name and model, which allows identifying individual devices.

I've wanted to use whatever River is using currently for a start but yes, displays in general can be identified by several properties. Which begs to expose it to River's public API in some way:
It can be a simple configuration of what should be remembered and matched
or it can be some output change _event_ which an external application can listen to, get a full history of views' outputs and reassign them :D

chcore:
Does anyone know of a work-around for this issue we can use for the time being? I have tried both wlr-randr and wlopm to manage the power state of my 2 displays, and both of them work, but on resume all views end up on just one display.

edrex:
Sway impl: https://github.com/swaywm/sway/pull/2115

chcore:
Is this feature still planned? I had to go back to sway last year because river would move all of my windows to one display every time they came out of standby mode.

I like to check in now and then, hoping for a resolution to this as I much prefer the way river does things.

zgb124:
Also chiming in to say this issue makes river unusable for me, sadly. Would love to see it fixed in 0.3 if possible.

[...]

PR #909, opened 2023-11-10
restore views to evacuated outputs on reconnect

Leon-Plickat:
implements #666 

untested, will test tomorrow when I am back at my desk.

(should have waited for #999 to be the next number, but oh well)

### DIFF
--- a/river/Root.zig
+++ b/river/Root.zig
@@ -286,6 +286,15 @@ pub fn deactivateOutput(root: *Self, output: *Output) void {
             if (view.inflight_transaction) {
                 view.commitTransaction();
             }
+
+            // Store outputs connector name so that views can be moved back to
+            // reconnecting outputs. Skip if there is already a connector name
+            // stored to better handle the case of multiple outputs being
+            // removed sequentially.
+            if (view.output_before_evac == null) {
+                const name = mem.span(output.wlr_output.name);
+                view.output_before_evac = util.gpa.dupe(u8, name) catch null;
+            }
         }
     }
     // Use the first output in the list as fallback. If the last real output
@@ -384,6 +393,20 @@ pub fn activateOutput(root: *Self, output: *Output) void {
             while (it) |seat_node| : (it = seat_node.next) {
                 const seat = &seat_node.data;
                 seat.focusOutput(output);
+            }
+        }
+    } else {
+        // Otherwise check if any views were previously evacuated from an output
+        // with the same (connector-)name and move them back.
+        var it = root.views.iterator(.forward);
+        while (it.next()) |view| {
+            const name = view.output_before_evac orelse continue;
+            if (mem.eql(u8, name, mem.span(output.wlr_output.name))) {
+                if (view.pending.output != output) {
+                    view.setPendingOutput(output);
+                }
+                util.gpa.free(name);
+                view.output_before_evac = null;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


zgb124:
Thank you for this, it seems to work as expected. Views are on the correct output after reconnecting.

I am trying out your layout generator (Stacktile). Is it also possible to restore on reconnect things like the number of primary views, or the width ratio between primary and secondary views?

My outputs or maybe my GPU trigger a reconnect every time they come out of standby, so I would need to go back and adjust these manually for each tag every time after stepping away from my PC and coming back.

ifreund:
@Leon-Plickat Is there any reason this is a draft aside from the lack of testing? Also, it looks like it needs a rebase.

zgb124:
Is this change still on track for a merge? One small issue that remains is that the order of the restored views are not always preserved.

Leon-Plickat:
I assume you tested it? Nice, so I won't need to dig out my second screen.

Window order depends on what you do with the windows on their new output. I don't think previous window order can be restored easily.

Also, I think if the user explicitly modifies windows (i.e. resizing, moving, order) I think we actually should make the new output their native output and not restore those windows to re-connected outputs.

zgb124:
> I assume you tested it? Nice, so I won't need to dig out my second screen.

Yes, happy to report it has been working great for me for the last few months.

> Also, I think if the user explicitly modifies windows (i.e. resizing, moving, order) I think we actually should make the new output their native output and not restore those windows to re-connected outputs.

I think that's a reasonable approach.

ifreund:
> Also, I think if the user explicitly modifies windows (i.e. resizing, moving, order) I think we actually should make the new output their native output and not restore those windows to re-connected outputs.

This makes me a bit worried about where exactly to draw the line here and how easy it will be to forget to clear the remembered output in one of these cases in code. I think prioritizing predictability and code simplicity might be the same thing here.

Leon-Plickat:
On Mon Mar 11, 2024 at 10:59 AM CET, Isaac Freund wrote:
> I think prioritizing predictability and code simplicity might be the same thing here.

I am not sure this applies to all cases.

F.e. a user might disconnect an external screen from a laptop and then
continues to use the windows evacuated from that output for a while,
resizes them, moves them to different tags, changes view order, etc.
Then a few hours later they reconnect to the same screen or maybe even
a different one.

The current code will move the windows "back" to that output, but that
may not be what the user expected.

ifreund:
> F.e. a user might disconnect an external screen from a laptop and then continues to use the windows evacuated from that output for a while, resizes them, moves them to different tags, changes view order, etc. Then a few hours later they reconnect to the same screen or maybe even a different one.

To be clear, I agree with you that it would be better to forget about a stored output when a user manually interacts with a window. My concern is how that behavior can be implemented in a robust and predictable way.

Leon-Plickat:
Perhaps we can only forget the previous output on float-toggle, fullscreen-toggle, resize, move and tag changes. Those are the only operations where it's clear that they are explicit. For window order changes it gets more muddy, since changing the order for one window changes it for adjacent ones as well. We could forget the output in the command handler, but I think it's better to just stick with the unambiguous options I listed above.

Leon-Plickat:
I think we can post-pone that discussion until someone actually complains and just merge this code as-is for now.

ifreund (approved):
Sure, lets try this approach and see how it feels in practice. Thanks for the patch!

Rebased and tweaked a few names for consistency.

PR #917, opened 2023-11-13
Add a function to get current tags

valoran-M:
Hello, for my first configuration, I wanted to create a command that would allow me to go to the next and prev workspace. Unfortunately, I couldn't find a function that would allow me to retrieve the current focus tag. So I added 2 small commands on riverctl.

I hope you'll find them useful.

Thank you for your work.


valoran-M:
I've just seen the pr #811. I went through another simpler approach.

Leon-Plickat:
Putting aside the question whether we want these commands in riverctl at all, this is not the implementation for `get-focused-tags` I would favour. River already has the river-status protocol which exposes this information, which should be used instead.

PR #924, opened 2023-11-25
XdgToplevel: be more forgiving to buggy clients (fixes frame glitch)

### LINKED ISSUE #807: Window frame glitches
Seen with imv and telegram programs. Telegram starts at system startup and when I send it to another tag, the window frame remembers its previous size until you click the mouse or shift the focus from the keyboard.
glitch

MaxVerevkin:
Some clients under some circumstances (see #807) after acknowledging a configure event first submit a buffer with the old size and only then submit the new buffer.

Let's log an error but still respect the clients actions. Fixes #807.

### DIFF
--- a/river/XdgToplevel.zig
+++ b/river/XdgToplevel.zig
@@ -325,15 +325,14 @@ fn handleCommit(listener: *wl.Listener(*wlr.Surface), _: *wlr.Surface) void {
                     "client initiated size change: {}x{} -> {}x{}",
                     .{ old_geometry.width, old_geometry.height, self.geometry.width, self.geometry.height },
                 );
-                if ((view.current.float or no_layout) and !view.current.fullscreen) {
-                    view.current.box.width = self.geometry.width;
-                    view.current.box.height = self.geometry.height;
-                    view.pending.box.width = self.geometry.width;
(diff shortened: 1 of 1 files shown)


charbelnicolas:
Hopefully it gets merged soon! This is very annoying and as far as I know river is the only window manager that has this issue...

MaxVerevkin:
This strictness makes `kitty --single-instance` unusable, but it also made me realize I was doing some things wrong in my clients.

ifreund:
I'm not totally sure that "clients are buggy" is the full story with #807. I haven't looked in enough detail to say for sure yet, but I don't believe that e.g. foot is doing the obviously broken thing that imv is.

I'd like to fully understand the cases where this buggy behavior can be seen before merging a band-aid fix like this which may hide an actual bug in river's transactions system.

MaxVerevkin:
I could also reproduce a similar glitch with foot under valgrind. In fact, it is enough to add a `usleep(100*1000)` in foot's configure callback and open a few footclients. However I believe this is is different story, related to transaction timeout: `error(transaction): timeout occurred, some imperfect frames may be shown`.

ifreund:
Superseded by a531311ac6a44fef2b0ba5c662c0ec243d63778f

Issue #789, opened 2023-02-24
[Feature Request] Move Focus To Window With App-Id


ZachIndigo:
After an amount of finagling and hackering, I've got [dwm style scratchpads](https://dwm.suckless.org/patches/scratchpads/) mostly working perfectly, except for one issue. When I spawn the window and assign it to a scratchpad tag, the newly spawned window has focus (which is good). Then when I hide the scratchpad tag, the window goes away (which is good). When I bring that tag back, the window comes back (good) but is not focused (bad). I would like a riverctl command to focus a window with a specific app-id, if that's possible.

Here are [the dotfiles I'm using](https://gitlab.com/zachir/dotfiles/-/tree/wayland/river).

At a high level, the command river_sp will toggle tag `$(( 1 << ($ARG + 9) ))` (where ARG is the scratchpad number), set the tag-mask to the same `$(( 1 << ($ARG + 9) ))`, then spawn the correct scratchpad program based off of whatever number ARG is, and then reset the tagmask to `$(( (1 << 9) - 1))`.

river_sp:
```
#!/bin/sh

#{{{ binpath
BINPATH="${XDG_CONFIG_HOME:-$HOME/.config}/river/sp"
#}}}

#{{{ spawnwindow
spawnwindow () {
  case "$1" in
    1) pgrep -x sp_z ||  ${BINPATH}/sp_z ;;
    2) pgrep -x sp_x ||  ${BINPATH}/sp_x ;;
    3) pgrep -x sp_c ||  ${BINPATH}/sp_c ;;
    4) pgrep -x sp_v ||  ${BINPATH}/sp_v ;;
    5) pgrep -x sp_b ||  ${BINPATH}/sp_b ;;
    6) pgrep -x sp_a ||  ${BINPATH}/sp_a ;;
    7) pgrep -x sp_s ||  ${BINPATH}/sp_s ;;
    8) pgrep -x sp_d ||  ${BINPATH}/sp_d ;;
    9) pgrep -x sp_f ||  ${BINPATH}/sp_f ;;
    10) pgrep -x sp_g || ${BINPATH}/sp_g ;;
    11) pgrep -x sp_q || ${BINPATH}/sp_q ;;
    *) printf "Unknown scratchpad $1!\n" ; exit 1 ;;
  esac
}
#}}}

#{{{ getopts
for i in "$@"; do
  case "$i" in
    z) ARG=1  ;;
    x) ARG=2  ;;
    c) ARG=3  ;;
[...]

PR #942, opened 2023-12-23
Add select flag to toggle-focused-tags command

TheFunctionalGuy:
Hello everyone!

### The Story
Because I used both _sway_ and _Hyprland_ for some time I was a little bit annoyed that I always had to press to buttons to 'emulate' **scratchpads** on _river_.
Therefore I sat down and implemented this change.

### The Change
When the `-select` flag is set in the `toggle-focused-tags` command, the command also tries to focus the first view in the given tag.
This is done so you don't have to first toggle the focus and then manually have to select the window.

### Additional Notes
Currently this change has been running for me for more than one month straight without any problems whatsoever.
I'm open for every kind of feedback and will implement the suggested changes, if you want to.

### DIFF
--- a/river/command/tags.zig
+++ b/river/command/tags.zig
@@ -17,11 +17,14 @@
 const std = @import("std");
 const mem = std.mem;
 
+const flags = @import("flags");
+
 const server = &@import("../main.zig").server;
 const util = @import("../util.zig");
 
 const Error = @import("../command.zig").Error;
 const Seat = @import("../Seat.zig");
+const View = @import("../View.zig");
 
 /// Switch focus to the passed tags.
 pub fn setFocusedTags(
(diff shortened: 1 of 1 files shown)


TheFunctionalGuy:
This should also close #789, if I understood the comment correctly.

ifreund:
I think that implementing #912 would be a much more composable and generally useful way to achieve what you want.

TheFunctionalGuy:
You mean by doing the following?
```sh
riverctl toggle-focused-tags $tag
riverctl focus-view 1
```

How do I ensure that the toggled tag(s) are always on top (e.g. at position 1)?
Or if that's not the case how do I get the correct position?

While testing this in theory the toggled tags always got inserted into the last position they had in the stack.
So always focusing the topmost window would not work.
Am I missing something?

When this questions are resolved I would happily implement the `riverctl focus-view n` command over the holidays because I'm on your boat considering this would be a more generally useful command to achieve what I want. 👍🏻

zgb124:
Just adding a +1 for this feature (or the discussed `riverctl focus-view n` if it would solve the same problem). I am a big fan of the floating terminal scratchpad in sway, and this patch makes it doable (or at least no longer frustrating) in river as well.

ifreund:
I think this is a lot of accidental complexity and that the correct way to implement a scratchpad feature with good UX is to hardcode that feature in the window management logic rather than make the existing tag management features complex/flexible enough to do this.

Since I don't have plans to hardcode scratchpads in the river compositor process and do have plans to push window management logic out of the river compositor process to a separate "window manager" process (i.e. a layout generator on steroids) I think I'll close this PR for now.

Scratchpads with good UX will be possible and probably much easier to implement once my `river-window-management` protocol vision is realized.

Issue #904, opened 2023-11-06
Add support for 'attach-mode aside'


0xfea5:
Back when I was using dwm, there was a patch called 'attach aside' which spawned new windows on top of the side stack instead of the master area. 

```
Initial state
---------------------
|         |         |
|         |         |
|    A    |    B    |
|         |         |
|         |         |
---------------------


After spawning a new window
---------------------
|         |         |
|         |    C    |<- new window spawning here
|    A    |---------|
|         |         |
|         |    B    |
---------------------
```

I believe this is a very intuitive behavior and it would be nice if it was an option for it to complement 'attach top/bottom'.

I am willing to implement this, however I'd like to see some feedback before doing so.

Leon-Plickat:
This is right now not implementable in a clean way, since the main area and window stack are concepts contained entirely in the layout generator. River itself has no idea about any of that, it just has a list of all windows.

As an alternative idea, you could create an attach mode which inserts new windows after `n` _visible_ windows. Keeping that `n` in sync with the amount of windows in the main area can be done with either a wrapper script or by having the layout generator change that setting automatically.

0xfea5:
> This is right now not implementable in a clean way, since the main area and window stack are concepts contained entirely in the layout generator. River itself has no idea about any of that, it just has a list of all windows.
> 
> As an alternative idea, you could create an attach mode which inserts new windows after `n` _visible_ windows. Keeping that `n` in sync with the amount of windows in the main area can be done with either a wrapper script or by having the layout generator change that setting automatically.

I see the problem here and your solution seems the only option. I'll try implement on my free time this and make a PR.

sje30:
This sounds like a great idea.  @0xfea5 did you make a start at this?

0xfea5:
> This sounds like a great idea. @0xfea5 did you make a start at this?

I've spent the last month getting more familiar with Zig. I have to automate updating the number of master windows that river "sees" since this kind of information is stored inside rivertile. I will try finish it soon so it may get merged on the next release.

johnae:
@0xfea5 For what it's worth, perhaps implementing something using https://github.com/MaxVerevkin/river-luatile would be feasible? I'm using that myself for playing around with layouts quickly and for general use as well. I believe this is also in the spirit of river (as opposed to building it into river or rivertile).

sje30:
Hi @johnae 

[...]

PR #947, opened 2024-01-03
river, rivertile: implement attach-mode aside.

0xfea5:
See: https://github.com/riverwm/river/issues/904

### DIFF
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -347,7 +347,7 @@ matches everything while _\*\*_ and the empty string are invalid.
 
 ## CONFIGURATION
 
-*attach-mode* *top*|*bottom*|*after <N>*
+*default-attach-mode* *top*|*bottom*|*after <N>*
 	Set the attach mode to be used by all outputs by default.
 
 	Possible values:
@@ -355,8 +355,8 @@ matches everything while _\*\*_ and the empty string are invalid.
 	- bottom: Appends the newly spawned view at the bottom of the stack.
 	- after <N>: Inserts the newly spawned view after N views in the stack.
 
-*default-attach-mode* *top*|*bottom*|*after <N>*
-	Alias to attach-mode.
+	Note that the deprecated *attach-mode* command is aliased to
+	*default-attach-mode* for backwards compatibility.
 
 *output-attach-mode* *top*|*bottom*|*after <N>*
 	Set the attach mode of the currently focused output, overriding the value of
--- a/doc/riverctl.1.scd
+++ b/doc/riverctl.1.scd
@@ -347,8 +347,20 @@ matches everything while _\*\*_ and the empty string are invalid.
 
 ## CONFIGURATION
 
-*attach-mode* *top*|*bottom*
-	Configure where new views should attach to the view stack.
+*attach-mode* *top*|*bottom*|*after <N>*
+	Set the attach mode to be used by all outputs by default.
+
+	Possible values:
+	- top: Prepends the newly spawned view at the top of the stack.
+	- bottom: Appends the newly spawned view at the bottom of the stack.
+	- after <N>: Inserts the newly spawned view after N views in the stack.
+
+*default-attach-mode* *top*|*bottom*|*after <N>*
+	Alias to attach-mode.
+
+*output-attach-mode* *top*|*bottom*|*after <N>*
+	Set the attach mode of the currently focused output, overriding the value of
+	default-attach-mode if any.
 
 *background-color* _0xRRGGBB_|_0xRRGGBBAA_
 	Set the background color.
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -31,9 +31,10 @@ const Mode = @import("Mode.zig");
 const RuleList = @import("rule_list.zig").RuleList;
 const View = @import("View.zig");
 
-pub const AttachMode = enum {
-    top,
-    bottom,
+pub const AttachMode = union(enum) {
+    top: void,
+    bottom: void,
+    after: usize,
 };
 
 pub const FocusFollowsCursorMode = enum {
@@ -112,7 +113,7 @@ default_layout_namespace: []const u8 = &[0]u8{},
 spawn_tagmask: u32 = std.math.maxInt(u32),
 
 /// Determines where new views will be attached to the view stack.
-attach_mode: AttachMode = .top,
+default_attach_mode: AttachMode = .top,
 
 /// Keyboard repeat rate in characters per second
 repeat_rate: u31 = 25,
--- a/river/Config.zig
+++ b/river/Config.zig
@@ -32,9 +32,9 @@ const RuleList = @import("rule_list.zig").RuleList;
 const View = @import("View.zig");
 
 pub const AttachMode = union(enum) {
-    top: void,
-    bottom: void,
-    after: usize,
+    top,
+    bottom,
+    after: u32,
 };
 
 pub const FocusFollowsCursorMode = enum {
--- a/river/View.zig
+++ b/river/View.zig
@@ -33,7 +33,6 @@ const SceneNodeData = @import("SceneNodeData.zig");
 const Seat = @import("Seat.zig");
 const XdgToplevel = @import("XdgToplevel.zig");
 const XwaylandView = @import("XwaylandView.zig");
-const PendingState = @import("Output.zig").PendingState;
 
 const log = std.log.scoped(.view);
 
@@ -422,7 +421,7 @@ pub fn setPendingOutput(view: *Self, output: *Output) void {
     switch (output.attachMode()) {
         .top => output.pending.wm_stack.prepend(view),
         .bottom => output.pending.wm_stack.append(view),
-        .after => |n| view.attach_after(&output.pending, n),
+        .after => |n| view.attachAfter(&output.pending, n),
     }
     output.pending.focus_stack.prepend(view);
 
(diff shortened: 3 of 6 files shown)


MaxVerevkin:
Maybe instead of introducing the concept of "main count" in the river itself, the attach mode can be called something like "nth" and be parametrized by an integer? Then you or your layout generator may choose to set the attach mode every time the "main count" changes.

```sh
riverctl attach-mode nth $N
```

river/Config.zig
```diff
@@ -34,6 +34,7 @@ const View = @import("View.zig");
 pub const AttachMode = enum {
     top,
     bottom,
+    aside,
```
Leon-Plickat, line 37: There should be a single name for this. I like either 'aside' or 'n-th'.

river/View.zig
```diff
@@ -36,6 +36,8 @@ const XwaylandView = @import("XwaylandView.zig");
 
 const log = std.log.scoped(.view);
 
+pub var nMainCount: u64 = 1;
```
Leon-Plickat, line 39: This is the wrong place to store this, IMO.

It'd be better to convert `AttachMode` from `enum` to `union(enum)`, which can then hold the index directly.

```zig
pub const AttachMode = union(enum) {
    top: void,
    bottom: void,
    aside: usize,
};
```

Also the attach mode would then need to be a per-output configuration, which means instead of a single `attach-mode` command there would need to be two commands, `output-attach-mode` and `default-attach-mode`, similar to how layout namespaces work.

river/command/layout.zig
```diff
@@ -83,3 +84,18 @@ pub fn sendLayoutCmd(
     layout.layout.sendUserCommand(args[2]);
     if (layout == output.layout) server.root.applyPending();
 }
+
+/// Called by the layout manager to update the number of Views in the main area.
```
Leon-Plickat, line 88: This command may also be called by the user directly.
The concept of main count should not exist in river, only the layout generator.
This should be folded into the `attach-mode` command:
```sh
riverctl attach-mode aside 2
```

rivertile/main.zig
```diff
@@ -179,6 +180,37 @@ const Output = struct {
...
+                                    std.log.err("seat not advertised", .{});
+                                },
+                                error.CannotGetRegistry => {
+                                    std.log.err("failed to get registry", .{});
+                                },
+                                error.RunCommandFailed => {
+                                    std.log.err("failed to run command", .{});
+                                },
+                            }
+                        };
```
Leon-Plickat, line 213: I am pretty sure this is out of scope for rivertile and will not be merged.

Rivertile is supposed to be a reference implementation with a very restricted feature set and has no ambitions to serve the use cases of a wide range of users. Users running (and creating) other layout generators is highly encouraged.

rivertile/riverctl.zig
```diff
@@ -0,0 +1,81 @@
...
+        .failure => |failure| {
+            // A small hack to provide usage text when river reports an unknown command.
+            if (mem.orderZ(u8, failure.failure_message, "unknown command") == .eq) {
+                std.log.err("unknown command", .{});
+                return;
+            }
+            std.log.err("{s}", .{failure.failure_message});
+        },
+    }
+}
```
Leon-Plickat, line 81: In the unlikely event that Isaac wants this feature in rivertile, this would be the wrong way to implement it.

Why create a new wayland connection when we can just use the one we already have for the layout protocol?

river/command.zig
```diff
@@ -69,6 +69,9 @@ const command_impls = std.ComptimeStringMap(
         .{ "map",                       @import("command/map.zig").map },
         .{ "map-pointer",               @import("command/map.zig").mapPointer },
         .{ "map-switch",                @import("command/map.zig").mapSwitch },
+        // NOTE: `set-main-count` should only be called by the layout manager and not directly.
+        // Advertises the number of views in the main area to river.
```
Leon-Plickat, line 73:
```suggestion
```

There is absolutely no reason users should not use this command directly. In fact I suspect layout generators calling it will be the minority.

river/View.zig
```diff
@@ -421,6 +423,14 @@ pub fn setPendingOutput(view: *Self, output: *Output) void {
...
         .top => output.pending.wm_stack.prepend(view),
         .bottom => output.pending.wm_stack.append(view),
+        .aside => {
+            attach_aside(
+                &output.pending.wm_stack,
+                output.pending.tags,
+                &view.pending_wm_stack_link,
+                .pending,
+            );
+        },
```
Leon-Plickat, line 433: If you make the function take `self: *View` and `output: *Output` as arguments, it would be a bit cleaner IMO.

Leon-Plickat (changes requested):

0xfea5:
Thank you both for the comments, I will take some time to revision the changes and make another PR.

river/View.zig
0xfea5: Your suggestion was what I initially did, however there is another place (View.zig:575) where `attach_aside()` is called without an `output` to be passed as a parameter. I don't like the way it looks either so If you have any better ideas I'd be glad to see.

ifreund (changes requested):
I have not looked at any of the code here in detail. There are some high level changes that need to be made though before I do a thorough review

1. Make this not a breaking change.
2. Drop all the rivertile changes, rivertile is feature complete.
3. Document the new feature in the riverctl man page.

0xfea5:
@ifreund @Leon-Plickat

I have pushed my most recent commits. I kept the changes simple and I believe the patch is in a better state now. The only part it can may be improved is for `View:attach_aside()` to have a cleaner signature, as Leon initially suggested.

doc/riverctl.1.scd
```diff
@@ -347,8 +347,17 @@ matches everything while _\*\*_ and the empty string are invalid.
 
 ## CONFIGURATION
 
-*attach-mode* *top*|*bottom*
-	Configure where new views should attach to the view stack.
+*default-attach-mode* *top*|*bottom*|*aside <N>*
```
ifreund, line 350: This rename is still a breaking change and needs to be reverted

0xfea5: I am thinking about adding `attach-mode` as an alias to `default-attach-mode` to keep both to keep both consistency (with options like default-layout/output-layout) as well as compatibility with current configurations. What do you think?

ifreund: Sure, an alias would be fine.

0xfea5: FYI, the change is up. You may review everything if you haven't already.

river/View.zig
```diff
@@ -476,6 +485,30 @@ pub fn applyConstraints(self: *Self, box: *wlr.Box) void {
...
 
+/// Attach current view after n Views on the stack
+pub fn attach_aside(
+    view: *Self,
+    comptime state_field: @Type(.EnumLiteral),
+    comptime state_link_field: @Type(.EnumLiteral),
+    wm_stack: anytype,
+    tags: usize,
+    n: usize,
+) void {
```
ifreund, line 496: This level of metaprogramming should not be necessary. I just pushed a small refactor to master that should allow this function to only deal with the pending wm_stack.

doc/riverctl.1.scd
```diff
@@ -347,8 +347,20 @@ matches everything while _\*\*_ and the empty string are invalid.
 
 ## CONFIGURATION
 
-*attach-mode* *top*|*bottom*
-	Configure where new views should attach to the view stack.
+*attach-mode* *top*|*bottom*|*aside <N>*
+	Alias to default-attach-mode.
```
ifreund, line 351: I'd prefer a mention in the description of `default-attach-mode` such as "attach-mode may be used as an alias for this command for backwards compatibility"

ifreund (changes requested):

ifreund:
Also, could we rename this to something other than "aside"? The word doesn't really make sense in the context of inserting to a list at an arbitrary index. I think @MaxVerevkin's suggestion of "attach-mode nth 5" is more clear, as is "attach-mode after 5" or "attach-mode index 5"

river/View.zig
0xfea5: That factor does indeed help. Perhaps we can move the `tags` inside `pending` to keep similar structure as `Output.pending`?

ifreund: I'm not sure what `tags` you are referring to here, could you be more precise?

0xfea5: I am referring to the two structs `Root.fallback.pending` and `Output.pending`. Seeing you have refactored `Root.fallback.inflight` away we can move `Root.fallback.tags` in `Root.fallback.pending` like this

**Root.zig:**
```zig
fallback: struct {
...
  pending: struct {
     tags: u32 = 1 << 0,
     focus_stack: wl.list.Head(View, .pending_focus_stack_link),
     wm_stack: wl.list.Head(View, .pending_wm_stack_link),
  },
...
},
```

That has the benefit of keeping the same struct as

**Output.zig:**
```zig
pending: struct {
    tags: u32 = 1 << 0,
    focus_stack: wl.list.Head(View, .pending_focus_stack_link),
    wm_stack: wl.list.Head(View, .pending_wm_stack_link),
},
```

while also simplifying the call to `attach_aside()` since we will only have to pass `Output.pending`/`Root.fallback.pending` depending on the situation.

ifreund: Ah, yeah making those the same type is a good idea. Just pushed 6e9bd83e1d0e2132d2def338b51f66867cd88907

0xfea5:
I've rebased and pushed my latest changes. I renamed `attach_aside` to `attach_after` for better clarity as well as refactor the parts of the codes that @ifreund suggested. Since there is no inflight state at the moment, `attach_after` is hard-coded to work with the pending state.

It seems that all the changes from `master` are included in my latest commit. I am not sure if that's the default behavior of rebase or if it's a mistake on my part. Hopefully, it won't cause much trouble when merging.

ifreund (approved):
LGTM, thanks for the patch!

> It seems that all the changes from master are included in my latest commit. I am not sure if that's the default behavior of rebase or if it's a mistake on my part. Hopefully, it won't cause much trouble when merging.

Everything looks fine to me, I just added a commit with some final style tweaks of my own and squashed everything up :)

0xfea5:
Good to know everything worked well at the end. Thank you to everyone who made suggestions and/or comments, it helped me a lot.

PR #949, opened 2024-01-03
Fix #863

### LINKED ISSUE #863: Keyboard groups only work with existing keyboards, not new ones
Only already connected keyboards will be added to keyboard groups. Disconnecting and reconnecting the keyboard (both physically or by switching TTY and back) will not re-add it to the group.

Leon-Plickat:
Previously new keyboards would not be added to already existing keyboard groups on (re-)connect. Only during the creation of the groups themselves were devices added to them. This meant that only keyboards connected during startup - before the init is executed - would work with groups in a typical river session.

Closes #863

### DIFF
--- a/river/Keyboard.zig
+++ b/river/Keyboard.zig
@@ -1,6 +1,6 @@
 // This file is part of river, a dynamic tiling wayland compositor.
 //
-// Copyright 2020 The River Developers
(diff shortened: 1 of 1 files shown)


ifreund (approved):
LGTM, thanks!

PR #950, opened 2024-01-03
Keyboard group glob

Leon-Plickat:
Use globber for matching identifiers in keyboard groups.

Real-world use cases:
* A physical keyboard exposes multiple virtual keyboards with similar but not exactly equal names. Using globs simplifies the configuration *a lot* in this case.
* A user wants to have all keyboards in a group, possible with the `*` glob.

Depends on #949


ifreund (commented):
This seems like a good improvement to me!

This PR could use a rebase (sorry about that, I tweaked a comment) and a mention in the man page that globs are allowed.

ifreund (approved):
LGTM, Thanks!

PR #952, opened 2024-01-04
Implement cursor_shape_v1

### LINKED ISSUE #932: Support the wp-cursor-shape-v1 protocol
This should be fairly easy to implement once the update to wlroots 0.17 is merged.

MaxVerevkin:
Also set `xcursor_name` field of the `Cursor`. Currently it is always `null`.

I've put `handleRequestSetCursorShape` in `Server.zig` because that's where the manager is created. I should probably move it to `Cursor.zig`?

Closes #932


ifreund (approved):
This LGTM, Thanks!

I think handling having `handleRequestSetCursorShape()` in Server.zig is correct. There's only one listener needed regardless of how many Seats/Cursors created.

PR #956, opened 2024-01-06
Focus follows cursor focuses empty outputs

larstiq:
Cursor: When `focus-follows-cursor` is set and the cursor moves onto an output where no views are present on the currently visible tags, focus the output itself instead of an individual view.

This is useful e.g. when you want to spawn a terminal on your empty monitor or switch it to a different tag.  Previously such changes would happen to the monitor on which you previous focus was, despite the cursor being somewhere else.

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -909,6 +909,9 @@ pub fn checkFocusFollowsCursor(self: *Self) void {
             self.seat.focus(view);
             server.root.applyPending();
         }
+    } else {
+        // The output doesn't contain any views, just focus the output.
+        self.updateOutputFocus(self.wlr_cursor.x, self.wlr_cursor.y);
     }
 }
 


river/Cursor.zig
```diff
@@ -909,6 +909,9 @@ pub fn checkFocusFollowsCursor(self: *Self) void {
             self.seat.focus(view);
             server.root.applyPending();
         }
+    } else {
+        // The output doesn't contain any views, just focus the output.
+        self.updateOutputFocus(self.wlr_cursor.x, self.wlr_cursor.y);
```
larstiq, line 914: Personally I can't live without this so I find it hard to imagine how people have been coping so far and I didn't add any configuration, but possibly not everyone wants this behaviour and it should be configurable _somehow_.

In that case I'll need some help figuring out along which axes there should be configurability.

Leon-Plickat: I think this good behaviour and don't see the need to make this configurable.

I suspect the people using focus-follows-cursor just happen to not use multiple monitors right now. Also I think this feature would have been requested and implemented earlier if it was common to f.e. launch programs by clicking on a graphical launcher / dock, but the "canonical" ways to launch things on river are maps and fuzzel right now.

ifreund (approved):
LGTM, Thanks! And yeah, I see no reason to make this configurable.

PR #959, opened 2024-01-07
use globs for input config

Leon-Plickat:
implements #741


Leon-Plickat:
Whoops, didn't see #819 

I think this one is a bit more straight-forward though. Multiple configs matching an output is implicitly supported

Leon-Plickat:
I was initially going to leave input configs in order of declaration, but ordering by generality does make more sense on second thought.

ifreund (approved):
LGTM, Thanks!

PR #976, opened 2024-02-03
Change example init config float and csd rules

Tomcat-42:
Just a small fix to the init example that I think went unnoticed.


Leon-Plickat:
You made the change backwards. You re-introduced the old commands. River on git master uses the rule commands.

Tomcat-42:
My bad, I downloaded river from Aur and assumed it was the latest version.

PR #989, opened 2024-02-15
Adding a delay option before applying focus-follows-cursor

ahamid-ha:
Using focus-follows-cursor could be chaotic especially if there are many floating windows.
So I added a command to wait till the cursor is stable over a view for a configurable time.
For example to not change the focus unless the cursor remains for 500ms above a view:
```
riverctl focus-follows-cursor normal
riverctl focus-follows-cursor timeout 500
```


ahamid-ha:
I agree decoupling focus and rendering order would solve most cases. In fact I was looking to add another command to select the behavior, either focus change only or focus change and raising the view. But I felt it is a much bigger change.
But I think the delay option is still a valid need. Some applications act on focus change, for example closing a popup or stopping/resuming a running media.

ifreund:
Sorry, but I won't be merging this. I don't believe that timeouts like this lead to good, predictable UX in the long run. There are valid issues that this kind of timeout can workaround but they should be addressed at their root cause, for example as I have outlined for the focus follows cursor issue.

PR #990, opened 2024-02-16
Improve documentation for -c parameter in --help message

LordMZTE:
This clarifies that the `-c` parameter overrides the default init executable and isn't run in addition to it in the `--help` dialog.


ifreund (approved):
Thanks for the patch!

PR #993, opened 2024-02-19
Implement -no-xwayland flag

novakne:
Add a cli flag to `river` command to disable xwayland at runtime if river has been build with xwayland support.

It does works as expected right now but a few questions, no sure how types should be handled, for e.g. `xwayland_override_redirect: if (build_options.xwayland) *wlr.SceneTree else void,`?
And some others things that I'll write in a review to be more clear.

closes #913

### DIFF
--- a/river/Cursor.zig
+++ b/river/Cursor.zig
@@ -249,17 +249,19 @@ pub fn setTheme(self: *Self, theme: ?[*:0]const u8, _size: ?u32) !void {
         if (theme) |t| if (c.setenv("XCURSOR_THEME", t, 1) < 0) return error.OutOfMemory;
 
         if (build_options.xwayland) {
-            try xcursor_manager.load(1);
-            const wlr_xcursor = xcursor_manager.getXcursor("left_ptr", 1).?;
-            const image = wlr_xcursor.images[0];
-            server.xwayland.setCursor(
-                image.buffer,
-                image.width * 4,
-                image.width,
-                image.height,
-                @intCast(image.hotspot_x),
-                @intCast(image.hotspot_y),
-            );
+            if (server.xwayland) |xwayland| {
+                try xcursor_manager.load(1);
+                const wlr_xcursor = xcursor_manager.getXcursor("left_ptr", 1).?;
+                const image = wlr_xcursor.images[0];
+                xwayland.setCursor(
+                    image.buffer,
+                    image.width * 4,
+                    image.width,
+                    image.height,
+                    @intCast(image.hotspot_x),
+                    @intCast(image.hotspot_y),
+                );
+            }
         }
     }
 
@@ -937,6 +939,7 @@ fn updateFocusFollowsCursorTarget(self: *Self) void {
             },
             .xwayland_override_redirect => {
                 assert(build_options.xwayland);
+                assert(server.xwayland != null);
                 self.focus_follows_cursor_target = null;
             },
         }
--- a/river/Server.zig
+++ b/river/Server.zig
@@ -64,7 +64,7 @@ new_toplevel_decoration: wl.Listener(*wlr.XdgToplevelDecorationV1),
 layer_shell: *wlr.LayerShellV1,
 new_layer_surface: wl.Listener(*wlr.LayerSurfaceV1),
 
-xwayland: if (build_options.xwayland) *wlr.Xwayland else void,
+xwayland: if (build_options.xwayland) ?*wlr.Xwayland else void,
 new_xwayland_surface: if (build_options.xwayland) wl.Listener(*wlr.XwaylandSurface) else void,
 
 foreign_toplevel_manager: *wlr.ForeignToplevelManagerV1,
@@ -83,7 +83,7 @@ layout_manager: LayoutManager,
 idle_inhibitor_manager: IdleInhibitorManager,
 lock_manager: LockManager,
 
-pub fn init(self: *Self) !void {
+pub fn init(self: *Self, runtime_xwayland: bool) !void {
     self.wl_server = try wl.Server.create();
     errdefer self.wl_server.destroy();
 
@@ -130,9 +130,13 @@ pub fn init(self: *Self) !void {
     self.layer_shell.events.new_surface.add(&self.new_layer_surface);
 
     if (build_options.xwayland) {
-        self.xwayland = try wlr.Xwayland.create(self.wl_server, compositor, false);
-        self.new_xwayland_surface.setNotify(handleNewXwaylandSurface);
-        self.xwayland.events.new_surface.add(&self.new_xwayland_surface);
+        if (runtime_xwayland) {
+            self.xwayland = try wlr.Xwayland.create(self.wl_server, compositor, false);
+            self.new_xwayland_surface.setNotify(handleNewXwaylandSurface);
+            self.xwayland.?.events.new_surface.add(&self.new_xwayland_surface);
+        } else {
+            self.xwayland = null;
+        }
     }
 
     self.foreign_toplevel_manager = try wlr.ForeignToplevelManagerV1.create(self.wl_server);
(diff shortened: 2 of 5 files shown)


novakne:
Well not a review since the code is not in the changed files...

Not sure if an `if (server.runtime_xwayland)` statement is needed here or maybe an `assert`
https://github.com/riverwm/river/blob/master/river/Cursor.zig#L938-L940

Same here, not sure if it would cause any problem to let it like that
https://github.com/riverwm/river/blob/master/river/Root.zig#L131

river/Server.zig
```diff
@@ -66,6 +66,7 @@ new_layer_surface: wl.Listener(*wlr.LayerSurfaceV1),
 
 xwayland: if (build_options.xwayland) *wlr.Xwayland else void,
 new_xwayland_surface: if (build_options.xwayland) wl.Listener(*wlr.XwaylandSurface) else void,
+runtime_xwayland: bool,
```
ifreund, line 69: I think we should make `Server.xwayland` have type `?*wlr.Xwayland` rather than using a separate bool field. I think it would be a lot less error prone than leaving the pointer undefined and needing to remember to check a separate bool every time before accessing it.

river/LockManager.zig
```diff
@@ -207,7 +209,9 @@ fn handleUnlock(listener: *wl.Listener(void)) void {
     }
 
     if (build_options.xwayland) {
-        server.root.layers.xwayland_override_redirect.node.setEnabled(true);
+        if (server.runtime_xwayland) {
```
ifreund, line 212: There's no need to split this in to 2 separate if statements.
```
if (build_options.xwayland and server.xwayland != null) {
```

ifreund (changes requested):
Thanks for working on this!

> Not sure if an if (server.runtime_xwayland) statement is needed here or maybe an assert
https://github.com/riverwm/river/blob/master/river/Cursor.zig#L938-L940

An `assert(server.xwayland != null);` would be good.

> Same here, not sure if it would cause any problem to let it like that
https://github.com/riverwm/river/blob/master/river/Root.zig#L131

I think we should make this have type `?*wlr.SceneTree` and initialize it to null if xwayland is disabled at runtime.

river/LockManager.zig
novakne: Right, I wasn't sure if two statements would be the same as `if (build_options.xwayland and server.runtime_xwayland)`at buildtime, i.e. would the code be in the build with the second option?

ifreund: They are semantically equivalent in zig, `and` does not evaluate the second operand if the first is false.

river/Server.zig
ifreund: The only examples I've seen where an optional makes sense are the `Server.xwayland` and `Root.layers.xwayland_override_redirect` struct fields since these are global and we can avoid creating the objects they point to if xwayland is disabled at runtime.

The other examples you've listed don't seem like they would gain anything by using a optional type.

novakne:
@ifreund any preference, when needed, between:

```zig
if (build_options.xwayland and server.xwayland != null) {
    server.xwayland.?.<function>
}

// or 
if (build_options.xwayland) {
    if (server.xwayland) |xwayland| {
        xwayland.<function>
    }
}
```

ifreund:
> @ifreund any preference, when needed, between:
> 
> ```zig
> if (build_options.xwayland and server.xwayland != null) {
>     server.xwayland.?.<function>
> }
> 
> // or 
> if (build_options.xwayland) {
>     if (server.xwayland) |xwayland| {
>         xwayland.<function>
>     }
> }
> ```

I prefer the second one :)

novakne:
Should be better, even if I'm not super fan of the `xwayland_override_redirect` code

river/Cursor.zig
```diff
@@ -936,7 +938,7 @@ fn updateFocusFollowsCursorTarget(self: *Self) void {
                 self.focus_follows_cursor_target = null;
             },
             .xwayland_override_redirect => {
-                assert(build_options.xwayland);
+                assert(server.xwayland != null);
```
ifreund, line 941: The old assertion should stay, just add the new one on the next line

river/LockManager.zig
```diff
@@ -96,8 +96,8 @@ fn handleLock(listener: *wl.Listener(*wlr.SessionLockV1), lock: *wlr.SessionLock
     if (manager.state == .unlocked) {
         manager.state = .waiting_for_lock_surfaces;
 
-        if (build_options.xwayland) {
-            server.root.layers.xwayland_override_redirect.node.setEnabled(false);
+        if (build_options.xwayland and server.xwayland != null) {
+            if (server.root.layers.xwayland_override_redirect) |xor| xor.node.setEnabled(false);
         }
```
ifreund, line 101:
```suggestion
        if (build_options.xwayland) {
            if (server.root.layers.xwayland_override_redirect) |xor| xor.node.setEnabled(false);
        }
```

river/LockManager.zig
```diff
@@ -206,8 +206,8 @@ fn handleUnlock(listener: *wl.Listener(void)) void {
...
     }
 
-    if (build_options.xwayland) {
-        server.root.layers.xwayland_override_redirect.node.setEnabled(true);
+    if (build_options.xwayland and server.xwayland != null) {
+        if (server.root.layers.xwayland_override_redirect) |xor| xor.node.setEnabled(true);
```
ifreund, line 210:
```suggestion
    if (build_options.xwayland) {
        if (server.root.layers.xwayland_override_redirect) |xor| xor.node.setEnabled(true);
```

river/Root.zig
```diff
@@ -128,7 +128,15 @@ pub fn init(self: *Self) !void {
...
     const outputs = try interactive_content.createSceneTree();
-    const xwayland_override_redirect = if (build_options.xwayland) try interactive_content.createSceneTree();
+    const xwayland_override_redirect = blk: {
+        if (build_options.xwayland) {
+            if (server.xwayland != null) {
+                break :blk try interactive_content.createSceneTree();
+            } else {
+                break :blk null;
+            }
+        }
```
ifreund, line 138:
```suggestion
        if (build_options.xwayland and server.xwayland != null) {
            break :blk try interactive_content.createSceneTree();
        } else {
            break :blk null;
        }
```

river/XwaylandOverrideRedirect.zig
```diff
@@ -114,7 +114,9 @@ pub fn handleMap(listener: *wl.Listener(void)) void {
 
 fn mapImpl(self: *Self) error{OutOfMemory}!void {
     const surface = self.xwayland_surface.surface.?;
-    self.surface_tree = try server.root.layers.xwayland_override_redirect.createSceneSubsurfaceTree(surface);
+    if (server.root.layers.xwayland_override_redirect) |xor| {
+        self.surface_tree = try xor.createSceneSubsurfaceTree(surface);
+    }
```
ifreund, line 119: There's no need for an `if` here, just assert that the scene tree is non-null with `.?`

ifreund (changes requested):
Just a bit more minor style cleanup then this should be ready to land :)

river/Root.zig
novakne: If I do this, if `build_options.xwayland` is false it means that we have a type null instead of void

novakne: Should be good to land then :)

ifreund (approved):
Thanks for the patch! I ended up changing my mind on the xwayland_override_redirect scene tree thing. I think it's cleaner to just create it regardless of whether `-no-xwayland` is passed or not, the amount of memory "wasted" is insignificant.

ifreund:
Well, I've failed to trick github into saying this PR got merged. Codeberg's manually merged feature is so much nicer...

Issue #992, opened 2024-02-17
Crash after turning off and on output


pmkap:
This happened after executing this mapping twice in quick succession

`riverctl map $mode Super F4 spawn "wlr-randr --output eDP-1 --toggle"`

```
(gdb)  
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007de544f98393 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007de544f476c8 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007de544f2f4b8 in __GI_abort () at abort.c:79
#4  0x00000000002a7746 in os.abort () at /usr/lib/zig/std/os.zig:601
#5  0x00000000002b1154 in debug.panicImpl (trace=<optimized out>, first_trace_addr=..., msg=...) at /usr/lib/zig/std/debug.zig:441
#6  0x000000000023e189 in builtin.default_panic (msg=..., error_return_trace=0x2f4c, ret_addr=...) at /usr/lib/zig/std/builtin.zig:813
#7  0x00000000002a6502 in XdgToplevel.handleCommit () at /usr/lib/zig/std/fmt.zig:122
#8  wayland_server_core.Listener(*types.compositor.Surface).setNotify__struct_12131.wrapper (listener=<optimized out>, data=<optimized out>)
    at /home/peter/.cache/paru/clone/river-git/src/river/zig-cache/o/0f0f76efb816dd509b0f9aac6ab5ae06/wayland_server_core.zig:559
#9  0x00007de54531001e in wl_signal_emit_mutable (signal=signal@entry=0x1d41d50, data=data@entry=0x1d41a80) at ../wayland-1.22.0/src/wayland-server.c:2241
#10 0x00007de54517110a in surface_commit_state (surface=0x1d41a80, next=0x1d41c00) at ../wlroots-0.17.1/types/wlr_compositor.c:502
#11 0x00007de544dcf4f6 in ffi_call_unix64 () at ../src/x86/unix64.S:104
#12 0x00007de544dcbf5e in ffi_call_int (cif=cif@entry=0x7ffe57cd61a0, fn=<optimized out>, rvalue=<optimized out>, avalue=<optimized out>, closure=closure@entry=0x0) at ../src/x86/ffi64.c:673
[...]

PR #994, opened 2024-02-20
Root: handle output disable during a transaction

ifreund:
Attempts to fix https://github.com/riverwm/river/issues/992


ifreund:
merged, just github being lame.

PR #995, opened 2024-02-20
command/input: add map-to-output

### LINKED ISSUE #635: Support mapping input devices to outputs
If using a laptop with a touchscreen and an external monitor, it is desirable to have the touchscreen inputs interpreted relative to the laptop screen specifically. I'm not aware of any way to do this mapping automatically, so we probably need something akin to sway's `map_to_output` input configuration.

LordMZTE:
This adds support for a `input X map-to-output Y` command.

closes #635

### DIFF
--- a/river/InputConfig.zig
+++ b/river/InputConfig.zig
@@ -218,28 +218,28 @@ pub const MapToOutput = struct {
     output_name: ?[]const u8,
 
     fn apply(map_to_output: MapToOutput, device: *wlr.InputDevice) void {
-        const output = out: {
+        const wlr_output = blk: {
             if (map_to_output.output_name) |name| {
                 var it = server.root.active_outputs.iterator(.forward);
-                while (it.next()) |outp| {
-                    if (mem.eql(u8, mem.span(outp.wlr_output.name), name)) {
-                        break :out outp.wlr_output;
+                while (it.next()) |output| {
+                    if (mem.eql(u8, mem.span(output.wlr_output.name), name)) {
+                        break :blk output.wlr_output;
                     }
                 }
             }
-
-            break :out null;
+            break :blk null;
         };
 
         switch (device.type) {
             .pointer, .touch, .tablet_tool => {
                 log.debug("mapping input '{s}' -> '{s}'", .{
                     device.name,
-                    if (output) |o| o.name else "<no output>",
+                    if (wlr_output) |o| o.name else "<no output>",
                 });
 
                 // TODO: support multiple seats
-                server.input_manager.defaultSeat().cursor.wlr_cursor.mapInputToOutput(device, output);
+                const seat = server.input_manager.defaultSeat();
+                seat.cursor.wlr_cursor.mapInputToOutput(device, wlr_output);
             },
 
             // These devices do not support being mapped to outputs.
@@ -270,8 +270,9 @@ tap: ?TapState = null,
 
 pub fn deinit(self: *Self) void {
     util.gpa.free(self.glob);
-    if (self.@"map-to-output".output_name) |name|
-        util.gpa.free(name);
+    if (self.@"map-to-output".output_name) |output_name| {
+        util.gpa.free(output_name);
+    }
 }
 
 pub fn apply(self: *const Self, device: *InputDevice) void {
@@ -281,17 +282,11 @@ pub fn apply(self: *const Self, device: *InputDevice) void {
     inline for (@typeInfo(Self).Struct.fields) |field| {
         if (comptime mem.eql(u8, field.name, "glob")) continue;
 
-        if (@as(if (@typeInfo(field.type) == .Optional)
-            field.type
-        else
-            ?field.type, @field(self, field.name))) |setting|
-        {
+        if (comptime mem.eql(u8, field.name, "map-to-output")) {
+            @field(self, field.name).apply(device.wlr_device);
+        } else if (@field(self, field.name)) |setting| {
             log.debug("applying setting: {s}", .{field.name});
-            if (comptime mem.eql(u8, field.name, "map-to-output")) {
-                setting.apply(device.wlr_device);
-            } else {
-                setting.apply(libinput_device);
-            }
+            setting.apply(libinput_device);
         }
     }
 }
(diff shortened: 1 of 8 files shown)


river/InputConfig.zig
```diff
@@ -265,6 +302,14 @@ pub fn parse(self: *Self, setting: []const u8, value: []const u8) !void {
                 const ret = c.libevdev_event_code_from_name(c.EV_KEY, value.ptr);
                 if (ret < 1) return error.InvalidButton;
                 self.@"scroll-button" = ScrollButton{ .button = @intCast(ret) };
+            } else if (comptime mem.eql(u8, field.name, "map-to-output")) {
+                if (self.@"map-to-output") |mto|
+                    if (mto.output_name) |name|
+                        util.gpa.free(name);
+
+                self.@"map-to-output" = MapToOutput{
+                    .output_name = if (std.mem.eql(u8, value, "disable")) null else try util.gpa.dupe(u8, value),
```
ifreund, line 311: This doesn't match the man page docs and is inconsistent with other input configuration settings. 
```suggestion
                    .output_name = if (std.mem.eql(u8, value, "disabled")) null else try util.gpa.dupe(u8, value),
```

river/InputConfig.zig
```diff
@@ -294,6 +339,10 @@ pub fn write(self: *Self, writer: anytype) !void {
                 try writer.print("\tscroll-button: {s}\n", .{
                     mem.sliceTo(c.libevdev_event_code_get_name(c.EV_KEY, setting.button), 0),
                 });
+            } else if (comptime mem.eql(u8, field.name, "map-to-output")) {
+                try writer.print("\tmap-to-output: {s}\n", .{
+                    setting.output_name orelse "disable",
```
ifreund, line 344:
```suggestion
                    setting.output_name orelse "disabled",
```

river/InputConfig.zig
```diff
@@ -232,9 +261,13 @@ tap: ?TapState = null,
 @"pointer-accel": ?PointerAccel = null,
 @"scroll-method": ?ScrollMethod = null,
 @"scroll-button": ?ScrollButton = null,
+@"map-to-output": ?MapToOutput = null,
```
ifreund, line 264: What's the difference between this field being null and `MapToOutput.output_name` being null? it seems to me that there is no difference and they shouldn't both be optional.

Note that for the libinput config options we use a null field here to indicate that the default value from libinput is used which may depend on hardware/libinput version/whatever. That's not the case here though, the default is no mapping and is determined by us.

river/InputConfig.zig
```diff
@@ -232,9 +261,13 @@ tap: ?TapState = null,
...
 
 pub fn deinit(self: *Self) void {
     util.gpa.free(self.glob);
+    if (self.@"map-to-output") |mto|
+        if (mto.output_name) |name|
+            util.gpa.free(name);
```
ifreund, line 270: Please use braces for `if`s that are more than one line.

river/InputManager.zig
```diff
@@ -138,6 +138,21 @@ pub fn inputAllowed(self: Self, wlr_surface: *wlr.Surface) bool {
...
+        if (conf.@"map-to-output") |_| {
+            var dev_iter = self.devices.iterator(.forward);
+            while (dev_iter.next()) |dev| {
+                if (@import("globber").match(dev.identifier, conf.glob)) {
+                    conf.apply(dev);
+                }
+            }
+        }
+    }
+}
```
ifreund, line 154: This is incorrect as it will apply all configuration settings for configs with a non-null map-to-output field, not just the map-to-output setting, and potentially break the property that the config with the most specific matching glob takes precedence.

I think this function should not try to take shortcuts as an "optimization" and rather perform a full reconfiguration. That is, use the following code from `command/input.zig` and refactor that flie to use this function as well.
```zig
    var it = server.input_manager.devices.iterator(.forward);
    while (it.next()) |device| {
        for (server.input_manager.configs.items) |config| {
            if (globber.match(device.identifier, config.glob)) {
                config.apply(device);
            }
        }
    }
```

river/Root.zig
```diff
@@ -331,6 +333,7 @@ pub fn deactivateOutput(root: *Self, output: *Output) void {
         root.notifyLayoutDemandDone();
     }
     while (output.layouts.first) |node| node.data.destroy();
+    server.input_manager.reassignDevices();
```
vyivel, line 336: wlroots will automatically detach cursor from output if it's gone from the layout, so this is probably not necessary?

river/InputConfig.zig
```diff
@@ -214,6 +214,35 @@ pub const ScrollButton = struct {
...
+        };
+
+        var cur_node = server.input_manager.seats.first;
+        while (cur_node) |seat_node| {
+            log.debug("mapping input '{s}' -> '{s}'", .{
+                device.name,
+                if (output) |o| o.name else "<no output>",
+            });
+            seat_node.data.cursor.wlr_cursor.mapInputToOutput(device, output);
+            cur_node = seat_node.next;
```
ifreund, line 241: I'd rather hardcode input_manager.defaultSeat() for now and leave a `// TODO support multiple seats` comments than leave correct looking code that will cause wlroots to log a load of pointless errors.

Also, only input devices of type pointer, touch, and tablet_tool are supported for mapping to outputs and we don't support tablet_tool input devices at all yet. Please switch on `device.type` and ignore this setting if unsupported by the device.

ifreund (commented):
Thanks for taking this on! Here's some initial review :)

river/Root.zig
ifreund: I don't think wlroots does actually, at least not wlroots 0.17.

ifreund: Rather, yes wlroots has a listener for handling an output being removed from the layout but this handler does not do anything with output mappings.

LordMZTE: Correct, river will segfault after an output with a device  mapped to it is removed without this.

river/InputConfig.zig
LordMZTE: The reason I decided to do it this way is that, with your suggestion, the `apply` function wouldn't be called in case the output is null. This would result in outputs no longer being explicitly unmapped, breaking a `map-output disabled` on devices that are currently mapped as well as unmapping on input disconnect.

Leon-Plickat: I don't follow.

Whether the apply function is called solely depends on whether the field is set, not on what outputs are present either way, no?

LordMZTE: Yes, but suppose a user does `map-to-output DP-1`. The field is now set to a non-null value, and when apply is called afterwards, the input device is correctly mapped. Afterwards, the user may run `map-to-output disabled`. The field is now set to null, and nothing happens once the `InputConfig` is applied, preventing the device from being unmapped again.

What I meant when I said "the output is null" earlier is that the `output_name` field is null.

ifreund (approved):
Thanks for the PR! Rather than spend another couple rounds of review I've gone ahead and made the tweaks I want to keep river's code stye consistent and IMO simplify things :)

Issue #1013, opened 2024-03-05
attach-mode below/above not working properly


pmkap:
My PR #1012 is not working correctly. It breaks completely when using more than one tag. It is a bit more complicated then my simple patch from #1012, as we also have to take the active tags into account. I will look into it.

PR #1014, opened 2024-03-05
View: Add attachRelative function

pmkap:
Fixes the problem described in #1013.

### DIFF
--- a/river/Output.zig
+++ b/river/Output.zig
@@ -60,20 +60,6 @@ pub const PendingState = struct {
     ///
     /// This includes both floating/fullscreen views and those arranged in the layout.
     wm_stack: wl.list.Head(View, .pending_wm_stack_link),
-    ///
-    /// Index of the first view in focus_stack in the wm_stack list.
-    pub fn focusedIndex(self: *PendingState) ?u32 {
-        var focus_stack_it = self.focus_stack.iterator(.forward);
-        const focused_view = focus_stack_it.next() orelse return null;
-
-        var it = self.wm_stack.iterator(.forward);
-        var count: u32 = 0;
-        while (it.next()) |other| {
-            if (other == focused_view) break;
-            count += 1;
-        }
-        return count;
-    }
 };
 
 wlr_output: *wlr.Output,
--- a/river/View.zig
+++ b/river/View.zig
@@ -542,13 +542,13 @@ pub fn applyConstraints(self: *Self, box: *wlr.Box) void {
 }
 
 /// Attach after n visible, not-floating views in the pending wm_stack
-pub fn attachAfter(view: *Self, pending_state: *Output.PendingState, n: usize) void {
+pub fn attachAfter(view: *Self, output_pending: *Output.PendingState, n: usize) void {
     var visible: u32 = 0;
-    var it = pending_state.wm_stack.iterator(.forward);
+    var it = output_pending.wm_stack.iterator(.forward);
 
     while (it.next()) |other| {
         if (visible >= n) break;
-        if (!other.pending.float and other.pending.tags & pending_state.tags != 0) {
+        if (!other.pending.float and other.pending.tags & output_pending.tags != 0) {
             visible += 1;
         }
     }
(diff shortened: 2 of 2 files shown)


river/Output.zig
```diff
@@ -70,7 +71,7 @@ pub const PendingState = struct {
         var count: u32 = 0;
         while (it.next()) |other| {
             if (other == focused_view) break;
-            count += 1;
+            if (other.current.tags & self.tags > 0) count += 1;
```
ifreund, line 74: Then needs to check `pending` tags not current tags as we are working with pending output state.

ifreund (commented):
Thanks for the fix!

ifreund:
Actually, there's another problem with this. It still doesn't match the counting of `view.attachAfter()` exactly as that function skips floating views for some reason.

pmkap:
> Actually, there's another problem with this. It still doesn't match the counting of `view.attachAfter()` exactly as that function skips floating views for some reason.

I see. I think it should be enough then to skip the floating views also when counting the index, see latest force-push.

pmkap:
I think for the attachAfter command it makes sense to ignore floating views, it's more intuitive...

river/Output.zig
```diff
@@ -70,7 +71,9 @@ pub const PendingState = struct {
         var count: u32 = 0;
         while (it.next()) |other| {
             if (other == focused_view) break;
-            count += 1;
+            if (other.pending.tags & self.tags > 0 and !other.pending.float) {
+                count += 1;
+            }
```
ifreund, line 76: At this point, this function is a lot more specialized than the name suggests. I think we should move it to `View` and call it something like `attachIndex` or similar.

pmkap: @ifreund I am not sure I agree, `attachIndex` sounds the same as `attachAfter`. 

Other ideas
* make function `attachRelative` with an enum {above, below} in `View`
* move the `focusedIndex` out of the PendingState, but still in Output and make the name more descriptive.

river/View.zig
```diff
@@ -557,6 +556,28 @@ pub fn attachAfter(view: *Self, pending_state: *Output.PendingState, n: usize) v
...
 
+/// Attach above or below the currently focused view
+pub fn attachRelative(view: *Self, pending_state: *Output.PendingState, mode: AttachRelativeMode) void {
+    var focus_stack_it = pending_state.focus_stack.iterator(.forward);
+    const focused_view = focus_stack_it.next();
+
+    if (focused_view == null) {
+        pending_state.wm_stack.append(view);
+        return;
+    }
```
ifreund, line 567:
```suggestion
    const focused_view = focus_stack_it.next() orelse {
        pending_state.wm_stack.append(view);
        return;
    }
```

river/View.zig
```diff
@@ -557,6 +556,28 @@ pub fn attachAfter(view: *Self, pending_state: *Output.PendingState, n: usize) v
...
+    var it = pending_state.wm_stack.iterator(.forward);
+    while (it.next()) |other| {
+        if (other == focused_view) {
+            switch (mode) {
+                .above => it.current.prev.?.insert(&view.pending_wm_stack_link),
+                .below => it.current.insert(&view.pending_wm_stack_link),
+            }
+            return;
+        }
+    }
```
ifreund, line 578: This appears to have the same issue of not respecting tags as the original patch.

river/View.zig
```diff
@@ -557,6 +556,28 @@ pub fn attachAfter(view: *Self, pending_state: *Output.PendingState, n: usize) v
     it.current.prev.?.insert(&view.pending_wm_stack_link);
 }
 
+/// Attach above or below the currently focused view
+pub fn attachRelative(view: *Self, pending_state: *Output.PendingState, mode: AttachRelativeMode) void {
+    var focus_stack_it = pending_state.focus_stack.iterator(.forward);
+    const focused_view = focus_stack_it.next();
```
ifreund, line 562: I'm wondering if there is a case where the first view in the pending focus stack is not visible with the pending tags of the output. I don't think there is, so it would be good to add an assert that `focus_view.pending.tags & pending_state.tags != 0` here as we rely on that fact.

ifreund (commented):
I like having this feature entirely encapsulated in one function in View, good call :)

river/View.zig
pmkap: @ifreund 
It seems not to be necessary here from my testing.
Because we are inserting directly in the `wm_stack` without the detour over the index counting...

river/View.zig
pmkap: Ah yes I forgot orelse can be a block. Thanks!

river/View.zig
pmkap: This assertion fails when opening the first view on tag with other tags already in use.

river/View.zig
ifreund: Indeed, another benefit of having this logic all in one function!

river/View.zig
ifreund: Ah yeah, makes sense. It would make me happy if that edge case was more visible in this code to stop me from adding a bad assertion or otherwise breaking things in the future. Maybe it just needs a comment to prevent future regressions.

ifreund (approved):
I took the liberty of applying the final polish myself, thanks for the fix!

pmkap:
@ifreund 
Thanks for making this feature possible and for finishing this PR!
Peter

PR #1015, opened 2024-03-08
Fix/missing completions

novakne:
I feel like that was a good time to do this before 0.3, still some works to do for bash and fish if someone else want, and for some descriptions, as noted in the commit message.

I also update the mailmap for me, don't think it needs  a separate PR for this


ifreund (approved):
LGTM, thanks for taking care of this! As you know, I'm quite guilty of neglecting the completion scripts myself :)

PR #1026, opened 2024-03-15
river: use "pencil" cursor shape for TabletTool

Leon-Plickat:
Use "pencil" cursor shape instead of "default" for tablet tool devices.

### DIFF
--- a/river/TabletTool.zig
+++ b/river/TabletTool.zig
@@ -186,7 +186,7 @@ pub fn proximity(tool: *TabletTool, tablet: *Tablet, event: *wlr.Tablet.event.Pr
 
             tool.wlr_cursor.warpAbsolute(tablet.device.wlr_device, event.x, event.y);
 
-            tool.wlr_cursor.setXcursor(tablet.device.seat.cursor.xcursor_manager, "default");
+            tool.wlr_cursor.setXcursor(tablet.device.seat.cursor.xcursor_manager, "pencil");
 
             tool.passthrough(tablet);
         },
@@ -259,6 +259,8 @@ fn passthrough(tool: *TabletTool, tablet: *Tablet) void {
             tool.wp_tool.notifyMotion(result.sx, result.sy);
             return;
         }
+    } else {
+        tool.wlr_cursor.setXcursor(tablet.device.seat.cursor.xcursor_manager, "pencil");
     }
 
     tool.wp_tool.notifyProximityOut();


ifreund:
My only concern here is whether this "pencil" shape is present in all cursor themes in practice. If not, we need a fallback.

Leon-Plickat:
I did some research:

wlroots uses libxcursor for cursors. `XCursor.3` states:

>  Xcursor is a simple library designed to help locate and load cursors.  Cursors can be loaded from files or memory.  A library of common cursors exists which map to the standard Xcursor names.
> [...]
> Xcursor  is  built  in  a  couple  of  layers; at the bottom layer is code which can load cursor images from files.  Above that is a layer which locates cursor files based on the library path and theme.  At the top is a layer which builds cursors either out of an image loaded from a file or one of the standard X cursors.

So apparently bitmaps for all standard x cursors are already provided. The pencil cursor is mentioned in `X11/cursorfont.h`, so it should be fine to just use it.

I must admit I could not find where the actual bitmaps are located, however I think since "pencil" is part of the standard / original x cursors, a theme not providing it can be considered broken. We also have precedence of using other non-"default" cursors for move and resize without checking whether these exist and it is perfectly possible for a theme to not define "default" either. Since loading such a broken theme only results in none of the missing cursors being displayed and an error message being logged, I believe it's fine to not check for whether any of these cursors images exist in a theme before using it.

FWIWthe old X11 cursors and Adwaita and Breeze include the "pencil" cursor, which are arguable the three most widely used cursor themes.

ifreund:
> wlroots uses libxcursor for cursors

Technically it doesn't, it uses a port of `libwayland-cursor` which in turn borrows code from libxcursor I belive.

First, it tries to load a cursor theme from disk from these paths: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/5a5aa725713bd51acfb0ec44092b9d6806dbd171/xcursor/xcursor.c#L484-530

In addition, wlroots embeds a small default cursor theme here which allows river to draw a cursor even if no xcursor themes are installed on the system: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/xcursor/cursor_data.h

The `default` cursor and all of the resize cursors we use are included in the embedded fallback theme. The `move` cursor we use isn't, though there is an `all-scroll` cursor that looks somewhat similar but worse IMO. There's no pencil-like cursor in the fallback data unfortunately.

In practice, I think pretty much any real linux system will want a real cursor theme installed so I don't care too much about having the prettiest fallback cursor selection. I think we should probably fallback to the basic `default` cursor image if any of the fancier images we try to use aren't available. However, it seems like the current wlroots API makes that quite annoying to implement currently. I reckon we can probably get away without worrying about it. If we happen to get a bug report eventually we can reconsider :)

river/TabletTool.zig
```diff
@@ -259,6 +259,8 @@ fn passthrough(tool: *TabletTool, tablet: *Tablet) void {
             tool.wp_tool.notifyMotion(result.sx, result.sy);
             return;
         }
+    } else {
+        tool.wlr_cursor.setXcursor(tablet.device.seat.cursor.xcursor_manager, "pencil");
```
ifreund, line 263: What's up with this change? It's not mentioned in the commit message...

Leon-Plickat: Merely changing the cursor to `"pencil"` on line 189 does not have the desired result: When a "new" tablet tool appears, i.e. the user moves the pen towards the pad, the cursor will be correct, however the `else` branch is needed because otherwise moving the tablet tool off a surface will not update the cursor image. So if a client sets it to something custom, that would persist until either the tablet tool disappears and re-appears or until the client or another client sets a different cursor image, which IMO is not the desired behaviour.

Leon-Plickat:
I wonder whether "move" and "pencil" could be added to the wlroots built-in theme. Both should be fairly common so at least to me that would make sense.

river/TabletTool.zig
ifreund: Yeah that makes sense, I think I'd consider that a bugfix actually.

ifreund (approved):
LGTM, Thanks for the patch!

PR #1033, opened 2024-05-12
Make the cursor warp when swapping windows

ymcx:
The cursor should stay on top of the focused window when swapping, just like what was agreed upon in https://codeberg.org/river/river/issues/983.


ifreund:
Perhaps this trick will help: https://github.com/riverwm/river/pull/1034

